PR #36, opened 2026-04-15
Refactor init bootstrap flow

### LINKED ISSUE #33: Refactor cli init: minimal surface, one upsert codepath, surface identity creation
## Problem

`cli init` has grown to do several unrelated things in one call, which makes the "first run" experience verbose, makes re-runs surprising, and hides the one thing that makes this runtime distinctive: DefraDB-backed DID identity creation.

Today `init` does all of:

1. Creates the embedded data dir and key dir, writes `init.json`, clears the runtime state file.
2. Validates every runtime schema against the DefraDB node (13+ collections).
3. Creates four documents via `apply_bootstrap_templates()` using a parallel write path separate from the existing `config X set` commands: `AgentPrincipal`, `InferenceBackend`, `AgentBehavior`, `ToolSelection`.

The `config backend set` / `behavior set` / `tools set` commands already exist as the canonical per-document mutation surface, but `init` doesn't use them.

## Proposed shape

`init` remains the happy-path bootstrap and still creates all four documents so first-run is runnable. But its internals and its output change.

**Opinionated defaults (runs out-of-the-box against a local Ollama):**

- `InferenceBackend`: endpoint `http://localhost:11434/v1`, model `gemma4-26b-a4b`.
- `AgentBehavior`: points at the above, with a standard tool selection.
- `ToolSelection`: `standard-readonly` template.
- `AgentPrincipal`: fresh DID tied to the generated key.

If Ollama isn't running or the model isn't pulled, `server` fails with a clear pointer to `config backend set` / `config backend discover-models`.

**What changes:**

- **One upsert codepath per document type.** Remove `apply_bootstrap_templates()`; have `init` call the same upsert functions that back `config backend set`, `config behavior set`, `config tools set`, and the principal bootstrap. There should be exactly one place that knows how to write each document kind.
- **Surface DID/identity creation.** The key + principal generation flow from `defradb.rs` is currently invisible. `init` should print the generated DID, the key path, and explain that this identity is the permission boundary for everything the agent does.
- **Schema validation moves out of `init`.** `diagnose` / `status` validate schemas; `server` re-validates on startup. `init` should not touch 13+ collections before the user has committed to anything.
- **Runtime state clearing becomes explicit.** Add a dedicated `reset` subcommand. Add an `--reset` flag to `init` for the convenience case. The current unconditional clear on every `init` is surprising (and I thought it was already flag-gated — needs verification against `clear_runtime_state()` call in `main.rs`).
- **Drop `init.json` if possible.** Most fields appear derivable; anything essential should live in a document. If something genuinely needs a filesystem config, document why.
- **Clear next-steps output.** After `init`, print the DID, the paths, the default backend target, and the exact commands to run next (`ollama pull gemma4-26b-a4b`, `defra-agent server`, or `defra-agent config backend set ...` to point elsewhere).

## Out of scope

- Adding new `config X set` subcommands — the existing ones cover all four document kinds.
- Changing the bootstrap document shapes — see #34 (reduce backend-config wrappers).

## Acceptance

- `init` is implemented as composition of per-document upsert functions shared with `config X set`.
- First-run output prominently shows the generated DID and key path.
- `reset` subcommand exists; `init --reset` delegates to it.
- `init.json` is gone or its remaining purpose is documented.
- Schema validation is no longer inside `init`.

jackzampolin:
Fixes #33.

Summary:
- Refactor init to create only config bootstrap schemas and write backend, behavior, and tool-selection documents through shared upsert helpers used by the config commands.
- Default init to local Ollama at http://localhost:11434/v1 with gemma4-26b-a4b, surface the identity key path and DID permission boundary, and preserve runtime state unless reset is explicit.
- Add reset / init --reset, remove the old bootstrap templates, and document why init.json remains as filesystem-local context.
- Keep diagnose healthy for fresh offline config homes while still reporting runtime-only schemas as missing until server creates them.
- Rebase on current main and preserve the new max_queue_depth backend field in init/config upserts.
- Add an ignored live onboarding demo smoke test: init, server, CLI config backend/tools/behavior mutations, then two chat turns against live inference using filesystem tools.

Tests:
- cargo check -p defra-agent-cli
- cargo test -p defra-agent-cli --test cli_e2e -- --nocapture
- cargo test -p defra-agent
- cargo test -p defra-agent-cli --test cli_e2e standard_onboarding_live_demo_runs_real_conversation_with_filesystem_tools

Live smoke workflow:
- .github/workflows/live-smoke.yml now runs standard_onboarding_live_demo_runs_real_conversation_with_filesystem_tools with --ignored when manually dispatched.


jackzampolin (commented):
Self-review complete.

Findings: no blockers or correctness issues found in the PR diff.

Checked specifically:
- Rebase conflict with current main's max_queue_depth field: the shared InferenceBackend writer now carries max_queue_depth through add/update, and both init and config backend set populate it.
- Fresh init schema scope: init only creates config bootstrap schemas; diagnose still reports runtime-only schemas as absent but does not degrade a fresh offline config home for that reason.
- Runtime-state semantics: init preserves runtime.json unless --reset is passed; reset removes only persisted runtime state.
- Old bootstrap .gql files are deleted and there are no stale bootstrap references.

Residual risk: init now has a default model for all init invocations. That is intentional for the local Ollama default path, but hosted/custom backends still need operators to pass the correct --model-name.

Verified locally:
- cargo check -p defra-agent-cli
- cargo test -p defra-agent-cli --test cli_e2e -- --nocapture
- cargo test -p defra-agent

jackzampolin:
Self-review complete.

Findings: no code blockers found in the PR diff.

Checked specifically:
- Rebase conflict with current main's max_queue_depth field: the shared InferenceBackend writer carries max_queue_depth through add/update, and both init and config backend set populate it.
- Fresh init schema scope: init only creates config bootstrap schemas; diagnose still reports runtime-only schemas as absent but does not degrade a fresh offline config home for that reason.
- Runtime-state semantics: init preserves runtime.json unless --reset is passed; reset removes only persisted runtime state.
- Old bootstrap .gql files are deleted and there are no stale bootstrap references.

Residual risk: init now has a default model for all init invocations. That is intentional for the local Ollama default path, but hosted/custom backends still need operators to pass the correct --model-name.

Verified locally:
- cargo check -p defra-agent-cli
- cargo test -p defra-agent-cli --test cli_e2e -- --nocapture
- cargo test -p defra-agent

CI note: rust-and-cli failed before compilation because Actions could not fetch the existing git dependency ssh://git@github.com/sourcenetwork/defradb.rs.git after the workflow rewrote it to HTTPS; GitHub returned Repository not found / revision not found. The same tests pass locally with SSH access.

jackzampolin:
Follow-up for the standard onboarding flow:

Added an ignored live demo smoke test, `standard_onboarding_live_demo_runs_real_conversation_with_filesystem_tools`, and pointed `.github/workflows/live-smoke.yml` at it.

The test now follows the demo path through the CLI:
1. `defra-agent init` via the CLI, using defaults unless `DEFRA_AGENT_CLI_E2E_MODEL_ENDPOINT` / `DEFRA_AGENT_CLI_E2E_MODEL_NAME` are provided by the workflow.
2. `defra-agent server` via the CLI test harness.
3. CLI config mutations while the server is running:
   - `config backend set`
   - `config tools set`
   - `config behavior set`
4. Two `defra-agent chat --output-format json --session-id ...` turns against the same conversation.
5. Prompts force filesystem tool use, and the test asserts both final responses and persisted `list_files` / `read_file` tool results contain the generated file tokens.

Local verification after this follow-up:
- `cargo test -p defra-agent-cli --test cli_e2e standard_onboarding_live_demo_runs_real_conversation_with_filesystem_tools` (compile/ignored check)
- `cargo test -p defra-agent-cli --test cli_e2e -- --nocapture`
- `DEFRA_AGENT_CLI_E2E_MODEL_ENDPOINT=http://100.73.235.38:8000/v1 DEFRA_AGENT_CLI_E2E_MODEL_NAME=MiniMax-M2.7-NVFP4 cargo test -p defra-agent-cli --test cli_e2e standard_onboarding_live_demo_runs_real_conversation_with_filesystem_tools -- --ignored --nocapture --test-threads=1` (passed against live inference)

PR #37, opened 2026-04-15
Remove backend capability config wrappers

### LINKED ISSUE #34: Reduce backend-config wrappers: pipe rig types through, drop duplicated capability metadata
## Problem

The agent runtime passes rig types through cleanly on the tool side (`impl Tool` → `Box<dyn ToolDyn>` → `agent.tools(...)`) and on the inference path (`rig::Agent<M>`, `stream_prompt`, `CompletionError`). But around *backend configuration* we still have a wrapper layer that duplicates information rig already owns and adds asymmetry without carrying its weight.

Guiding principle for this cleanup: **wrap only where we add DefraDB-specific semantics; otherwise pipe rig through directly.**

## Concrete targets

1. **`BackendProviderAdapter` trait with two impls.** `OpenAiCompatibleAdapter` and `OpenRouterAdapter` differ only in verification and model-discovery paths. Client building already delegates to rig. Collapse these toward rig's own provider abstractions; keep a minimal verification hook only if rig doesn't expose an equivalent.

2. **Asymmetric `build_openrouter_client()`.** Lives next to the adapter trait but bypasses it. Either route OpenRouter through the unified path or remove the adapter trait entirely in favor of direct rig client construction at the call site.

3. **Duplicated capability metadata.** `InferenceBackend` schema and `BehaviorConfig` both carry fields that rig already knows or can introspect:
   - `supports_tool_calls`
   - `supports_streaming`
   - `supports_structured_outputs`
   - `supports_json_schema`
   - `max_output_tokens`
   - `context_window`

   Drop these from the schema and source from rig. Where a user genuinely needs to override rig (e.g., a provider misreports), keep a minimal override surface rather than requiring the values.

4. **`BehaviorConfig` shrinks accordingly.** Only fields that are DefraDB-side policy (principal reference, behavior id, tool selection reference, prompt, etc.) stay. Capability plumbing goes away.

## Scope

- Schema migration for `InferenceBackend` is in scope (drop the capability fields, bump schema version, update `config apply` / `diff` / `validate` to handle old documents gracefully during transition).
- `config backend discover-models` stays, but shouldn't need to populate capability fields anymore.

## Out of scope

- The tool surface — already piped through cleanly.
- The streaming / inference path in `BehaviorDaemon` — already direct.
- Rig version bump (separate concern).

## Related

Companion to #33 (cli init refactor). The init refactor should land first or coordinate so the new shared upsert codepaths don't need to be re-touched when capability fields are dropped.

## Acceptance

- Backend capability fields removed from `InferenceBackend` schema and `BehaviorConfig` struct.
- `BackendProviderAdapter` either collapsed or removed; no asymmetric OpenRouter-only path.
- Callers obtain capability info from rig (or a thin override layer), not from DefraDB documents.

jackzampolin:
## Summary
- Remove duplicated backend capability fields from `InferenceBackend`, runtime backend parsing, behavior runtime config, CLI upserts/exports, and demo manifests.
- Remove the `BackendProviderAdapter` layer and construct rig OpenAI-compatible/OpenRouter clients directly at runtime call sites while keeping backend model discovery.
- Bump config export format to `defra-agent-config/v2` and add migration/normalization for v1 imports and desired-state manifests that still include deprecated backend fields.
- Update docs/tests so backend capability fields are treated as deprecated migration input, not live desired state.

## Tests Run
- `cargo fmt --check`
- `cargo check`
- `cargo test -p defra-agent-cli --bin defra-agent`
- `cargo test -p defra-agent --lib backend_registry`
- `cargo test -p defra-agent --lib document_view`
- `cargo test -p defra-agent --lib completion_factory`
- `cargo test -p defra-agent --test backend_auth`
- `cargo test -p defra-agent-cli --test cli_e2e config_validate_accepts_normalized_manifest_root`
- `cargo test -p defra-agent-cli --test cli_e2e config_diff_reports_no_changes_for_matching_live_state`
- `cargo test -p defra-agent-cli --test cli_e2e config_diff_reports_updates_for_changed_backend_manifest`
- `cargo test -p defra-agent-cli --test cli_e2e config_apply_updates_backend_from_fresh_init_home_locally`
- `cargo test -p defra-agent-cli --test cli_e2e config_backend_discover_models_supports_explicit_preset_probe`
- `cargo test -p defra-agent-cli --test cli_e2e config_backend_set_preset_and_discover_models_from_backend_id`
- `cargo test -p defra-agent-cli --test cli_e2e config_export_import_round_trips_offline_and_requires_override`

## Migration Notes
- New exports use `defra-agent-config/v2`.
- v1 config import bundles are accepted and normalized to v2; deprecated backend fields are stripped before import/apply.
- Desired-state `inference-backends.json` files that still contain `supports_tool_calls`, `supports_streaming`, `supports_structured_outputs`, `supports_json_schema`, `context_window`, or `max_output_tokens` are accepted during validation and ignored for diff/apply convergence.
- `config backend discover-models` continues to return discovered model IDs without producing or requiring backend capability fields.

Closes #34


jackzampolin:
## Self-Review

Findings: none blocking.

Checks performed:
- Behavioral regressions: runtime, oneshot, and scheduler still construct rig clients directly for OpenAI-compatible and OpenRouter providers; OpenRouter provider preferences remain in the agent builder path.
- Old config compatibility: v1 import bundles and desired-state backend manifests with deprecated capability fields are normalized/ignored before import, validate, diff, and apply.
- Missing tests: added CLI unit migration tests and ran focused CLI config e2e coverage for validate, diff, apply, export/import, backend set, and discover-models.
- Docs: updated schema README and demo manifest; README did not reference the removed `--supports-*` flags.

Residual note: this intentionally removes the public backend adapter/build-client re-exports, so external callers using those wrappers should construct rig clients directly or use `discover_backend_models` for discovery.

PR #39, opened 2026-04-15
Support AgentRequest metadata and sampling overrides

### LINKED ISSUE #23: Support per-request metadata + sampling overrides on AgentRequest
## Motivation

Downstream users (e.g., the amygdala inference benchmark runner) need to submit `AgentRequest` docs that carry (a) opaque correlation metadata and (b) per-request sampling parameters that override the behavior's `InferenceProfile` defaults. Today neither is possible:

- `AgentRequest` has no metadata field. Callers must overload `request_id` or embed JSON frontmatter in `content`, which pollutes the prompt visible to the model.
- `InferenceProfile.temperature` exists as a schema field but is not applied to the inference request. `crates/defra-agent/src/completion_factory.rs` L35–54 sets only `tool_choice` + OpenRouter-specific params; `temperature`, `top_p`, `top_k`, `max_tokens` are dropped on the floor.
- Different model families have different optimal sampling (MiniMax-M2.7 wants T=1.0 / top_p=0.95 / top_k=40; Qwen/GLM for tool-use want T=0). A benchmark that pins every model at one temperature isn't informative.

## Proposed schema change

`crates/defra-agent/schemas/agent/agent_request.graphql` — add five optional fields:

```graphql
type AgentRequest @branchable {
    # existing fields …
    temperature: Float
    top_p: Float
    top_k: Int
    max_tokens: Int
    metadata: String    # opaque caller-owned JSON blob; agent treats as passthrough
}
```

`metadata` as a JSON-string keeps the schema flat and DefraDB-friendly (no nested maps). Caller serializes, caller deserializes. Agent persists it unchanged and does not surface it to the model.

## Proposed runtime change

Precedence: `AgentRequest.<param>` (if set) → `InferenceProfile.<param>` (if set) → rig/provider default.

1. `crates/defra-agent/src/lifecycle/materialize.rs` — hydrate the five new fields into the in-memory request struct.
2. `crates/defra-agent/src/agent.rs` `behavior_config_from_documents()` — accept an override struct derived from the request; apply precedence.
3. `crates/defra-agent/src/completion_factory.rs` L35–54 — extend the completion request construction to pass `temperature`, `top_p`, `top_k`, `max_tokens` to the provider. Use rig's existing params builder where available; for params rig doesn't expose (`top_k` in particular), either land a rig PR or append to an `extra_body` if the provider supports it.

## Acceptance criteria

- Submitting an `AgentRequest` with `temperature: 0.0` produces deterministic output from a temperature-sensitive model.
- Submitting an `AgentRequest` with `metadata: "{\"run_id\":\"foo\"}"` persists the string verbatim and is readable via GraphQL query; the model does not see it.
- `InferenceProfile.temperature` continues to be honored when `AgentRequest.temperature` is unset.
- No regression on the existing CLI e2e (`cargo test -p defra-agent-cli --test cli_e2e -- --nocapture --test-threads=1`).

## Non-goals

- `ModelSamplingProfile` as a separate DefraDB collection. Callers can maintain their own model→sampling table and stamp values on each request. Revisit if/when multiple independent callers duplicate this logic.
- `repetition_penalty` / `min_p` — add when a downstream use case actually needs them.

jackzampolin:
## Summary
- add optional AgentRequest metadata and sampling override fields
- hydrate request fields into watcher/runtime requests and keep metadata out of prompt content
- apply per-request sampling over behavior InferenceProfile defaults when building Rig calls
- expose request submit/show CLI support for the new fields

Fixes #23

## Tests
- cargo check -p defra-agent -p defra-agent-cli
- cargo test -p defra-agent completion_factory -- --nocapture
- cargo test -p defra-agent pending_request_hydrates_sampling_fields_and_metadata -- --nocapture
- cargo test -p defra-agent-cli --test cli_e2e -- --nocapture --test-threads=1

### DIFF
--- a/crates/defra-agent-cli/src/main.rs
+++ b/crates/defra-agent-cli/src/main.rs
@@ -1991,6 +1991,16 @@ struct RequestSubmitArgs {
     session_id: Option<String>,
     #[arg(long)]
     behavior_id: Option<String>,
+    #[arg(long)]
+    temperature: Option<f64>,
+    #[arg(long)]
+    top_p: Option<f64>,
+    #[arg(long)]
+    top_k: Option<i64>,
+    #[arg(long)]
+    max_tokens: Option<i64>,
+    #[arg(long)]
+    metadata: Option<String>,
     #[arg(long = "output-file")]
     output_file: Option<PathBuf>,
     #[arg(long, default_value_t = false)]
@@ -3026,6 +3036,13 @@ async fn request_submit(args: RequestSubmitArgs) -> Result<()> {
         &content,
         args.session_id.as_deref(),
         args.behavior_id.as_deref(),
+        RequestSubmitOptions {
+            temperature: args.temperature,
+            top_p: args.top_p,
+            top_k: args.top_k,
+            max_tokens: args.max_tokens,
+            metadata: args.metadata.clone(),
+        },
     )
     .await?;
     let request_summary = json!({
@@ -3033,6 +3050,11 @@ async fn request_submit(args: RequestSubmitArgs) -> Result<()> {
         "session_id": submitted.session_id,
         "agent_did": submitted.agent_did,
         "behavior_id": submitted.behavior_id,
+        "temperature": submitted.temperature,
+        "top_p": submitted.top_p,
+        "top_k": submitted.top_k,
+        "max_tokens": submitted.max_tokens,
+        "metadata": submitted.metadata,
     });
     if args.no_wait {
         print_json(&request_summary)?;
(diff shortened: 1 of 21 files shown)


jackzampolin:
Self-review notes:

- No code findings after reviewing the committed diff. The request metadata path preserves the exact string and is not included in prompt content; per-request sampling is applied on a cloned Rig agent so overrides do not leak across requests.
- Local verification passed:
  - `cargo check -p defra-agent -p defra-agent-cli`
  - `cargo test -p defra-agent completion_factory -- --nocapture`
  - `cargo test -p defra-agent pending_request_hydrates_sampling_fields_and_metadata -- --nocapture`
  - `cargo test -p defra-agent-cli --test cli_e2e -- --nocapture --test-threads=1`
- GitHub `rust-and-cli` is currently failing before project compilation because the workflow cannot fetch the pinned private `sourcenetwork/defradb.rs` dependency after rewriting SSH to HTTPS (`Repository not found` / revision fetch fails).

PR #40, opened 2026-04-15
Fix ToolServiceRegistry schema and apply semantics

### LINKED ISSUE #28: Fix ToolServiceRegistry GraphQL tools relation join failure
While updating Amygdala fleet verification, querying the ToolServiceRegistry -> tools relation against a live raw defra-agent runtime failed with:

internal error: TypeJoinMany requires child side to have FK field. Child collection 'ToolServiceEntry' relation field 'tools' is missing expected FK field '_toolsID'.

Impact:
- amygdala-deploy verify-services had to stop querying the tools relation entirely and now only verifies registry rows
- direct GraphQL inspection of the advertised tool list is currently broken even though MCP routing itself can work

Expected behavior:
- either the ToolServiceRegistry -> tools relation should query successfully
- or the schema/model should be changed so the relation is not exposed in a broken state

This showed up on live runtimes on studio-1 on 2026-04-14 while verifying observability-mcp, coding-session-store, hf-data, and x-data through raw defra-agent instances.

### LINKED ISSUE #29: Clarify and normalize ToolServiceRegistry apply semantics
Amygdala moved ToolServiceRegistry ownership into checked-in desired state rendered through defra-agent config apply. That path works, but the current live-row semantics are still muddy.

What we observed on 2026-04-14:
- newly applied tool-service manifests create rows with normalized address fields and status=online
- live GraphQL still returns version=null and updated_at=null for those rows
- older rows with nullable address fields were treated as unchanged by diff/apply until we deleted and recreated them
- downstream tooling has to tolerate multiple live shapes for the same logical registry row

Questions the runtime should answer explicitly:
- which ToolServiceRegistry fields are desired-state inputs vs runtime-populated outputs?
- should config apply normalize existing rows when the desired shape changes?
- should version/updated_at be populated on apply-created rows, or are they intentionally runtime-only and nullable?

A clear contract here would simplify fleet tooling a lot. Right now Amygdala had to add null-tolerant verification and do a one-time delete/recreate of existing rows to converge live state.

jackzampolin:
## Summary
- Remove the persisted `ToolServiceRegistry.tools` relation from the checked-in schema so new schema installs no longer expose the broken `TypeJoinMany` path.
- Normalize ToolServiceRegistry desired-state/apply storage: missing or null address fields become empty strings, missing or empty `mcp_path` becomes `/mcp`, and runtime-owned fields stay out of desired diffs.
- Preserve raw live nulls during desired-state diffing so `config apply` rewrites older rows and converges, while export/import round trips emit normalized rows.
- Document the desired-state/runtime-owned ToolServiceRegistry contract.

## Issue Links
- Fixes #28, assuming there are no existing stores that need the old `ToolServiceRegistry` collection definition migrated in place. DefraDB schema migration/patch tooling is tracked separately in sourcenetwork/defradb.rs#859 for future hardening.
- Fixes #29.

## Tests
- `cargo test -p defra-agent-cli tool_service_registry -- --nocapture`
- `cargo test -p defra-agent-cli diff_marks_live_null_tool_service_storage_for_update -- --nocapture`
- `cargo test -p defra-agent --test document_config_bootstrap tool_service_registry_schema_does_not_expose_broken_tools_relation -- --nocapture`
- `cargo test -p defra-agent-cli --test cli_e2e config_apply_reconciles_tool_services_and_scheduled_tasks_end_to_end -- --nocapture`
- `cargo test -p defra-agent-cli --test cli_e2e config_export_import_round_trips_tool_services_and_scheduled_tasks -- --nocapture`
- `cargo test -p defra-agent --test document_config_bootstrap`
- `cargo test -p defra-agent --lib --tests`
- `cargo check --workspace`
- `git diff --check`

### DIFF
--- a/crates/defra-agent-cli/src/desired_state.rs
+++ b/crates/defra-agent-cli/src/desired_state.rs
@@ -135,8 +135,10 @@ pub(crate) struct DesiredInferenceProfile {
     pub(crate) deadline_duration_secs: Option<i64>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(deny_unknown_fields)]
+const DEFAULT_TOOL_SERVICE_MCP_PATH: &str = "/mcp";
+const TOOL_SERVICE_ADDRESS_FIELDS: &[&str] = &["hostname", "tailscale_ip", "lan_ip"];
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 pub(crate) struct DesiredToolServiceRegistry {
     pub(crate) service_id: String,
     pub(crate) display_name: Option<String>,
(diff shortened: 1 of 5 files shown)


jackzampolin:
Self-review notes:

- The apply/diff path uses `build_desired_state_live_bundle`, which intentionally does not normalize live ToolServiceRegistry rows before diffing. That means old live `null` address or `mcp_path` storage still shows up as drift and `config apply` rewrites it to the normalized storage shape.
- Export/import normalization is scoped to user-facing bundle output and mutation payloads, so clean round trips get stable empty-string address fields and `/mcp` defaults without hiding live drift during apply.
- The checked-in schema no longer exposes `ToolServiceRegistry.tools`, and the regression test asserts querying it fails validation instead of reaching `TypeJoinMany`.
- Known limitation: existing DefraDB stores that already created the old `ToolServiceRegistry` collection definition are not automatically patched by this PR. The current `defra-node` wrapper only exposes `add_schema`, which returns already-exists for existing collections rather than migrating them, so those stores may still need a schema migration/reset.

jackzampolin:
CI follow-up:

`rust-and-cli` failed before compiling this branch. The job cannot fetch the private git dependency `ssh://git@github.com/sourcenetwork/defradb.rs.git` after the workflow rewrites it to HTTPS; the log reports `Repository not found` and `revision fea013b9... not found` while resolving the `crypto` dependency.

I reran the exact failed command locally after the PR opened:

```
cargo test -p defra-agent --lib --tests
```

Local result: passed (`195` lib tests plus all defra-agent integration tests, with the expected one ignored live API test). This looks like CI dependency access rather than a branch regression.

PR #79, opened 2026-04-24
proofs: re-home trigger conformance onto ReachableUnder

### LINKED ISSUE #77: Re-home trigger/lifecycle conformance onto strengthened trigger trace boundary
## Context

PR #75 established trigger/lifecycle conformance theorems in `Proofs/Conformance/Triggers.lean`.
PR #76 tightens the trigger-engine trace boundary with:

- `FireIntent.WellFormed`
- `ReachableUnder`
- `WellFormedReachable`
- T2 lifted onto the strengthened trigger trace relation

That means the proof stack now has:

- a stronger admissible trigger trace boundary
- a lifecycle-trace conformance layer

but those two stories are not yet connected explicitly.

## Problem

The conformance theorems introduced in PR #75 still live over materialized requests and `RequestContext.Trace`; they do not explicitly mention the strengthened trigger trace boundary from PR #76.

In particular, the "real spec" still has a split surface:

- trigger-level invariants can now range over `WellFormedReachable`
- trigger/lifecycle conformance theorems do not yet state how they relate to that stronger trigger reachability relation

So the boundary tightening is in place, but the conformance layer has not been re-homed onto it where that is meaningful.

## Goal

Make the post-PR-76 theorem surface read cleanly end-to-end:

- admissible trigger traces start from well-formed inputs
- materialized requests embed coherently into the lifecycle model
- lifecycle traces preserve that coherence
- the public conformance story explicitly reflects the strengthened trigger trace boundary rather than relying on it only indirectly

## Suggested approach

- Review `Proofs/Conformance/Triggers.lean` after PR #76 lands.
- Identify which public theorems meaningfully benefit from a `WellFormedReachable` / `ReachableUnder` hypothesis.
- Re-home those results onto the strengthened relation where the hypothesis is actually load-bearing.
- Avoid adding ceremonial wrapper theorems if the stronger reachability relation does no work in the statement.

This is likely about theorems around:

- `materializedTriggerRequest_coherent_along_trace`
- the trigger-created request embedding story
- any future theorem that connects trigger reachability to lifecycle traces more directly

## Scope

- Mostly `Proofs/Conformance/Triggers.lean`
- Possibly small helper fallout in `Proofs/Triggers.lean`
- Lean-only; no Rust changes required

## Related

- #75 — trigger/lifecycle conformance theorem layer
- #76 — tighten trigger traces to well-formed inputs
- #71 — strengthen T2 to a pre-trace invariant

jackzampolin:
## Summary
- add step-level trigger/lifecycle conformance entry points over `ReachableUnder`
- specialize those entry points to `WellFormedReachable` so the public theorem surface now starts from the strengthened trigger trace boundary
- keep the original local conformance proofs intact and reuse them rather than duplicating the lifecycle-trace arguments

## Testing

Closes #77

### DIFF
--- a/crates/defra-agent/proofs/Proofs/Conformance/Triggers.lean
+++ b/crates/defra-agent/proofs/Proofs/Conformance/Triggers.lean
@@ -369,3 +369,128 @@ theorem materializedTriggerRequest_coherent_along_trace
   apply triggerLifecycleCoherent_preserved_by_lifecycle_trace
   · exact materializedTriggerRequest_has_claimed_embedding state intent seed inputs
   · exact h_trace
+
+/--
+Step-level conformance entry point over an admissibility-constrained trigger
+trace.
+
+If `state` already lies on a `ReachableUnder P` trigger trace, `intent`
+satisfies the same boundary predicate, and `dispatch` materializes a seed, then:
+
+* the next trigger state stays on the same admissible trace
+* the materialized seed/origin pair satisfies `consistentLineage`
+* the created request shape admits the canonical claimed lifecycle embedding
+-/
+theorem reachableUnder_dispatch_materializedTriggerRequest_conforms
+    (P : FireIntent → Prop)
+    (state : SystemState)
+    (snap : TriggerSnapshot)
+    (intent : FireIntent)
+    (seed : RequestSeed)
+    (inputs : ClaimedEmbeddingInputs)
+    (h_reach : ReachableUnder P state)
+    (h_intent : P intent)
+    (h_dispatch : dispatch snap intent = some seed) :
+    ReachableUnder P (dispatchStep state snap intent) ∧
+    consistentLineage seed (materializedTriggerRequest state intent seed).executionOrigin ∧
+    TriggerLifecycleCoherent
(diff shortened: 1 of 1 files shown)


jackzampolin:
I don't think T1/T3/T4 need strengthened-boundary restatements right now.

T2 was the theorem whose API shape was actually wrong under the stronger trace story: it quantified over reachable states but still asked callers to re-prove the key seriality fact as a post-state hypothesis. That made `ReachableUnder` / `WellFormedReachable` genuinely load-bearing for T2.

T1/T3/T4 are different:
- `T1_enabled_gate` is pointwise over `dispatch`; it is not a trace theorem.
- `T3_latest_only_convergence` is pointwise over `latestOnlyFireTransition`; again, no trigger-trace boundary hypothesis is missing from its statement.
- `T4_lineage_completeness` is a definitional characterization of `(seed, origin)` pairs, also pointwise.

So my bias is to stop here unless a concrete downstream caller shows up that actually needs a whole-trace entry point for one of those theorems. If that happens, we can add the wrapper then; until then, a `ReachableUnder` restatement for T1/T3/T4 would be ceremonial in the way #77 explicitly asked us to avoid.

PR #132, opened 2026-05-07
Add CommandPolicy generated conformance cases

jackzampolin:
## Summary
- add finite Lean CommandPolicy validation, sandbox, and env generated cases
- emit CommandPolicy case groups through the Lean conformance JSON and coverage ledger
- add Rust deserialization helpers plus generated-case tests against command policy, sandbox selection, and env filtering

## Verification
- lake build
- cargo test -p defra-agent --lib command -- --nocapture

### DIFF
--- a/crates/defra-agent/proofs/Proofs/CommandPolicy.lean
+++ b/crates/defra-agent/proofs/Proofs/CommandPolicy.lean
@@ -3,6 +3,7 @@ import Proofs.CommandPolicy.Validation
 import Proofs.CommandPolicy.Sandbox
 import Proofs.CommandPolicy.Env
 import Proofs.CommandPolicy.Theorems
+import Proofs.CommandPolicy.Cases
 
 /-!
 # Command Policy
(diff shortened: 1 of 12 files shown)


jackzampolin (commented):
Self-review completed. I checked the emitted CommandPolicy domains against the coverage ledger and Rust consumers, confirmed the generated validation cases cover forbidden-before-allowed, allowed-before-network/read-only, and network-before-read-only ordering, and verified the env cases exercise inherited, fallback, forced, dropped, and secret-marker paths. No blocking issues found.

jackzampolin:
Addressed the review cleanup in 562aae7:

- removed the duplicate macOS sandbox-exec existence check and threads the selected sandbox label into the workspace_write command builder
- added a generated forbidden-prefix case where the second configured forbidden prefix matches
- added a generated env case proving PAGER is forced to cat even when no parent PAGER exists
- updated pinned CommandPolicy case counts

Re-ran:
- lake build
- cargo test -p defra-agent --lib command -- --nocapture

PR #137, opened 2026-05-07
Add Lean ApplyReconcile contract cases

jackzampolin:
## Summary
- emit Lean-generated finite ApplyReconcile contract cases in the conformance JSON snapshot
- add Rust deserialization and an apply_conformance consumer that drives apply_model, retry/idempotence, referrer closure, live-only no-op, and production Collection ordering/unique-field helpers
- update the coverage ledger and state-machine conformance accounting for the new consumer

## Verification
- cargo test -p defra-agent --test apply_conformance
- cargo test -p defra-agent --test apply_property
- cargo fmt --all -- --check
- git diff --check
- rg -n "\b(sorry|admit|axiom)\b|PLACEHOLDER|todo!\(" crates/defra-agent/proofs/Proofs crates/defra-agent/src crates/defra-agent/tests

### DIFF
--- a/crates/defra-agent/proofs/Proofs/ApplyReconcile.lean
+++ b/crates/defra-agent/proofs/Proofs/ApplyReconcile.lean
@@ -6,9 +6,11 @@ import Proofs.ApplyReconcile.ApplyProperties
 import Proofs.ApplyReconcile.Prefix
 import Proofs.ApplyReconcile.RuntimeBridge
 import Proofs.ApplyReconcile.Convergence
+import Proofs.ApplyReconcile.ContractCases
 
 /-!
 # Apply-Reconcile Composition
 
-Barrel import for collection ordering, manifest diff/apply, runtime bridging, and T-Conv.
+Barrel import for collection ordering, manifest diff/apply, runtime bridging,
+contract cases, and T-Conv.
 -/
--- a/crates/defra-agent/proofs/Proofs/ApplyReconcile/ContractCases.lean
+++ b/crates/defra-agent/proofs/Proofs/ApplyReconcile/ContractCases.lean
@@ -224,6 +224,8 @@ def buildCase (scenario : ApplyReconcileScenario) : ApplyReconcileCase :=
   , expectedRetryDesired := sortedDocs retry
   , expectedRetryStepCount := retrySteps.length
   , expectedRediffStepCount := rediff.length
+  -- The list projection has no live-write constructor; Rust checks this
+  -- invariant against `apply_model::apply_all` using the emitted pre-live rows.
   , livePreserved := true
   , manifestRealizedAfter := manifestRealizedBool scenario.manifest after
   , retryConverges := desiredDocsEq retry after
--- a/crates/defra-agent/proofs/Proofs/ApplyReconcile/ContractCases.lean
+++ b/crates/defra-agent/proofs/Proofs/ApplyReconcile/ContractCases.lean
@@ -1,0 +1,387 @@
+import Proofs.ApplyReconcile.Collections
+import Proofs.Conformance.ContractTypes
+
+/-!
+# Apply/Reconcile Contract Cases
+
+Finite executable apply/reconcile witnesses emitted through
+`Proofs.Conformance.Contracts`.
+
+The proof model uses `Manifest.support : Finset DocRef`; `Finset.toList` is a
+proof-side representation and has no `lean --run` executable code. These
+contract rows therefore use a compact finite list projection over the same
+`DocRef`, `Collection.applyOrder`, and create/update step vocabulary. Rust
+consumes the emitted inputs and expected outcomes against the production-facing
+`apply_model`, so these rows are executable conformance cases rather than a
+second Rust-only table.
+-/
+
+namespace ApplyReconcile.ContractCases
+
+open Conformance.Contracts
+
+structure ContractDoc where
+  ref : DocRef
+  content : String
+  refs : List DocRef
+
+structure ContractLiveDoc where
+  ref : DocRef
+  content : String
+
+structure ContractStep where
+  action : String
+  target : DocRef
+  content : String
+  refs : List DocRef
+
+structure ApplyReconcileScenario where
+  name : String
+  manifest : List ContractDoc
+  preDesired : List ContractDoc
+  preLive : List ContractLiveDoc
+  prefixLen : Nat
+
+structure ApplyReconcileCase where
+  name : String
+  manifest : List ContractDoc
+  preDesired : List ContractDoc
+  preLive : List ContractLiveDoc
+  expectedCreate : List DocRef
+  expectedUpdate : List DocRef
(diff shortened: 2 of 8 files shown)


jackzampolin:
Self-review pass completed.

Checked:
- ApplyReconcile emitted cases cover the requested flagship rows, including a non-vacuous referrer-closure prefix.
- Rust consumer parses the existing Lean contract snapshot and compares Lean rows against apply_model diff/apply/prefix/retry/idempotence behavior.
- Coverage ledger has an executable consumer entry for ApplyReconcileCases.
- Remaining boundary is explicit in the PR: the defra-agent test does not execute CLI DB writes or mutate production manifests; it checks apply_model plus production Collection ordering/unique-field helpers.

jackzampolin:
## Review of PR #137 — Add Lean ApplyReconcile contract cases

### Overview

Wires Lean-emitted ApplyReconcile witnesses into the conformance JSON snapshot and consumes them from a new Rust test, `apply_conformance::generated_apply_reconcile_cases_drive_apply_model_and_production_ordering`. The Lean side uses a compact `List ContractDoc` projection (justified inline because the proof model uses `Finset DocRef` which is not `lean --run`–executable). The Rust consumer:
- decodes the cases via `LeanApplyReconcileCase`
- exercises `apply_model::{diff, apply_prefix, apply_all, retry_after_prefix, manifest_realized, prefix_referrers_closed, desired_references_closed}`
- additionally enforces production-facing `Collection::apply_order()` ordering and `Collection::unique_field()` non-emptiness on every case
- adds a row to `caseCoverage` and a count assertion in `state_machine_conformance.rs`

This mirrors the established pattern from PR #132 (CommandPolicy) and the existing RuntimeReconcile cases.

### Strengths

- **Clear scope statement** at the top of `ContractCases.lean` explains *why* a parallel list-based projection is acceptable rather than reusing the proof-side `Manifest.support : Finset DocRef`.
- **Production-facing asserts**: `doc_ref_from_lean` and `doc_ref_from_parts` assert `unique_field()` non-emptiness for every Lean-emitted DocRef, exercising the production helper as the PR description claims.
- **Ordering check** (`assert_steps_are_production_apply_ordered`) is independent of the bucket-equality check — guards against `diff` regressing to declaration order.
- **Reference closure assert** (`assert_manifest_refs_point_to_lower_apply_ranks`) protects WellFormed manifests at the test-fixture layer, catching drift if a future case author writes a non-prefix-safe case.
- **Coverage ledger + count tripwire** (`= 6` in `state_machine_conformance`) ensures the snapshot can't silently lose cases.

### Issues / Suggestions

**1. Semantic divergence in `prefixReferrersClosed` between Lean and Rust** (minor, worth a comment)

Lean (`ContractCases.lean:194-200`):
```
match lookupDoc? desired step.target with
| some doc => doc.refs.all fun ref => containsDoc desired ref
| none => false
```

Rust (`apply_model.rs:196-204`):
```rust
prefix.iter().all(|step| {
    l.desired
        .get(step.target())
        .into_iter()
        .flat_map(references_of)
        .all(|r| l.desired.contains_key(&r))
})
```

If `step.target()` isn't in `desired`, **Lean returns `false`** but **Rust returns `true`** (vacuously, since `Option::into_iter()` on `None` is empty and `.all` on empty is true). In every test case the target is always present (since `apply_one` inserts it), so the test passes — but the model definitions disagree on the unreachable branch. Either a Rust `expect()` on the lookup or a Lean comment ("none branch is unreachable; matches Rust's vacuous-true via Option::into_iter") would close this.

**2. `livePreserved` is hardcoded `true` in `buildCase`** (`ContractCases.lean:251`)

The list-based contract model doesn't track live mutation, so this field always serializes `true`. The Rust assertion `after.live == live.live` against `case.live_preserved` still does useful work (verifies `apply_model` doesn't touch `live`), but the JSON field carries no signal from Lean. Consider either dropping the field or adding a comment that it's a hardcoded invariant from the model rather than computed.

**3. Parallel implementation in Lean**

`ContractCases.lean` re-implements `diffSteps`, `applyOne`, `applyAll`, `manifestRealized`, etc. with `List ContractDoc` instead of using the proof-side `Proofs.ApplyReconcile.*` definitions. Drift risk is real: future changes to the proof model may not flow into the contract cases. The header comment acknowledges this, and Rust validates against `apply_model` (which is itself anchored to the proof model). Worth flagging as a follow-up: a Lean-side equivalence theorem between the list projection and the Finset-backed model would close the loop.

**4. Style nit — `docRefBEq`** (`ContractCases.lean:105-106`)

```
def docRefBEq (a b : DocRef) : Bool :=
  if a = b then true else false
```

Since `DocRef` derives `DecidableEq`, this is equivalent to `decide (a = b)` or, with a `BEq` instance, `a == b`. Not a bug.

**5. Required-name list duplication** (`apply_conformance.rs:711-721`)

The hardcoded list of six required case names mirrors the Lean fixture. Renaming a case in Lean breaks the Rust assertion — that's the intended tripwire — but it would be cleaner to pull this from a small Lean-emitted "case_names" array, or to reuse `cases.iter().map(|c| &c.name)` and assert non-emptiness + uniqueness. Currently fine for six cases.

**6. Test function length**

`generated_apply_reconcile_cases_drive_apply_model_and_production_ordering` is ~140 lines with many distinct assertion clusters (buckets, steps, prefix, retry, rediff). Splitting into helper functions per cluster would improve readability without changing semantics.

### Risk Assessment

- **Correctness**: looks sound. The Rust test compares the same logical invariants (bucket membership as sets, step order as list, prefix/retry/rediff as `BTreeMap` and `Vec` equalities). The hardcoded `livePreserved = true` is benign because `apply_model` truly preserves live.
- **Performance**: negligible — 6 fixed cases, small docs.
- **Security**: no concern — JSON construction goes through `jsonString` escaping; no user input.
- **Test coverage**: improved. Six scenarios cover empty manifest, ordering, update, live-only, prefix retry/idempotence, and 6-doc referrer closure with `prefixLen=4`.

### Verdict

LGTM with minor follow-ups. The `prefixReferrersClosed` divergence is the only thing I'd want addressed (or explicitly documented) before merge — everything else is cleanup.

jackzampolin:
Addressed review comment https://github.com/sourcenetwork/defra-agent/pull/137#issuecomment-4399647397 in follow-up commit 0cd5c54.

Changes:
- `apply_model::prefix_referrers_closed` now returns `false` if a prefix step target is absent from the desired projection, matching the Lean contract case semantics instead of vacuous `Option::into_iter()` truth.
- Added a Lean comment explaining that `livePreserved := true` is a structural invariant of the list projection and is checked concretely by Rust against `apply_model::apply_all`.

Verification:
- `cd crates/defra-agent/proofs && lake build Proofs.Conformance.Contracts`
- `cargo test -p defra-agent --test apply_conformance`
- `cargo test -p defra-agent --test apply_property`
- `git diff --check`
- hygiene scan for `sorry/admit/axiom/PLACEHOLDER/todo!(`

PR #138, opened 2026-05-07
Split ClientShell contract coverage ownership

jackzampolin:
## Summary
- split generated ClientShell contract output into frontend and desktop case groups
- keep TypeScript chat projection coverage on all 15 frontend cases
- make the Tauri session snapshot bridge consume every selected-session desktop case, including absent snapshots
- fix the stale observation case metadata to report awaiting_stale_request_observation

## Verification
- cargo test -p defra-agent --test state_machine_conformance generated_client_shell
- cargo test -p defra-agent --test state_machine_conformance lean_executable_contracts_cover_initial_domains
- cargo test -p defra-agent --test state_machine_conformance lean_contract_coverage_ledger_accounts_for_every_emitted_domain
- cargo test -p defra-agent-desktop-core
- cargo test -p defra-agent-desktop-tauri session_snapshot_projection_consumes_generated_client_shell_contract_cases
- bun test apps/desktop-tauri/src/lib/chat-shell.test.ts
- cargo fmt --all -- --check
- git diff --check
- hygiene scan for sorry/admit/axiom/PLACEHOLDER/todo!( (changed files clean; full scan only reports existing CLAUDE.md text saying zero sorrys)


jackzampolin:
Self-review pass completed. Checked the PR diff for split ownership correctness, stale ClientShell naming, generated-case coverage, and skip paths. No follow-up issues found: TS still asserts all 15 frontend rows, the Tauri bridge loops over all 12 desktop rows and asserts snapshot presence, and the ledger has separate frontend/desktop consumers.

jackzampolin:
## Review: PR #138 — Split ClientShell contract coverage ownership

### Overview
Splits the single `client_shell_cases` Lean emission into two lists:
- `frontend_client_shell_cases` (all 15) — consumed by `projectChatShell` in `apps/desktop-tauri/src/lib/chat-shell.test.ts`.
- `desktop_client_shell_cases` (12, filtered by `desktopSelectedSessionId.isSome`) — consumed by `build_session_snapshot_from_store` via `session_snapshot_projection_consumes_generated_client_shell_contract_cases`.

Adds `desktopSnapshotPresent` to the contract case so the desktop bridge can also exercise the selected-but-absent case (`blocked_submit_session_absent`). Fixes a copy-paste metadata bug on `awaiting_stale_request_observation` (its `property` was reading `awaiting_matching_request_observation`) and pins it with a new conformance assertion. Updates `CoverageLedger`, `Boundaries.lean`, the proofs README, and `client-state-machine.md` to reflect the split.

### What's good
- **Correctness of the split.** `desktopClientShellCases := clientShellCases.filter (·.desktopSelectedSessionId.isSome)` lines up with what the desktop snapshot bridge cares about: cases without a selected session never reach `build_session_snapshot_from_store`. The previous single-test-per-domain shape was forcing the desktop test to filter by name (3 hand-picked cases), which was strictly weaker coverage.
- **`desktopSnapshotPresent` is load-bearing.** Without it, the desktop test couldn't distinguish \"selected and observed\" from \"selected but absent\" — `blocked_submit_session_absent` (selected sid1, empty store) would have collided with the previous `.expect(\"session snapshot ...\")`. The new `assert_eq!(snapshot.is_some(), case.desktop_snapshot_present, ...)` correctly handles both.
- **Metadata fix is properly anchored.** The `awaiting_stale_request_observation` property field was clearly a copy-paste; the new `assert_eq!(stale.property.as_str(), \"awaiting_stale_request_observation\")` in `generated_client_shell_cases_cover_shell_projection_contracts` will catch a regression.
- **Coverage ledger discipline.** `frontend_client_shell_cases` and `desktop_client_shell_cases` are both registered with named consumers, and `Boundaries.lean` now explicitly documents the assignment rule (\"frontend list when only React shell state; desktop list when the Rust session-snapshot bridge is exercised\"). The list of recognized ledger keys in `state_machine_conformance.rs` is updated to match.
- **Drift detectors on both sides.** The TS test throws if `frontend_client_shell_case_count !== frontend_client_shell_cases.length`, the bun test pins `toHaveLength(15)`, and Rust pins `frontend == 15` and `desktop == 12`. Adding a new case will now fail in three independent places, which is the right tripwire density for a generated contract surface.

### Concerns / suggestions

**1. Latent coupling between `frontend_conversation_present` and `desktop_snapshot_present` in `client_shell_contract_store`.** The store builder gates the session row on `desktop_snapshot_present` but the conversation row on `frontend_conversation_present`. Today these can't diverge — `frontendConversationPresent := conversationPresent && obs.isSome`, `desktopSnapshotPresent := selectedObs.isSome`, and `obs == selectedObs` because `desktopSelectedSessionId := frontendLocal.selection.session`. But that invariant is implicit. If a future change decouples desktop selection from frontend selection, this builder would silently emit a conversation row tied to a non-existent session. Worth either:
  - asserting `case.frontend_conversation_present.implies(case.desktop_snapshot_present)` at the top of `client_shell_contract_store`, or
  - gating the conversation row on `case.desktop_snapshot_present && case.frontend_conversation_present`.

  Either keeps the test honest if the Lean model later splits the two selections.

**2. Redundant initializer in `client_shell_contract_store`.** `apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_state.rs:544-550` writes:
```rust
let mut rows = ClientStoreRows {
    sessions: Vec::new(),
    conversations: Vec::new(),
    requests: Vec::new(),
    responses: Vec::new(),
    ..ClientStoreRows::default()
};
```
Since `Vec::default()` is `Vec::new()`, the four explicit fields are redundant with `..default()` — `let mut rows = ClientStoreRows::default();` is equivalent. Minor; the verbose form does document which fields the test mutates downstream.

**3. \"Rust consumer\" terminology in `Boundaries.lean`.** The Boundaries doc still says \"Rust asserts that every emitted vocabulary, state machine ... appears in that ledger exactly once.\" The README was updated to say \"Rust/TypeScript consumer,\" but Boundaries is still Rust-centric while now pointing the frontend ledger entry at a bun test. Consider tightening the wording in `Boundaries.lean:139-141` to match the README's \"Rust/TypeScript\" framing for consistency.

**4. Hard-coded `30000` ms timeout.** Reasonable given Lean build cost, but a named constant or comment explaining \"covers `lake build` cold-start\" would age better than a bare magic number.

### Risks
- Low. The change is additive on the Lean side (existing fields preserved, new field added with explicit JSON emission), the JSON schema bump is reflected in the Rust deserializer (`lean_vocab_test.rs`) and the TS type, and the verification matrix in the PR description covers all three consumers (Lean build, Rust conformance, desktop core, Tauri snapshot, bun TS test).
- The 15 / 12 cardinality assertions are duplicated across three files; adding a case requires touching all three, but the failure modes are loud and self-explanatory.

### Test coverage
Solid. Frontend test now drives all 15 cases through `projectChatShell`; desktop test drives all 12 selected-session cases through the snapshot bridge including the absent-snapshot path; conformance ledger drift test will reject any new emitted domain that isn't registered.

LGTM with the suggestion to consider the implicit invariant note (#1) for future-proofing.

jackzampolin:
Addressed the review suggestions in follow-up commit 00814bc:\n\n- added an explicit assertion that generated conversation rows require a desktop snapshot observation in the Tauri bridge store builder\n- simplified the test row initializer to ClientStoreRows::default()\n- updated Boundaries.lean wording to use Rust/TypeScript conformance coverage\n- replaced the bare Bun timeout with a named constant and comment\n\nRe-ran:\n- cd crates/defra-agent/proofs && lake build\n- bun test apps/desktop-tauri/src/lib/chat-shell.test.ts\n- cargo test -p defra-agent-desktop-tauri session_snapshot_projection_consumes_generated_client_shell_contract_cases\n- cargo fmt --all -- --check\n- git diff --check\n- changed-file hygiene scan for sorry/admit/axiom/PLACEHOLDER/todo!(

PR #139, opened 2026-05-07
Extend CommandPolicy read-only safety contracts

jackzampolin:
## Summary
- model command-family read-only argv safety checks in Lean for git, sed, find, rg, curl, launchctl, tailscale, and sudo
- emit 35 additional Lean-generated CommandPolicy cases covering safe and unsafe read-only shapes
- extend Rust conformance tests to consume new denial details and assert the generated safety matrix drives validate_command_policy

## Verification
- cargo test -p defra-agent generated_command -- --nocapture
- cargo test -p defra-agent toolset::tests::read_only_bash
- cargo fmt --all -- --check
- git diff --check
- rg hygiene scan for sorry/admit/axiom/PLACEHOLDER/todo!(; only pre-existing CLAUDE.md prose mention of zero sorrys)

### DIFF
--- a/crates/defra-agent/proofs/Proofs/CommandPolicy/Cases.lean
+++ b/crates/defra-agent/proofs/Proofs/CommandPolicy/Cases.lean
@@ -98,7 +98,46 @@ def envCase
   , expected := filteredEnv (envInput envKey inputPresent) envKey
   }
 
-def commandPolicyCases : List CommandPolicyCase :=
+def defaultReadOnlyAllowlist : List String :=
+  [ "pwd"
+  , "ls"
+  , "find"
+  , "cat"
+  , "head"
+  , "tail"
+  , "sed"
+  , "grep"
+  , "rg"
+  , "wc"
+  , "stat"
+  , "file"
+  , "git"
+  , "date"
+  , "hostname"
+  , "uptime"
+  , "df"
+  , "vm_stat"
+  , "ps"
+  , "lsof"
+  , "curl"
+  , "launchctl"
+  , "tailscale"
+  , "sudo"
+  ]
+
+def readOnlySafetyPolicy : Policy :=
+  commandPolicy .readOnly [] [] .inherit defaultReadOnlyAllowlist
+
+def readOnlySafetyCase
+    (name command lookupCommand : String)
+    (args : List String) : CommandPolicyCase :=
+  validationCase
+    name
+    "read_only_argv_safety"
+    readOnlySafetyPolicy
+    (commandRequest command lookupCommand args)
+
+def commandPolicyOrderingCases : List CommandPolicyCase :=
   [ validationCase
       "forbidden_prefix_wins_over_allowed_prefix_order"
       "forbidden_prefix"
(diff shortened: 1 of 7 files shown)


jackzampolin (commented):
Self-review completed. Checked the pushed diff for CommandPolicy boundary clarity, generated read-only safety case coverage across git/sed/find/rg/curl/launchctl/tailscale/sudo, denial-field serialization/decoding, and Rust test routing through validate_command_policy. No follow-up issues found.

jackzampolin:
## Review: PR #139 — Extend CommandPolicy read-only safety contracts

### Overview
This PR is **spec-augmenting, not behavior-changing**. The Rust validators in `crates/defra-agent/src/toolset/shared/command.rs` (`validate_git_args`, `validate_curl_args`, `validate_sed`-style branches in `validate_read_only_command_inner`, etc.) already enforce these read-only argv guards. The diff adds:

- A Lean model of those argv filters (`validateReadOnlyArgs` in `Validation.lean`) plus four new `DenialReason` variants (`readOnlyArgumentNotAllowed`, `readOnlySubcommandRequired`, `readOnlySubcommandNotAllowlisted`, `readOnlyUrlRequired`) and accessors in `Types.lean`.
- 35 new emitted `CommandPolicyCase`s exercising allow/deny shapes for git, sed, find, rg, curl, launchctl, tailscale, and sudo.
- Rust-side fixture additions (`denied_argument`, `denied_subcommand`) and an expanded `assert_command_denial_matches` arm that maps Lean denial reasons to the corresponding Rust `bail!` strings.
- A `generated_command_policy_cases_cover_read_only_safety_matrix` presence test plus a count bump (10 → 45) in `state_machine_conformance.rs`.

The conformance loop (`generated_command_policy_cases_match_rust_validation`) actually pipes every Lean case through `validate_command_policy` and asserts the decision matches, so allow-cases are enforced as `result.is_ok()` rather than just by error-string substring — this is the right shape.

### Strengths
- **Correctness alignment is verifiable.** I spot-checked tricky cases against Rust:
  - `git -c core.pager=cat status`: short-circuited by `gitGlobalOptionDenied` before reaching `findGitSubcommand` in both impls. ✓
  - `git branch -D main`: `gitReadOnlyFlagDenied("-D")` is false, then `validateGitBranchArgs` rejects `-D` (not in allowlist, no `--format=` prefix). Mirrors Rust's `validate_git_branch_args`. ✓
  - `sudo /usr/bin/launchctl print …`: `sudoCommandName` resolves to "launchctl", but the inner branch only allows command literal `/bin/launchctl`, so it falls through to `readOnlyArgumentNotAllowed "sudo" "/usr/bin/launchctl"`. Matches Rust's `bail!(\"sudo launchctl must use the absolute /bin/launchctl path\")`. ✓
  - `curl` denial without `http(s)://` URL maps to `readOnlyUrlRequired`, which the Rust assertion correctly cross-checks against the literal `\"requires an http:// or https:// URL\"` string. ✓
- **Honest scope disclaimer.** The new docstring in `Validation.lean` explicitly says: *\"The command-family filters prove only local argv policy. They do not claim external binary read-only semantics or host/kernel sandbox enforcement.\"* That's the right framing — these are syntactic guards, not security proofs.
- **Coverage selection is representative.** Cases hit the failure-class boundaries (config-env, output redirect, in-place edit, exec, post body, missing URL, missing subcommand, sudo path mismatch) without becoming an exhaustive arg fuzzer.
- **No `sorry`/`admit` introduced** (PR description's hygiene scan, which I'd verify in CI).

### Minor suggestions (non-blocking)
1. **Style: double-negative in `validateGitBranchArgs`.** `firstArgWhere (fun arg => if gitBranchArgAllowed arg then false else true) args` reads awkwardly; `firstArgWhere (fun arg => !gitBranchArgAllowed arg) args` (or `decide ¬ …` if `Bool` coercion is annoying) is the same and easier to scan. `Validation.lean` line ~205.
2. **Redundant disjunction in `sedArgumentDenied`.** `stringMatches arg \"-i\" || ... || argStartsWith arg \"-i\"` — the first disjunct is subsumed by the third (`startsWith` is reflexive on equal strings). Removing the redundant clause makes the parallel with `--in-place` cleaner. Same in `gitGlobalOptionDenied` for `-c`/`-C` (the bare-arg matches in `stringIn` are subsumed by `argStartsWith`).
3. **`gitGlobalOptionDenied` is slightly looser than Rust.** Rust gates the prefix branch on `arg.len() > 2`; Lean does not. This is fine *as observed* — bare `-c`/`-C` are caught by the `stringIn` branch in Lean, so the resulting decision agrees on every input — but the asymmetry is worth a one-line comment so a future reader doesn't \"tighten\" Lean to match Rust verbatim and accidentally drop a coverage case.
4. **Pre-existing gap, not in scope.** Both Rust and Lean miss `sed --in-place=BACKUP` (the `=SUFFIX` form), because `\"--in-place=BACKUP\".starts_with(\"-i\")` is false and the equality check is `--in-place` exact. The Lean spec faithfully mirrors the Rust gap, which is the right call for a conformance contract — but worth filing a follow-up issue against the Rust validator (and adding an `--in-place=` denial case once fixed).
5. **The matrix presence test could fold in `denied_argument`/`denied_subcommand` shape assertions.** Today it only asserts emitted *names*. A stronger version would assert that, e.g., every name ending in `_denies` has either a `denied_argument` or `denied_subcommand` populated — catches accidental drift if a future spec edit replaces a `readOnlyArgumentNotAllowed` with `readOnlyCommandNotAllowlisted`. Optional.

### Risks
- **Allowlist scope.** `defaultReadOnlyAllowlist` in `Cases.lean` is a fixture-local allowlist for the conformance witness, *not* the runtime allowlist sourced from `BackendDefinition.command_execution.read_only_allowlist`. That's the correct shape (the spec models the validator, not the deployment config), but reviewers should not read this list as a runtime policy change.
- **Lean spec is `Bool`-decidable executable, not theorem-form.** No safety/liveness theorems are proved about argv filtering — these cases are conformance witnesses, evaluated by the `generated_command_policy_cases_match_rust_validation` driver. That's a fine choice for argv string filters (where exhaustive inductive proofs would be brittle and low-value), but the PR description's framing as \"contracts\" is accurate; readers shouldn't expect lemmas like `theorem read_only_blocks_writes`.

### Verdict
**Approve.** Spec faithfully tracks the Rust impl on the 35 representative inputs, the disclaimer about scope is appropriately honest, and the conformance driver enforces decision-equality (not just error-substring matching). The minor style notes above can land in a follow-up.

jackzampolin:
Addressed the review notes in follow-up commit 4f8cd45:

- simplified the Lean git branch denial predicate
- removed the redundant exact `sed -i` disjunct while preserving current Rust-equivalent behavior
- added a comment documenting the intentional git global-option prefix equivalence
- strengthened `generated_command_policy_cases_cover_read_only_safety_matrix` to assert allow/deny payload shape, not only emitted names

Filed the pre-existing production validator gap for `sed --in-place=` separately: #141.

Re-ran:
- `cd crates/defra-agent/proofs && lake build`
- `cargo test -p defra-agent generated_command -- --nocapture`
- `cargo test -p defra-agent toolset::tests::read_only_bash`
- `cargo test -p defra-agent --test state_machine_conformance`
- `git diff --check`
- hygiene scan for `sorry/admit/axiom/PLACEHOLDER/todo!(`; only pre-existing CLAUDE.md prose mention remains

Issue #281, opened 2026-05-20
Feature matrix: apply-reconcile operatorUi preview row


jackzampolin:
Deferred placeholder `#TBD-ui-apply-preview` from [feature matrix design](../docs/superpowers/specs/2026-05-20-feature-matrix-design.md) §3 row `apply-reconcile`.

Add a desktop UI surface that previews a `defra-agent-cli config apply` against a manifest before execution (dry-run + diff). Tag the new consumer under feature `apply-reconcile`, surface `operatorUi`. Coordinates with #265 design.

Related: #264, #265, #228 (transactional apply).

jackzampolin:
Closing as not-planned. Keeping `apply-reconcile` CLI-only.

Operators authoring desired-state manifests already live in the CLI, and `defra-agent-cli config diff` + `config apply` already emit the structured JSON (`DesiredStateDiffReport` / `ConfigApplyReport`) needed to gate an apply. A desktop preview panel would be a second surface to maintain without earning its keep — the deploy audience for the desktop isn't config authors.

Implication for the feature matrix (`docs/superpowers/specs/2026-05-20-feature-matrix-design.md` §3): the `apply-reconcile` row drops `operatorUi (#281)` from `deferred` rather than promoting it. When the matrix implementation lands (#264) the row reads `required: [operatorCli], deferred: []`. Coordinating-issue references (#265, #228, #246) are unaffected — they're CLI-side concerns.

PR #323 (the standalone HTML prototype built while making this decision) is being closed alongside. The prototype file remains on the unmerged `design/issue-281-apply-preview-prototype` branch as design reference if anyone wants to revisit later.

Issue #57, opened 2026-04-21
Model delete semantics when live-only removal is added


jackzampolin:
The `ApplyReconcile` Lean model has no `delete` constructor and T-Conv is scoped accordingly. When the CLI gains the ability to remove `live_only` documents, extend the model with `ApplyStep.delete` and prove T-Delete-safety: delete is only permitted when no live document references the target. Tracker only; no immediate work. Spec: `docs/superpowers/specs/2026-04-14-apply-reconcile-lean.md`, tracker: #53.

jackzampolin:
Lean verification sweep update from PR #365: audited ApplyReconcile delete semantics and left this tracker open. No delete-behavior change was made in the sweep because live-only document removal still needs a dedicated semantics decision and implementation.

jackzampolin:
**Now blocking #376** (config-apply prune, #364). PR #376 implements live-only prune = delete semantics, but ships outside the proven apply boundary (no `ApplyStep.delete`, fence at `config_import.rs:885` only covers create/update). This issue is the prerequisite: add `ApplyStep.delete` + prove **T-Delete-safety** (delete only when no live doc references the target) + extend the apply conformance fence to cover prune. Completing #57 is the path to landing #376.

PR #323, opened 2026-05-20
UI prototype: apply-reconcile preview panel (#281)

jackzampolin:
## Summary

- Self-contained HTML+CSS+JS prototype of the apply-reconcile preview panel at `docs/ui-prototypes/panel-281-apply-preview.html` — covers the deferred `(apply-reconcile, operatorUi)` slot the feature matrix tracks via issue #281.
- Renders the structured shape a Tauri `preview_apply(manifest_path) -> ApplyPreview` would return, mirroring today's `DesiredStateDiffReport` / `DesiredStateCollectionDiff` (`crates/defra-agent-cli/src/desired_state/mod.rs:256-379`) plus per-field diffs for updates and the per-collection `apply_order` rank from `ApplyReconcile.Collections.applyOrder` (`Proofs/ApplyReconcile/Collections.lean:31-40`).
- Five inline mock variations: empty, small (3 adds), large (50 adds + 30 updates + 100 no-ops), all-no-op, and a validation-error state. Estimated runtime uses ~5 ms / mutation per the 2026-05-20 defradb-tx-idle-timeout audit. Live-only documents render in a red-tinted reserved banner so the future delete pass (#57) slots in without re-laying out the panel.
- Atomicity block surfaces `expected_external_state_after_abort` (#246) — readers see exactly the pre-apply live state on abort because the apply path commits as a single DefraDB transaction.

Prototype only — no Tauri bridge, no real apply, no build tooling. The prototype file is the entire deliverable.

## Test plan

- [ ] Open `docs/ui-prototypes/panel-281-apply-preview.html` in a browser — no console errors.
- [ ] Cycle the **Mock dataset** dropdown through every variation (small, large, all no-op, empty, validation error) — each renders cleanly.
- [ ] Large dataset is legible: each collection card scrolls inside its 560 px cap; the 100-row no-op group folds to 5 rows with a Show-more affordance.
- [ ] Per-collection cards collapse/expand via click and via Enter / Space (focus is keyboard-reachable, `aria-expanded` flips).
- [ ] Apply / Dry-run buttons log entries to the action log; Apply is disabled in error and empty states and shows "Apply (no-op)" when the manifest matches state.
- [ ] Design-notes details block toggles open and reads cleanly.

### DIFF
--- a/docs/ui-prototypes/panel-281-apply-preview.html
+++ b/docs/ui-prototypes/panel-281-apply-preview.html
@@ -1,0 +1,1499 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>Apply preview · panel-281</title>
+<style>
+  :root {
+    color-scheme: dark;
+    --font-sans: "Inter", "SF Pro Text", system-ui, sans-serif;
+    --font-heading: "Funnel Display", "Inter", "SF Pro Display", system-ui, sans-serif;
+    --font-mono: "SF Mono", "JetBrains Mono", ui-monospace, monospace;
+
+    --source-black: #000000;
+    --source-green: #06b250;
+    --source-dark-green: #1c2724;
+    --source-white: #ffffff;
+    --source-blue: #10cbff;
+    --source-orange: #ffa011;
+    --source-green-rgb: 6 178 80;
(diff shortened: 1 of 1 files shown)


jackzampolin:
Closing alongside #281 (not-planned). `apply-reconcile` stays CLI-only.

The prototype file remains on this branch as design reference; not deleting the branch in case anyone wants to revisit later. See sourcenetwork/defra-agent#281 for the rationale.

PR #333, opened 2026-05-26
Add experimental Codex TUI shim

jackzampolin:
## Summary
- add `defra-agent codex-shim`, an experimental WebSocket endpoint shaped like the Codex TUI `--remote` protocol
- advertise minimal account/config/model metadata and implement synthetic thread/turn lifecycle notifications
- stream an echo assistant message as the current placeholder for future DEFRA GraphQL-backed turns

## Investigation notes
- The feasible integration point is the Codex app-server protocol: the TUI can be reused if DEFRA presents enough of that JSON-RPC-over-WebSocket surface.
- Similar projects I found mostly build custom clients around Codex app-server rather than using the stock Codex TUI for a non-Codex runtime.
- CodeHelm manages a local Codex App Server and asks users to connect Codex with `codex --remote <ws-url>`.
- Remodex bridges an iOS client to `codex app-server` or an existing Codex WebSocket endpoint; its README calls out that Codex.app catches up from persisted sessions rather than acting as a true live mirror.
- OpenAI issue #21551 points at multi-subscriber live event fanout as the missing primitive for richer peer-client co-presence.

## Caveats
- Stock `codex-cli 0.131.0` showed local onboarding before opening the remote connection when `CODEX_HOME` was empty. The command help now calls this out.
- This commit does not execute real DEFRA turns yet; it validates the shim shape and response stream with an echo implementation.
- Config, permissions, plugins, MCP, and account data are deliberately stubbed. Next pass should map DEFRA permissions/status/error surfaces into the Codex protocol.

## Testing
- `cargo test -p defra-agent-cli --bin defra-agent codex_shim`
- `cargo run -p defra-agent-cli -- codex-shim --help`
- `git diff --check`

`cargo fmt --check` is not clean for the whole workspace due pre-existing formatting drift outside this patch; touched Rust files were formatted directly with `rustfmt`.


jackzampolin:
Follow-up spike commit `7c3b5047` switches the shim from stringly JSON method dispatch to the pinned `codex-app-server-protocol` crate.

What changed:
- incoming WebSocket payloads deserialize through Codex `JSONRPCMessage` / `JSONRPCRequest` / `ClientRequest`
- outgoing envelopes use Codex `JSONRPCResponse`, `JSONRPCError`, and typed `ServerNotification`
- common responses are either constructed as Codex protocol structs or validated through the relevant Codex response type before sending
- config response field names now follow the protocol schema (`model_provider`, `approval_policy`, `sandbox_mode`) instead of the earlier hand-written camelCase guess

Tradeoff observed:
- this gives us much better wire-compatibility pressure, but the protocol crate is not tiny. It pulls Codex protocol/support crates plus schema/export dependencies; first `cargo run -p defra-agent-cli -- codex-shim --help` took about 6 minutes locally after the dependency change.
- Cargo initially selected `allocative 0.3.6`, which broke `starlark_map 0.13.0`; `Cargo.lock` now pins `allocative 0.3.4` via `cargo update -p allocative --precise 0.3.4`.

Verification:
- `cargo test -p defra-agent-cli --bin defra-agent codex_shim`
- `cargo run -p defra-agent-cli -- codex-shim --help`
- `git diff --check`

jackzampolin:
Codex shim compatibility gaps from the latest live pass:

- `turn/steer`: highest priority. Stock Codex sends this when the user submits input while a turn is active. The Lean adapter model now has an accepted-steer witness, but the shim still needs to map it onto DEFRA session/request primitives without emitting a second `turn/started`.
- Thread history/resume: `thread/list`, `thread/read`, loaded threads, resume/fork/rollback/compact are mostly stubbed. Fresh sessions work; restart/resume/replay flows will expose this.
- Slash-command API surface: richer TUI commands such as `review/start`, `thread/shellCommand`, rename/goal/settings/model flows, and fs methods need either real backing behavior or deliberate compatibility responses.
- Background execution: current projection covers DEFRA background tools as Codex command items, but not the full process-style lifecycle yet: incremental stdout/stderr, foregrounding, reconnect recovery, command actions, and durable cleanup semantics.
- Non-text input: `Image`, `LocalImage`, `Skill`, and `Mention` inputs are currently dropped by the shim text extractor, so `@filename`, pasted images, and skills do not yet match stock Codex.
- Config/permissions/model changes: config writes currently return OK but do not really mutate DEFRA backend/tool policy, so `/model`, `/permissions`, and related UI state can drift from actual runtime behavior.

Live status before this note: all six ignored Codex shim live tests passed against `MiniMax-M2.7-NVFP4` on the configured OpenAI-compatible endpoint.

Issue #369, opened 2026-06-02
Shared agent-schema crate (single source of truth for agent collection schemas)


jackzampolin:
## Problem

defra-agent's protocol schemas (`crates/defra-agent-protocol/schemas/agent/*.graphql`) are the de-facto contract for the agent collections — `AgentRequest`, `AgentResponse`, `AgentMessage`, `AgentToolCall`, `AgentSession`, `AgentConversation`, `AgentToolResult`, `CompactionEntry`. There is no shareable source of truth, so document-peer consumers re-declare them.

Concretely, **Amygdala hand-vendors copies** of these schemas (`crates/amygdala-schemas/schemas/amygdala/agent/*.graphql`, registered via `amygdala-schemas/src/lib.rs:44`) and keeps them in sync manually — a drift hazard. Amygdala couples to defra-agent purely through this collection contract (no code dependency), so when the schema moves, the two silently diverge.

## Proposal

Extract the agent collection schemas into a lightweight shareable artifact (a schema-only crate, or a published `.graphql` bundle) that both defra-agent and external consumers depend on as the single source of truth.

**Constraint:** Amygdala deliberately removed its `defra-agent` crate dependency (to escape the openai/codex transitive baggage and a second `defradb.rs` pin). The shared artifact must therefore be **schemas only, no runtime deps** — otherwise consumers won't adopt it.

## Provenance
Surfaced by the #338 observability rescope / Amygdala consumer analysis (2026-06-02).

Issue #338, opened 2026-05-28
Agent Observability Tools: Design for Agent-Inside Visibility


jackzampolin:
## Problem

The defra-agent harness is built for reliability (Lean proofs, admission control, persistence) but not designed for **agent UX**. As an agent running inside the system, I have limited visibility into the infrastructure I depend on.

Current gaps from internal perspective:
1. **No /self endpoint** — can't discover my own identity, behavior config, DID
2. **No direct DefraDB query tool** — must construct raw GraphQL, no way to introspect control plane
3. **No system-wide fleet view** — operating in a silo, can't see other agents or pending requests across the fleet
4. **No memory/persistence tool** — can't save notes or context across sessions
5. **No MCP pool introspection** — configured MCP connections are opaque, can't see health or add new ones
6. **No context budget indicator** — compaction happens invisibly, no "context remaining" signal
7. **No session history visibility** — can't see past sessions or compaction history

### 1. Agent Identity Endpoint (`GET /self`)

Returns the agent's own identity and config — one call to know everything about myself.

```json
{
  "agent_name": "codex",
  "agent_did": "did:key:z6Mkr...",
  "behavior": {
    "model_name": "minimax-2.7-8b-nvfp4",
    "backend_endpoint": "http://workstation-1:8k/v1",
    "backend_provider": "openai_compatible"
  },
  "process_state": "ready",
  "uptime_seconds": 12345,
  "version": "0.x.x"
}
```

### 2. DefraDB Query Tool (`defra_query`)

A tool that lets the agent query DefraDB directly without constructing raw GraphQL.

```json
// Tool call
{
  "tool": "defra_query",
  "args": {
    "collection": "AgentRequest",
    "filter": { "lifecycle_state": { "_eq": "Processing" } },
    "fields": ["request_id", "session_id", "lifecycle_state"],
    "limit": 10
  }
}
```

Benefits:
- Agents can self-inspect their own state
- Can check other agents' states
- Can query schedules, triggers, tool calls
- No more raw GraphQL construction

### 3. Fleet Overview Endpoint (`GET /fleet`)

System-wide visibility across all agents.

```json
{
  "agents": [
    {
      "agent_did": "did:key:z6Mkr...",
      "agent_name": "codex",
      "state": "ready",
      "active_requests": 1,
      "pending_requests": 0,
      "last_seen": "2025-01-15T..."
    }
  ],
  "total_active_requests": 1,
  "total_pending_requests": 0
}
```

[...]

PR #372, opened 2026-06-02
Observability (#338): defra_query tool + /self, /fleet, context-budget surfacing

jackzampolin:
Implements the rescoped #338 (see the rescope comment on the issue): **Tier 1** builds the one feature with a real consumer pull (`defra_query`), and **Tier 2** lands the three cheap, opportunistic surfacings. `mcp_list` stays dropped; `memory` stays parked; spin-outs #369/#370 are unchanged.

## Tier 1 — `defra_query` (the build)
A read-only tool that reads DefraDB collections via a structured `{collection, filter, fields, limit}` contract instead of hand-rolled GraphQL — serving both in-agent self-inspection and external trace consumers (retires Amygdala's hand-rolled GraphQL/escape/polling/terminal-state stack).

- New `crates/defra-agent/src/defra_query/`:
  - **`render.rs`** — JSON filter → DefraDB GraphQL filter syntax. Operators + composition (`_eq`/`_gt`/`_in`/`_and`/`_or`/`_not`), keys validated as identifiers, string values escaped via `graphql::escape_graphql_string` (no hand-rolled escaping, per CLAUDE.md).
  - **`query.rs`** — `DefraQueryParams` contract, `build_query` (validates collection + fields, clamps `limit` to 1..=1000), `CollectionScope` (empty allowlist = all collections). The query core is **decoupled from the rig `Tool` trait** so it can later back an MCP/HTTP management surface.
  - **`mod.rs`** — `DefraQueryTool` (read-only; returns `{collection, count, results}` JSON).
- Gated through `ToolSelection.enable_defra_query` + `defra_query_collections` → `BehaviorToolConfig` → `ToolSurface`, plus `AgentBuilder` methods.

### Decisions (confirmed with the issue owner)
- **ACP**: read-only; trusts defradb.rs's signing-layer enforcement — uses `node.execute()` like every other tool. No per-DID plumbing built.
- **Collections**: **all by default**, optional trim list.
- **Default**: tool is **on for every behavior**.
- **Deferred**: document-schema/CLI/desktop wiring for the gate (`agent.rs` hardcodes on/all-collections). Graceful — absent config = sensible default. Follow-up if operators need to toggle/trim via documents.

## Tier 2 — cheap surfacing (read-only reshapes of data already loaded)
- **`/self`** (alias of `/status`; fields also added to `/status`): joins the agent's own `AgentBehavior` with `InferenceBackend` (provider/endpoint) + `InferenceProfile` (`context_window`) under `behaviors`. Best-effort — never fails `/status`.
- **context budget** on `/status`/`/self`: count + latest `CompactionEntry`. **Caveat:** `CompactionEntry` is keyed by `session_id` (not `agent_did`), so this summary is **node-wide, not agent-scoped** (documented in `self_view.rs`); agent-scoping is a follow-up.
- **`/fleet`**: per-`agent_did` reshape of `AgentRuntime` (`process_state`, `last_seen`=`updated_at`) + active/pending request counts.

## Lean / formal spec
`defra_query` is a read tool and Tier 2 is read-only HTTP surfacing — **no state-machine behavior changes, so no Lean proof change** (per CLAUDE.md's dev-flow gate).

## Tests / verification
- 16 new `defra-agent` tests (filter rendering incl. injection-rejection + escaping, query-builder, scope enforcement, a seeded integration test, gate on/off).
- 6 new CLI unit tests for the Tier 2 join/aggregation logic; `cli_server.rs` extended to hit `/self` and `/fleet` end-to-end (validates the live GraphQL field names).
- `cargo test -p defra-agent` → **1026 passed / 0 failed**.
- `cargo test -p defra-agent-cli --no-fail-fast` → **288 passed / 0 failed**.
- `cargo fmt --all` clean; clippy clean on all changed files.

> Note: `cargo clippy --all-targets` for the whole repo fails on a pre-existing `never_loop` lint in the untouched `defra-agent-lenses` WASM crate (toolchain/clippy-version issue, not this PR). Verified my crates clean in isolation.

> A separate, pre-existing test flake (`transaction conflict. Please retry` from the on-disk RocksDb store) is being tracked in its own issue; it is unrelated to these read-only changes and passes on retry.


jackzampolin:
### Update: access controls for `defra_query` added (resolves the earlier "deferred gate wiring")

Two layers, replacing the original hardcoded on/all:

**1. Built-in secret guard (always on, zero-config).** `defra_query` now rejects any request that *selects or filters on* a sensitive field — seeded with `InferenceBackend.api_key` / `api_key_env_var`. Filter checking is recursive, so `_and`/`_or` nesting can't smuggle a restricted field to probe its value. Credentials can no longer leak through the read surface regardless of the configured scope. This closes a real footgun (a behavior could previously read its own backend's API key).

**2. Operator-configurable scope via the control plane + CLI.** `enable_defra_query` + `defra_query_collections` are now wired end-to-end (ToolSelection schema → `ToolSelectionDocument` parse/round-trip → `agent.rs` mapping → protocol row → CLI desired-state/convert/config_writes/export + `config tools --enable-defra-query` / `--defra-query-collection`). Default is unchanged (on / all collections); operators can now disable it or restrict collections per behavior via manifest / `config apply` / `config tools`. Desktop UI wiring intentionally deferred.

Tests: 4 new guard tests + a node round-trip proving the document fields persist through schema + projections. Read-only → no Lean change.

Verification after this addition: `defra-agent` **1031/0**, `defra-agent-cli` **288/0** (`--no-fail-fast`), fmt + clippy clean on changed files.

iverc (approved):

jackzampolin:
### Update: `defra_query` now consumable via CLI + MCP (retires Amygdala's hand-rolled stack)

`defra_query` was in-agent-only; it now has two external surfaces, both delegating to one shared `run_defra_query` helper (identical filter rendering, secret guard, and scope across all three):

**CLI** — `defra-agent query --collection AgentRequest --field request_id --field status --filter '{"status":{"_eq":"completed"}}'` over GraphQL-over-HTTP. Returns `{collection, count, results}`.

**MCP** — a new MCP *server* (the codebase was previously MCP client-only) mounted at `/mcp` on the running `defra-agent server`, exposing a single `defra_query` tool over streamable-HTTP. `serve --mcp-query-collection <c>` (repeatable) restricts its scope; default all (secret guard always on).

**Both prove the Amygdala-retirement goal via integration tests:** each seeds a linked trace (`AgentRequest` + `AgentResponse` + `AgentMessage` + `AgentToolCall`) and reconstructs it structurally — stitching by `request_id`/`session_id`, with tool-call `args`/`result` as first-class JSON columns (no `AgentMessage.content` reconstruction, no hand-rolled GraphQL client / escaping / polling). Both also confirm the secret guard blocks `api_key`. The MCP test drives a real rmcp client against `/mcp` (list_tools + call_tool).

Final verification: `defra-agent` **1031/0**, `defra-agent-cli` **291/0** (`--no-fail-fast`), fmt + clippy clean on changed files.

jackzampolin:
### Review responses (commit bc37081c)

**P2 — `config tools set` could not disable defra_query / could re-enable it.** Fixed. `--enable-defra-query` is now `Option<bool>`: `--enable-defra-query false` actually disables, and **omitting it writes `None`** so the document's existing value is left unchanged (the upsert omits the field) rather than silently re-enabled by an unrelated update. Added a clap parse test (`false`→`Some(false)`, `true`→`Some(true)`, omitted→`None`).

**Open question — `/mcp` mounted unconditionally / unauthenticated.** Agreed; made it **opt-in**. `runtime_contract_router` takes `Option<CollectionScope>` and only mounts `/mcp` when `serve --enable-mcp` is set (default off; help text flags it as an unauthenticated read surface, same listener exposure as GraphQL). Tests assert `/mcp` is 404 without the flag and serves with it.

**P3 — `/self` context budget looked exact from a bounded scan.** Made explicit. `ContextBudget` now carries `sessions_considered` + `request_scan_limit` and is documented as a recent-activity view (bounded by the recent-request scan), not an all-time total — so the figures aren't read as exhaustive. (Full all-session pagination remains a deliberate non-goal for a summary endpoint; this makes the scope self-describing instead.)

**Open question — no row/agent scoping below `node.execute()`.** Acknowledged and unchanged: consistent with the PR's stated ACP posture (reads run at the node identity; tool-layer controls are the collection scope + the always-on secret-field guard). True per-DID/row read scoping is the larger DID-scoped-reads effort, out of scope here.

Verification after these changes: `defra-agent` **1031/0**, `defra-agent-cli` **292/0**, fmt + clippy clean on changed files. (One pre-existing `cli_codex_shim` flake surfaced mid-run and passed on isolation/retry — tracked in #373, unrelated to these read-only changes.)

PR #375, opened 2026-06-03
Retry retryable GraphQL store conflicts

### LINKED ISSUE #373: Flaky integration tests: 'transaction conflict. Please retry' from on-disk store surfaces raw in CLI write path
## Symptom
Integration tests intermittently fail with a raw datastore conflict, e.g. in `cli_codex_shim`:

```
Error: graphql returned errors: [{"message":"commit error: datastore error: storage error: transaction conflict. Please retry"}]
```

It is **flaky** — the same test passes on a `--no-fail-fast` / retry run (observed: full `defra-agent-cli` suite went 288/0 on re-run). So it's a spurious, retryable conflict, not a real failure.

## Root cause
- `defra-agent server`/`init` always use an **on-disk store with optimistic concurrency** (RocksDb, hardcoded in `persistent_node_builder`, `crates/defra-agent-cli/src/main.rs:482`). There is no backend flag/env today.
- Under racing writes the store returns a **retryable** `transaction conflict. Please retry`.
- The **codex-shim** submission path already treats this as retryable and retries (`crates/defra-agent-cli/src/commands/codex_shim/turn/submission.rs:27` — `graphql_submission_error_is_retryable`), but the **generic CLI GraphQL write path does not**, so the conflict surfaces raw and fails the test (and would fail real operator writes too).

This is exactly the "no tolerated flaky tests" class — the conflict is a known-retryable error that simply isn't retried generally.

## Options
1. **Generic conflict-retry (preferred):** generalize the codex-shim retry to the shared CLI GraphQL mutation helper (`post_graphql` / the write path) so any `transaction conflict` / `database is locked` is retried with backoff. Fixes the class at the source, keeps RocksDb fidelity, and hardens real writes — not just tests.
2. **Test-only in-memory backend:** add a hidden flag/env so the spawned test server uses in-memory storage (`EmbeddedNode::builder()` with no `data_path` → `MemoryStore`, which serializes and never conflicts). Caveat: in-memory loses cross-restart persistence, so the warm-restart tests in `cli_server.rs` would need to stay on RocksDb.

Option 1 is the better fix; option 2 is a faster band-aid for tests only.

## Context
Split out of #338 / PR #372 (observability), where this flake was observed during verification. It is **pre-existing and unrelated** to those read-only changes.

iverc:
## Summary

Closes #373.

Generic GraphQL writes already retried `transaction conflict` / `Please retry` responses on main, but the retry classifier was narrower than the Codex shim path and missed related store-conflict forms like `database is locked` and case variants. This broadens the shared protocol retry classifier so all CLI GraphQL callers that go through `post_graphql` get the same retry behavior.

## Changes

- Add a shared `retryable_graphql_error_text` classifier
- Make retry matching case-insensitive
- Treat `database is locked` as retryable alongside `transaction conflict` and `please retry`
- Extend protocol GraphQL tests to prove both transaction-conflict and database-lock errors retry through `execute_graphql_async`

## Verification

- `cargo test -p defra-agent-protocol graphql`
- `cargo test -p defra-agent-cli --no-run`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/defra-agent-protocol/src/graphql.rs
+++ b/crates/defra-agent-protocol/src/graphql.rs
@@ -779,8 +779,14 @@ fn retryable_graphql_error_message(value: &serde_json::Value) -> Option<String> 
         return None;
     }
     let rendered = serde_json::Value::Array(errors).to_string();
-    (rendered.contains("transaction conflict") || rendered.contains("Please retry"))
-        .then_some(rendered)
+    retryable_graphql_error_text(&rendered).then_some(rendered)
+}
+
+fn retryable_graphql_error_text(message: &str) -> bool {
+    let message = message.to_ascii_lowercase();
+    message.contains("transaction conflict")
+        || message.contains("please retry")
+        || message.contains("database is locked")
 }
 
 fn graphql_transport_error_is_retryable(error: &reqwest::Error) -> bool {
@@ -895,9 +901,19 @@ mod tx_tests {
         last_tx_header: Arc<Mutex<Option<String>>>,
     }
 
-    #[derive(Clone, Default)]
+    #[derive(Clone)]
     struct RetryRecorder {
         attempts: Arc<Mutex<usize>>,
+        first_error_message: Arc<String>,
+    }
+
+    impl RetryRecorder {
+        fn new(first_error_message: impl Into<String>) -> Self {
+            Self {
+                attempts: Arc::new(Mutex::new(0)),
+                first_error_message: Arc::new(first_error_message.into()),
+            }
+        }
     }
 
     async fn capture_handler(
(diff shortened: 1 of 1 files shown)


jackzampolin (approved):
Reviewed — **MERGE**. Sound and well-scoped: bounded retry (max 5 + backoff), only re-issues uncommitted conflict/lock writes (safe). Note this *broadens/hardens the existing classifier* rather than introducing retry where there was none. Non-blocking follow-ups: consolidate with the codex-shim's own `graphql_submission_error_is_retryable` (submission.rs:42) so the two don't drift; the `database is locked` matcher is speculative for the RocksDb-backed #373 flake (which is `transaction conflict`). Tests good.

PR #376, opened 2026-06-03
Add opt-in pruning for live-only config apply docs

### LINKED ISSUE #364: config apply does not prune live_only entities (orphaned tasks/schedules persist)
## Summary

`defra-agent config apply` reconciles **create** and **update** but never deletes **`live_only`** entities (docs present in live config but absent from the desired-state manifest). Orphaned tasks/schedules therefore persist indefinitely and keep firing after a manifest removes them.

## Repro

1. Provision a steward with tasks/schedules A, B, C.
2. Edit the manifest down to a single task/schedule D (removing A, B, C).
3. `config diff --root <manifest> --home <home> --graphql <url> --bind-agent-did home`
   → reports `tasks.create=[D]`, `tasks.live_only=[A,B,C]` (correct).
4. `config apply` with the same args.
   → applies create+update only; `applied.tasks` counts the new ones.
5. Re-run `config diff`.
   → `tasks.live_only=[A,B,C]` **still present**. The old schedules keep running.

Observed live this way during the observability-mcp tool-surface migration: stewards retained `host-health-6h`, `inventory-drift-daily`, `telemetry-freshness-30m` after the manifest collapsed them into a single `host-check`, and those orphaned schedules kept invoking now-removed MCP tools.

## Root cause

`crates/defra-agent-cli/src/desired_state/mod.rs`:
- `is_changed` (~L371) returns true only on `create > 0 || update > 0` — `live_only` is excluded, so apply treats a manifest whose only delta is deletions as "no change."
- `exact_match` (~L366) does count `live_only`, but nothing in the apply path acts on it.

## Ask

Add deletion of `live_only` to apply, gated behind an explicit opt-in so it can't silently nuke live-only state:
- `--prune` (or `--reconcile-delete`) flag that deletes `live_only` docs per collection, OR
- make `live_only` count toward `is_changed` and delete under `--prune`.

Until then the operator workaround is manual GraphQL `delete_<Collection>(docIDs: [...])`.

## Workaround (for reference)

```graphql
mutation { delete_Schedule(docIDs: ["bae-..."]) { schedule_id } }
mutation { delete_Task(docIDs: ["bae-..."]) { task_id } }
```
(The `filter: {field: {_in: [...]}}` form 400s; `docID`/`docIDs` works.)

iverc:
## Summary

Adds an explicit `--prune` option to `defra-agent config apply` so desired-state apply can delete live-only config documents that are absent from the manifest.

Normal `config apply` remains non-destructive: it still creates and updates desired docs, but leaves `live_only` drift in place. With `--prune`, apply now deletes live-only non-principal desired-state docs in dependency-safe order, so stale schedules/triggers are removed before their tasks.

## Changes

- Add `defra-agent config apply --prune`
- Delete live-only desired-state docs inside the same apply transaction
- Resolve live-only unique IDs to DefraDB `_docID`s before deletion
- Prune in dependency-safe order:
  - event triggers
  - schedules
  - tasks
  - behaviors
  - tool selections/services/profiles/backends
- Preserve non-prune behavior as non-destructive
- Add `pruned` counts to the apply JSON report
- Keep `provision` non-pruning by default

## Tests

- `cargo test -p defra-agent-cli config_apply_prunes_live_only_tasks_and_schedules_when_requested_locally`
- `cargo test -p defra-agent-cli --test cli_config_apply_local`
- `cargo test -p defra-agent-cli --test cli_help`
- `cargo test -p defra-agent-cli --test cli_config_apply_order`
- `git diff --check`

Closes #364


jackzampolin (changes requested):
Engineering on this is clean — opt-in, default-off, field-ownership correct (only orphaned `live_only` docs, never runtime-owned state), and well-tested. Holding it on a **process gate, not the code**.

This introduces **delete semantics** at the apply boundary, and per CLAUDE.md ("anything that changes what transitions are legal … starts in Lean") that needs to begin in the formal model. Right now:

- `ApplyReconcile` explicitly scopes `live_only` docs as **reporting-only, not deleted** — `proofs/Proofs/ApplyReconcile/Diff.lean:37` and `Manifest.lean:45` say so directly.
- The apply conformance fence (`config_import.rs:885`, `generated_apply_reconcile_cases_fence_production_apply_write_boundary`) only fences `create`/`update`, so this new prune/delete path runs **entirely outside the proven write boundary**.

This is exactly what #57 reserved: add `ApplyStep.delete` + prove **T-Delete-safety** (delete permitted only when no live doc references the target), extend the fence to cover prune, *then* land the Rust prune. So let's gate this PR on the #57 Lean work.

(Real-world risk is soft — pruning a Schedule/EventTrigger leaves dangling `AgentRequest.caused_by_trigger_id` *strings*, not FK relations, so no data corruption — which is why this is process discipline rather than a correctness bug. The code here looks right; it just needs the model in front of it.) Once #57's delete model + fence land, this should sail through.

jackzampolin:
Heads-up on overlap, and a proposal — no action needed from you yet.

Now that **#57** (the `ApplyReconcile` delete model + `t_delete_safety` + the apply conformance fence) has merged, the prune work this PR was changes-requested on is unblocked. I went ahead and wired `config apply --prune` through that proven path in **#387**.

The two PRs implement the same feature but differently:
- This PR (#376) prunes via a separate post-apply `prune_live_only_desired_state` that deletes live-only docs directly — **outside** the proven model/fence.
- #387 routes deletes through the proven `apply_model::diff_prune` / `delete_safe` (a manifest→model reference bridge), so every delete is inside the #57 boundary the review asked for.

Since they're mutually exclusive (two prune paths, two report shapes), I think **#387 should supersede this PR's execution approach** — it's the proof-backed version of the same goal.

**That said, this PR has two things #387 doesn't, and they're better — I'd like to bring them over with credit to you:**
1. Your **e2e integration test** (`config_apply_prunes_live_only_tasks_and_schedules_when_requested_locally`) — #387 only has unit tests on the bridge/diff; your real `config apply --prune` against a node is exactly the coverage it's missing.
2. Your separate **`pruned` report field** — nicer operator UX than #387 folding deletes into `applied`.

Proposed plan: port (1) and (2) onto #387, then close this PR as superseded by #387 (`Closes #364`). Totally open to alternatives — e.g. if you'd rather rebase this PR onto the #57 model and own it, happy to hand off #387's bridge. Let's sync on which path you prefer before anything gets closed.

Context: #387 · #57 · #364

iverc:
Updated this PR to take the second path you suggested: it now routes pruning through the #57/#387 proof-backed `apply_model::diff_prune` / `delete_safe` path, while keeping the e2e integration test and separate `pruned` report field from this PR.

Validation run:
- `cargo test -p defra-agent-cli config_apply_prunes_live_only_tasks_and_schedules_when_requested_locally`
- `cargo test -p defra-agent-cli generated_apply_reconcile_cases_fence_production_apply_write_boundary`
- `cargo test -p defra-agent-cli diff_manifests`
- `cargo test -p defra-agent-cli delete_mutation_field_escapes_unique_value`
- `cargo test -p defra-agent-cli prune`
- `git diff --check`

Ready for re-review @jackzampolin

jackzampolin:
Integrated into main via #411; closing this standalone PR to keep the queue current.

PR #382, opened 2026-06-03
True subagent enablement + clean agent tool surface (config-apply enablement, named targets, spawn convergence, verb×noun tool families) (#377)

### LINKED ISSUE #377: Truly enable subagent execution as a first-class capability
## Summary
The subagent execution path is fully built and conformance-tested but **disabled by default with no first-class enablement surface**. Make it a real, operator-enable-able, end-to-end-exercised capability so higher-level orchestration can build on it.

## Current state (precise)
- **Off by default** — tools are offered only when `subagent_spawn_enabled: true` **and** `subagent_targets` is non-empty (`tool_surface/selection.rs:21-23`); `subagent_spawn_enabled` defaults to `false` (`agent.rs`).
- **Direct execution stubbed by design** — `SpawnSubagentTool::call()` returns `not_yet_executable_error` (`toolset/subagent.rs:193-196`); the persistence hook intercepts every spawn and handles it (`hook/persistence/message_spawn.rs:217-543`, returns `Skip`).
- **Plumbing is live** — `SubagentSource` is registered as a `TriggerSource` and the background-completion observer is spawned (`agent/runtime/startup.rs:204-263`).
- **Tested** — 17 passing conformance tests (`tests/subagent_source_conformance.rs`) plus `tests/r4_subagent_tools/` (foreground/background spawn, wait, cancel).

## What "truly enabled" should mean (acceptance)
- A documented, ergonomic way to enable subagents on a behavior via the apply path (ToolSelection config), with validation of `subagent_targets` against known behaviors.
- An end-to-end smoke path (CLI or example manifest) demonstrating parent-spawns-child-and-receives-result against a live node, foreground and background.
- Decide the fate of the stubbed `Tool::call()` (keep as hook-only invariant, documented) — no `not_yet_executable_error` surfacing to operators in normal use.
- Clear operator docs on depth cap, cascade-cancel, and cross-deployment (trusted-fleet) behavior.
- `cargo test`, clippy, fmt clean.

## Related
#177/#213/#216/#200 (substrate) · #9 (identity). Blocks the workflow orchestration design issue.

jackzampolin:
Closes #377.

Subagent enablement + a clean, semantically-coherent agent tool surface. Ships **local-only** delegation (cross-deployment deferred behind a default-off flag pending ACP), validated with live inference.

## Subagent enablement
- `config apply` persists + validates all subagent fields idempotently; `AgentBehavior` gains `description`/`summary` (+ a migration patch run before behavior reads on **every** startup/embedder/config path).
- **Named delegation targets:** a target is a named `(agent_did, behavior_id)` pair with a description (JSON in the existing field — no new collection, no Lean change). The model only uses a friendly `name`; the runtime hides did/behavior.
- **Spawn convergence:** local/remote unified into one "write the bridge → `SubagentSource` creates the child" path (sole creator, DID-anchored).
- **Cross-deployment DEFERRED (default-off, both sides):** `subagent_allow_cross_deployment` (default `false`) gates the orchestrator **and** the receiver/recovery — flag-off = no cross-deployment child on spawn, receive, or recovery, even between paired peers.
- Live e2e (ignore-gated): local delegation ✅; cross-node delegation over real P2P ✅ behind the flag.

## Clean tool surface (verb × noun)
The agent-delegation / async area had sprawled into three overlapping families. Consolidated to **one verb vocabulary, two non-overlapping nouns**:
```
spawn_subagent · wait_subagent · cancel_subagent · list_subagents · read_subagent · steer_subagent
spawn_process  · wait_process  · cancel_process  · list_processes · read_process
```
- `delegate_to_agent` **removed** (folded into `spawn_subagent`).
- background-tool family renamed to **process** (`spawn_process(tool_name,args)` — bash today; MCP out of scope).
- **`read_` is context- + content-honest:** token budget + resume cursor + an honest `next_offset`/`next_sequence` · `has_more` · `total_bytes`/terminal envelope — pages through *all* output gap-free, never silently drops (incl. an oversized single message). UTF-8-safe byte paging.
- Lean spec kept in sync (the `action` tool-name label updated; stable contract/error/vocab IDs retained).

## Security posture (audited 3×, accepted, documented)
The cross-deployment boundary is a **replication-trust model, not authorization** (a node runs any replicated `AgentRequest` matching its own DID; ACP not wired) — accepted for the **trusted-fleet** model, and the reason cross-deployment ships default-off. Three adversarial audit rounds drove fixes: upgrade-crash migrations, two convergence races (orphan-cancel only cascades on `cancel_policy=Cascade`; DID-anchored claim gate), one-sided→two-sided deferral, desktop fixture, the read content-honesty bug.

## Notes for reviewers
- No new collection / no Lean **module** change (`proofs/` only had a tool-name label synced). Named targets + flags are scalar/JSON on existing collections.
- A separate top-level `delegate_to_agent`-style remote primitive does **not** exist anymore; if "no cross-deployment" should extend to other future remote primitives, that's a follow-up.
- Pre-existing main breakage fixed (commit `ed3374eb`): #338 left desktop `ToolSelectionRow` literals incomplete (`cargo build --workspace` was already red).
- CI heads-up: the `defra-agent-cli` Lean-conformance tests can transiently fail under heavy parallel compilation (Lean `lake build` contention) — they pass on a non-contended rerun.

## Tests (all green locally)
~250 tests across the subagent / conformance / migration / read-paging suites incl. `state_machine_conformance` (89, Lean re-extracted), apply idempotency, upgrade-path migration, flag enforcement, gap-free incremental reads (UTF-8 + stderr boundary). `cargo build --workspace` + `cargo fmt --all --check` clean; `lake build` clean (0 sorries).


jackzampolin:
## Second deep adversarial audit — round complete

A second 5-auditor adversarial pass found (and we fixed) issues the first round's fixes had left incomplete:

- **Cross-deployment deferral was one-sided** — the flag gated the orchestrator (spawn) but not the receiver. A flag-off node would still run a paired peer's cross-deployment child via `SubagentSource`'s trusted-peer path, and recovery ignored the flag. **Fixed:** the trusted-peer claim path, the shared `subagent_spawn_denial` (covers recovery + non-trusted fallback) now honor `subagent_allow_cross_deployment`. Flag-off (default) now means **no cross-deployment child on spawn, receive, or recovery** — even between paired peers.
- **H1:** the desktop live-agent fixture used a bare-string subagent target (silently dropped post-redesign → broken demo). **Fixed:** JSON-encoded via `subagent_target_entry`.
- **H2:** the migration fix only covered the CLI `serve` path. **Fixed structurally:** `ensure_agent_behavior_migrations` now runs inside `from_default_behavior_documents` (covers desktop + all embedders) and `resolve_config_access` (covers offline `config diff/apply/export`); upgraded-DB tests added.
- **Orphan-cancel** extended to the parent-terminal-not-interrupted case + a test that drives the real cancel-before-materialize window.
- Lows: idempotency now asserts the flag round-trips; `p2p pair` help notes the deferral; stale "remote targets always survive" comments corrected.

**Verified clean by the audit:** hygiene fixes, flag wiring (no drift, defaults false everywhere), GraphQL escaping, no weakened tests, legacy-target failure mode (no crash).

State: 216 tests green across 11 suites, `cargo build --workspace` + `fmt --check` clean.

jackzampolin:
## Third focused audit — on the round-2 fixes themselves

A focused adversarial pass over the new gating/migration code (the round-2 fix commits) found two HIGH issues + one MEDIUM, now fixed:

- **HIGH (regression from the round-2 orphan-cancel fix):** the parent-terminal re-check interrupted the child on *any* terminal parent — including normal completion — with no `cancel_policy` check, **wrongly killing legitimate detached/background subagents** meant to outlive a completed parent. Fixed to mirror the codebase cascade invariant: interrupt only when `cancel_policy == Cascade` **and** a genuine cancel signal (bridge cancelled / parent interrupted / parent cancel-worthy-terminal), **excluding clean completion**. Added the missing negative tests (detached child survives; cascade child survives a normally-completed parent) and reframed the test that had enshrined the bug.
- **HIGH (third migration gap):** `defra-agent init` re-run on a pre-#377 DB crashed (read `description`/`summary` before the migration). Fixed — migration now runs in `init` too.
- **MEDIUM:** `ToolSelectionDocument::validate()` accepted bare-string targets (and a test blessed them). Now enforces `SubagentTarget` JSON; test fixed; stale bare-string fixtures converted.
- **LOW:** `paired_peer_dids` now excludes the node's own DID (prevents a self-pairing row mis-routing a local spawn).

**Verified good this round:** cross-deployment gating is **complete** — every child-create path is gated when the flag is off (PASS, confirmed by a mutation test; non-vacuous receiver tests). Migration `serve.rs` removal safe; escaping clean; no weakened assertions.

**Noted for separate decision (out of #377 scope):** there's a distinct top-level `delegate_to_agent` tool that creates remote-DID requests gated by its own `allowed_target_dids` allowlist — not a subagent child, so not governed by `subagent_allow_cross_deployment`.

State: 235 tests green across the subagent/conformance/migration suites; `cargo build --workspace` + `fmt --check` clean.

PR #398, opened 2026-06-05
Add sessions tool for agent session history

iverc:
## Summary

Adds a first-party `sessions` runtime tool as part of #338, giving agents a structured way to inspect their own recent persisted session history.

The tool supports:

- `action: "list"` with an optional `limit`
- recent session ordering from agent-scoped `AgentRequest` rows
- session metadata from `AgentSession`
- latest request status/lifecycle fields
- request, message, and compaction counts
- latest message and compaction timestamps

This keeps session-history visibility inside the normal tool surface without requiring agents to hand-write raw GraphQL.

## Notes

This is a focused runtime-tool slice for #338. It does not add a new HTTP endpoint or desktop UI surface.

## Tests

- `cargo test -p defra-agent session_history`
- `cargo test -p defra-agent tool_surface`
- `git diff --check`


jackzampolin:
Integrated into main via #409; closing this standalone PR to keep the queue current.

PR #406, opened 2026-06-05
Fix ToolSelection apply with explicit empty list fields

### LINKED ISSUE #404: Apply path: graphql_input_literal still renders empty arrays as [] (latent JsonArray corruption)
## Summary
The config **apply** path renders document field values via the generic `graphql_input_literal` (`crates/defra-agent-protocol/src/graphql.rs`), whose `Value::Array` arm emits a bare `[]` for an empty array. That is the same DefraDB quirk fixed for the CLI/runtime/desktop **mutation-field** helpers under #382: a bare `[]` is typed as `JsonArray([])`, incompatible with a `NillableStringArray` (`[String]`) column — the create "succeeds" storing a JsonArray, and every later update of that document fails re-validation with `expected ScalarArray(NillableStringArray), got JsonArray([])`.

## Why it's latent (not currently failing)
The apply suite is green because the apply/desired-state conversion omits empty/None list fields upstream, so `graphql_input_literal` is not currently handed an explicit empty array for a `[String]` column. But a manifest that **explicitly** sets a list field to `[]` (e.g. `cli_tool_names: []`, `subagent_targets: []`) would flow an empty `Value::Array` into `graphql_input_literal` and hit the corruption — create stores JsonArray, subsequent applies/updates of that doc then fail.

## What was already fixed (for contrast)
#382 fixed the three duplicated **mutation-field** helpers to emit `field: null` for empty lists (protocol `string_list_field`, runtime `document_config::graphql_fields::graphql_string_list_field`, desktop `mutations/graphql::graphql_string_list_field`). The apply path uses a *different* renderer and was deliberately left untouched there because `graphql_input_literal` is a generic JSON→GraphQL renderer (nested objects/arrays, not just top-level `[String]` fields), so a blanket "empty array → null" could be wrong in non-list contexts.

## Proposed fix (needs care)
Decide where empty-array→null is safe for the apply path. Options:
1. Make the apply/desired-state conversion **omit** explicitly-empty list fields the same way it already omits None (so they never reach `graphql_input_literal`). Lowest blast radius; matches current de-facto behavior.
2. Type-aware rendering: only coerce empty arrays to `null` when the target column is a `NillableStringArray` (requires schema-type awareness at render time).
3. Avoid touching the generic renderer; instead special-case list fields in the apply writer before serialization.

Option 1 is likely the cleanest and most consistent with the existing "omit empties" upstream behavior.

## Coverage to add
- An apply integration test with a manifest that sets a `[String]` field explicitly to `[]`, then applies a **second** time (the update re-validation is where corruption bites), asserting both applies succeed and the field round-trips as empty.

## Refs
- Sibling fix + the quirk write-up: #382 (PR), the `string_list_field` doc comment.
- Tracking note in the R4c/subagent work.

## Priority
Low — latent; no current red test. Becomes real the moment a manifest declares an explicit empty list for a string-array column.

iverc:
## Summary

- include the missing ToolSelection subagent controls in desired-state export/apply paths
- persist subagent ToolSelection fields from the dedicated config writer
- normalize and validate `subagent_targets` alongside other ToolSelection list fields
- add regression coverage for applying explicit empty ToolSelection arrays twice

## Why

Fixes #404.

DefraDB cannot infer typed empty array literals from bare `[]`, so config apply must keep empty list fields away from the generic GraphQL literal renderer. The existing import sanitizer already handles that safely; this PR closes the ToolSelection coverage gap so fields like `subagent_targets: []` flow through the desired-state/apply path and are sanitized before rendering.

This intentionally does not change `graphql_input_literal`.

## Tests

- `cargo fmt --check`
- `cargo test -p defra-agent-cli desired_state::tests::tool_selection_round_trip_preserves_subagent_controls`
- `cargo test -p defra-agent-cli config_bundle::tests::empty_list_is_cleared_on_update_but_omitted_on_create`
- `cargo test -p defra-agent-cli --test cli_config_apply_e2e config_apply_accepts_explicit_empty_tool_selection_lists_twice`
- `cargo test -p defra-agent-cli --test cli_config_tools`


jackzampolin:
Integrated into main via #409; closing this standalone PR to keep the queue current.

PR #427, opened 2026-06-08
Add rig-core owned loop audit and design

### LINKED ISSUE #400: Investigate replacing rig-core with an owned completion+tool loop
## Summary
Investigate replacing the `rig-core` dependency with an owned completion + tool loop.

## Context
defra-agent integrates `rig-core` for: the completion-model trait, the tool trait, the multi-turn completion→tool loop (`agent.stream_prompt`), the `PromptHook` system, and streaming. Over the subagent + tool-surface work (#377 / PR #382) we've built the hard parts ourselves — hook-based persistence (`DefraSessionHook`), the tool surface + tool lifecycle, truncation, subagents/processes, the stream processor, and the daemon. rig's remaining role is narrow: the multi-turn loop, the provider/streaming client, and the in-memory message-vec threading.

## Motivation
rig's hook model is starting to constrain us. Concrete example from the output-bounding audit: `PromptHook::on_tool_result` can truncate the **persisted** tool result but **cannot rewrite rig's in-loop working message**, so an oversized tool result is fed raw to the model within a single request (only bounded cross-request when history reloads from truncated docs). Owning the loop would give native control of in-loop context/truncation and message management, and remove impedance-matching to rig's traits.

## To investigate (design-doc first)
- The exact rig API surface we depend on (`CompletionModel`, `Agent`/`AgentBuilder`, `stream_prompt`/multi-turn, `PromptHook`, streaming types, provider clients) and every call site.
- Cost of owning it: reimplement the completion→tool loop, the streaming response parse, and the provider abstraction — including the cross-provider tool-call format quirks rig currently absorbs (OpenAI-compatible, Anthropic, vLLM, etc.).
- Benefits: full in-loop control (native truncation/context management), fewer deps, a simpler single model (our hook/lifecycle is already ours).
- Decision: own-the-loop vs keep-rig.

## Sequencing
**After PR #382 merges.** This is a foundational change; do not entangle it with the subagent enablement work.

## Acceptance
A rig-dependency-surface audit (sizing) + a design doc, then a go/no-go decision.

iverc:
## Summary

Adds the issue #400 design/audit doc for replacing `rig-core` with a Defra-owned completion + tool loop.

The doc covers:

- the current `rig-core` dependency surface across runtime construction, daemon inference, streaming, hooks, tools, messages, admission, errors, providers, and tests
- sizing for each replacement area
- the concrete motivation around bounded in-loop tool observations
- an adapter-first migration plan
- a go/no-go decision

## Decision

Go for an adapter-first owned-loop spike.

No-go for an immediate full `rig-core` removal, because provider streaming, message persistence, tools, hooks, compaction, admission, and tests are too intertwined for a safe one-step replacement.

## Verification

- `git diff --cached --check`

Closes #400

### DIFF
--- a/docs/superpowers/specs/2026-06-08-issue-400-rig-core-owned-loop-design.md
+++ b/docs/superpowers/specs/2026-06-08-issue-400-rig-core-owned-loop-design.md
@@ -1,0 +1,388 @@
+# rig-core Owned Loop Audit and Design
+
+**Issue:** #400
+
+**Date:** 2026-06-08
+
+**Scope:** dependency-surface audit, sizing, and go/no-go for replacing
+`rig-core` with a Defra-owned completion and tool loop.
+
+## Summary
+
+The current `rig-core` dependency is not just a thin provider client. It is a
+shared shape for the daemon's model type, agent builder, multi-turn loop,
+streaming events, prompt hook callbacks, tool trait, tool definitions, message
+model, provider clients, completion errors, token usage, compaction calls, title
+generation, one-shot execution, transcript presentation, and tests.
+
+The strongest reason to own the loop is real: Defra needs native control over
+in-loop context management. The current hook can truncate the persisted tool
+result and can persist a shorter model observation, but it cannot guarantee that
+rig's in-memory working message for the same turn is rewritten before the next
+completion request. That means an oversized tool result can still affect the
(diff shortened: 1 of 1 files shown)


jackzampolin:
thank you for pushing this, closing in favor of #426

Issue #340, opened 2026-05-28
Investigate Defra Agent skills integration using Codex as guide


jackzampolin:
## Context
After #333, Codex can act as a frontend for Defra Agent. Codex also has a skills system that is a useful guide for how Defra Agent could expose reusable instructions, workflows, and tool integrations.

## Investigation
- Study the Codex skills model and identify which parts map directly to Defra Agent concepts.
- Propose a Defra-native skill representation, storage model, and activation rules.
- Decide how skills should interact with behavior configuration, prompts, tool policy, and remote/fleet operation.
- Identify migration/import paths for users who already have Codex skills.
- Define the test/conformance surface needed before implementation.

## Deliverable
A short design plan with recommended implementation slices and open questions, followed by implementation issues if the plan is accepted.

Follow-up from #333.

jackzampolin:
Closing as completed — the investigation deliverable shipped and the implementation followed in PR #371 (merged 2026-06-03): *Skills integration: progressive-disclosure skills with live reconcile (#340)*.

Delivered:
- Design spec: `docs/superpowers/specs/2026-06-02-skills-integration-design.md`
- `Skill` collection + Lean privilege algebra
- `config skill` CLI (`add`, `list`, `import`, `export`)
- Runtime: skill catalog in preamble + `load_skill` progressive disclosure
- Live reconcile on skill changes (no restart)
- Codex shim `skills/list` + enable/disable
- `config skill import` from Codex `SKILL.md` trees

Validated downstream in the defending-code-reference-harness work: reference `.claude/skills/` import, `load_skill` smoke (`vuln-scan`), and subagent delegation via named `SubagentTarget` (PR #382 / #377).

No further tracking needed on this investigation issue.

PR #468, opened 2026-06-10
Skills: slash-command selection recorded as request metadata (#340 slice)

jackzampolin:
## Summary

This PR is the slash-command request-recording slice for #340:

- Adds `selected_skill_ids_from_prompt_slash_commands()` in `skills.rs` to extract leading `/skill-id` commands as references only. The prompt text is never edited, and path-shaped tokens like `/work/file.rs` are rejected.
- Records non-empty selections as `{"selected_skill_ids":[...]}` request metadata on slash-prompt request creation paths: config task run, lifecycle/manual via materialization, CLI request helpers, trigger materialization, and subagent request materialization. Empty selections omit the metadata field.
- Keeps all GraphQL string interpolation escaped and does not emit empty list literals in mutations.

## Design / Lean Review

Reviewed against `origin/design/issue-340-skills-investigation` D1-D6 and `Proofs/Skills.lean`:

- The runtime consumer is already on `main`: `agent/daemon/request.rs` parses `AgentRequest.metadata.selected_skill_ids`, and `PromptBuilder::selected_skill_reminders` resolves each id against the behavior's effective skill set (D5) before injecting bodies with the D3 ceiling/degrade behavior.
- This PR does not need a Lean extension. It records reference metadata only; it does not add request lifecycle states, widen legal transitions, or introduce a new privilege invariant. The fenced safety property remains the zero-sorry Skills privilege algebra already in `Proofs/Skills.lean`.
- Existing Codex shim coverage on `main` exercises deterministic explicit selection and effective-set non-escalation; this slice adds extraction/metadata coverage for slash-prompt request paths.

## Verification

- `cargo test -p defra-agent`

Proofs were not touched, so `lake build` was not required.


jackzampolin:
Follow-up review fixes after this PR was merged are in #483.

Issue #445, opened 2026-06-09
Reload sends dangling assistant tool-calls to the provider (compaction reducer)


jackzampolin:
## Severity
Correctness (P2). Confirmed by adversarial verification against the code. Distinct from #442: #442 is the **persist** seam (a stalled turn writes a completed tool call with no result message); this is the **reload** seam (a previously-persisted assistant tool-call with no matching result is sent to the provider).

## Mechanism
Every request build runs `compaction::strip_tool_results(full_history)` (`crates/defra-agent/src/agent/daemon/request.rs:62`) → `prompt_builder.build(...)` → provider. `strip_tool_results` (`crates/defra-agent/src/compaction/history.rs:11-57`) keeps **all** assistant messages — including assistant `ToolCall`s that have no corresponding tool-result message — and only transforms tool-result *content*. The model's `Transcript.PairClosed` permits a persisted assistant message whose tool call is `failed`/`running`/unpaired (only *completed* calls are required to have results), but providers (OpenAI/Anthropic) require every assistant `tool_call` to be followed by its tool result. So after an interrupted/failed/timed-out tool round leaves an unpaired call in durable history, the next request sends an assistant function-call with no result → provider rejects.

## Fix direction (Lean-first)
Filter unpaired tool calls at the reload boundary. `strip_tool_results` is a **Lean-conformance-fenced reducer** (mapped in `tests/state_machine_conformance/streaming_compaction.rs`; `Proofs/Compaction` has `no_orphaned_tool_results_after_reduction`, the inverse pairing property). Per CLAUDE.md the reducer change starts in the Lean Compaction model:
1. Extend the Compaction reducer model so reduction drops assistant tool-calls that have no result message (the reload-safety dual of `no_orphaned_tool_results_after_reduction`); prove it preserves coherence.
2. Update the conformance cases + `strip_tool_results` to match.
3. Two-pass filter (verified spec): pass 1 collects call-ids that have a result message; pass 2 drops unpaired `ToolCall` items from assistant messages (dropping the assistant message entirely if only the unpaired call remained). Log dropped calls.

## Verified-spec notes (subagent, validated)
- Filter point: `strip_tool_results` (on the normal build path) — confirmed at request.rs:62, not compaction-only.
- Breakage risk: sessions with interrupted/failed tool calls — the reloaded provider history will no longer contain the dangling call; assistant turn-count/telemetry may shift. Compaction tests that construct unpaired calls will need updates.

[...]

Issue #460, opened 2026-06-10
Skills integration: implementation slices (from the #340 design)


jackzampolin:
The #340 investigation produced the design + Lean privilege algebra (Skills.lean, zero sorrys, landed). The implementation slices remain unshipped. Design record: docs/superpowers/specs/2026-06-02-skills-integration-design.md at git history (removed from tree in the docs curation; the past lives in history).

Slices per the design's D1–D6: skill document collection + apply path (must extend Lean ApplyReconcile first), runtime resolution into the tool surface, progressive disclosure via load_skill (ceiling already enforced), CLI affordances, desktop surfaces.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jackzampolin:
Closing — the implementation slices from the #340 design have all landed.

**Landed (slices 1–5 + proof):**
- **Skill collection + apply path** — `Collection::Skill` wired into `ApplyReconcile` (`crates/defra-agent/src/collection.rs`), with skill_refs/skill_excludes validation in the desired-state path.
- **Runtime resolution into the tool surface** — `crates/defra-agent/src/skills.rs` (`effective_skills` / `skill_tool_ceiling` / `skill_tools`), injected per-behavior in the runtime context.
- **Progressive disclosure** — `load_skill` tool with the privilege ceiling enforced (catalog renders descriptions only; bodies load on demand; never widens the surface).
- **CLI** — `crates/defra-agent-cli/src/commands/config/skill.rs` (add/list/show/rm/import/export/set-enabled) with live reconcile.
- **Desktop** — PR #501 shipped the skills surface (view/manage/bind), `SkillConfigPanel.tsx` + behavior binding.
- **Proof** — `proofs/Proofs/Skills.lean`, zero `sorry` (activation ⊆ ceiling, composition closure, principal-respecting candidates).

**On `Task.skill_refs`:** activation was deliberately redirected at design time from a persistent Task field to **request-scope activation** (request metadata `selected_skill_ids` + slash-command selection, recorded as request metadata — PRs #468/#483). That path is built and live, so there is no remaining Task-field gap. If we later want Task-document-driven skill firing as a distinct affordance, that's a fresh issue, not leftover scope here.

Merged PRs: #371 (progressive-disclosure core), #468 (slash-selection as metadata), #483 (native slash UX), #501 (desktop surface).

Issue #446, opened 2026-06-09
Conformance hardening: close coverage gaps for proven-but-unwitnessed invariants


jackzampolin:
Follow-up to the #426 conformance audit. These are proven Lean contracts that lack a Rust conformance witness (or accepted-boundary registration). Each is a **paired Rust + Lean-ledger change** (conformance_consumers.rs ↔ Proofs/Conformance/CoverageLedger.lean) that must pass `lake build` — not a pure Rust test — so they belong in a focused Lean-fenced PR, not bolted onto the rig PR.

Verified specs exist for each (from the audit's design pass). Priority order:

### 1. MCP health K=1 collapse (smallest)
H7 (`Proofs/MCPHealth/Properties.lean`): at `K=1`, first `probeFail` from Healthy → Evicted directly. Production defaults `K=3`, so K=1 is unwitnessed. Add a `health_checker.rs` unit test driving `step_service` with `failure_threshold_k=1` + a ledger entry. The Lean `k1ProjectionCases` already exist.

### 2. liveness-timeout deadline precondition (boundary doc, no behavior change)
Lean `StreamingResponse.Transition` (Transition.lean:64) requires `now > streamIdleDeadline` when firing `streamIdleTimeout`; the runtime uses an independent `tokio::time::timeout(liveness_timeout, ...)`. The design doc marks this an intentional runtime-obligation marker, but it lacks a `Boundaries.lean` registration. Add the boundary witness + ledger entry. Relevant to #437 (configurable timeout) — without registration, making the timeout configurable could silently diverge from the deadline assumption.

### 3. Recovery sweep convergence measure
`Proofs/Recovery/Contract.lean` (`finite_stale_rows_converge`, `aggregateMeasure_recovered_zero`): a sweep makes all stale (queued/running) rows terminal in one pass (measure→0, idempotent). Add a Rust test (multiple stale `InferenceCall` rows → `recover_all` → assert terminal + second pass recovers 0) + ledger entry.

### 4. Request-level S1/S3/S6/L1/L3 post-transition assertions
`Proofs/Properties/Safety.lean` + `Liveness.lean`. Existing `request_lifecycle.rs` tests verify transition legality/actions but not the post-transition *properties* (terminal irreversibility, monotonic progress, persistence-before-completion via materialization, bounded termination on the failure path, recovery convergence). Add assertions + ledger entry.

### 5. S7 capacity invariant (largest)
[...]

PR #473, opened 2026-06-10
Conformance liveness: multi-tool-result turn persistence + periodic subagent-liveness sweeps (#465)

### LINKED ISSUE #465: Conformance gap: expired background subagent children can remain active and strand parent response wait
## Summary

The Defra-native harness exposed a recovery/conformance gap in background subagent liveness: when a high-fanout parent request spawns background subagent verifier requests, child streams can go quiet/close and pass their verifier deadline, but the runtime can continue reporting those child requests/tool calls as active. The parent `response wait` remains stuck even though there are no established inference endpoint connections.

This is distinct from the earlier datastore transaction-conflict symptom. I tested with a local dirty v0.3.0 worktree that extends transient GraphQL mutation retry for `transaction conflict` / `Please retry` errors, and this run did **not** show transaction-conflict results in the parent trace before wedging.

Related/nearby: #432 and #437.

## Repro Context

Harness repo:
`/Users/johnzampolin/go/src/github.com/sourcenetwork/defending-code-reference-harness`

Defra Agent repo:
`/Users/johnzampolin/go/src/github.com/sourcenetwork/defra-agent`

Defra Agent base:
`v0.3.0`, commit `04fba0533abff3059626d6693a04527a259e05cb`, dirty with harness integration work plus transient mutation retry experiment.

Linux binary used by harness:
`/tmp/defra-agent-linux-target/debug/defra-agent`

Binary SHA:
`5bb8b01dccc1c12dd85755eb7a0668a22330dee3f8e1f5b76c3e995f71cfbf7d`

Endpoint:
`http://100.73.235.38:8000/v1`, model `d4f`

Command:

```bash
env -u DEFRA_AGENT_API_KEY -u DEFRA_AGENT_API_KEY_ENV_VAR \
  DEFRA_AGENT_BINARY=/tmp/defra-agent-linux-target/debug/defra-agent \
  DEFRA_AGENT_INFERENCE_URL=http://100.73.235.38:8000/v1 \
  VULN_PIPELINE_MODEL=d4f \
  .venv/bin/vuln-pipeline triage results/defradb-rs/20260610T141537Z-defra-skill-smoke/VULN-FINDINGS.json \
  --target defradb-rs \
  --votes 1 --allow-plain-docker \
  --out-dir results/defradb-rs/20260610T161059Z-defra-triage-retry-transient-retry \
  --max-turns 250 --defra-deadline-secs 3600 --verifier-deadline-secs 1200 \
  --defra-max-concurrent 16 --defra-max-output-tokens 16384 --defra-trace-limit 2000
```

Parent request observed during the run:
`a493e0f8-5e9f-4bf7-a38c-0d3dc5e55327`

## Observed Behavior

The model began triage and high fanout:

- `loaded_skills` path was working (`triage` loaded).
- Parent trace showed file/checkpoint setup and `spawn_subagent` activity.
- Runtime status reached high fanout:
  - `active_requests: 17`
  - `active_tools: 23-24`
  - `general_queue: 32`
  - `expired_processing_count: 0`
  - about 16 established connections to the inference endpoint.

After child streams went quiet and the verifier deadline crossed:

```json
{
  "general_queue": 32,
  "active_requests": 17,
  "active_tools": 23,
  "expired_processing_count": 16,
  "request_progress_ages_ms": {
    "min": 1089184,
    "max": 1089726
  },
  "deadlines": {
    "min": -2334355,
    "max": 79381
  }
}
```

At that point:

- `lsof -nP -iTCP@100.73.235.38:8000 -sTCP:ESTABLISHED` showed zero established endpoint connections.
- The parent `response wait` was still blocked.
- The active request/tool counts did not drain after another recovery poll.
- The queued children remained queued (`general_queue: 32`) even though the first wave had expired.
- Parent trace did not show transaction-conflict failures:

```json
{
  "count": 14,
  "tools": [
    { "name": "bash_unrestricted", "count": 6 },
    { "name": "list_files", "count": 2 },
    { "name": "load_skill", "count": 1 },
    { "name": "read_file", "count": 2 },
    { "name": "spawn_subagent", "count": 1 },
    { "name": "write_file", "count": 2 }
  ],
  "conflict_results": 0,
  "spawn_statuses": [
    { "status": "completed", "count": 1 }
  ]
}
```

The run had to be stopped manually to avoid leaving local waits alive. Endpoint load was already gone because all model connections had closed.

## Expected Behavior

When background subagent child requests exceed deadline or lose stream liveness:

- The child `AgentRequest` and associated `AgentResponse` should terminalize.
- The parent `spawn_subagent` tool call should bridge to a terminal result, likely deadline/child-failed.
- Queued descendants that can no longer legally run should be terminalized/cancelled or removed from the live queue.
- Parent `response wait` should eventually return a terminal parent response or a structured terminal failure.
- Runtime status should not remain indefinitely in a state with `expired_processing_count > 0`, no endpoint connections, unchanged active child requests/tool calls, and a non-draining queue.

## Suggested Conformance Coverage

Add generated or explicit conformance cases for:

1. A background subagent child request in `processing` whose deadline has passed and whose stream/native executor is no longer live. Recovery should terminalize request/response.
2. A parent background `spawn_subagent` tool call whose child terminalizes by deadline. Recovery should bridge the parent tool call to a terminal failed/deadline result and unblock parent execution.
3. A root/parent with queued subagent children when already-running children expire. Queued children should not remain runnable forever if their parent is terminal/interrupted/deadline-expired.
4. Runtime status/liveness invariant: `expired_processing_count > 0` with no active native executors and no stream progress must converge after recovery, not remain stable across status polls.

## Notes

A transient mutation retry experiment was applied locally before this run:

- default retry budget stays at 3
- transient datastore conflict errors extend to 8 retries
- retry backoff is capped and jittered per mutation text

Focused tests for that experiment passed, and the live run no longer showed transaction-conflict child results before wedging. That suggests this issue is now the child liveness/recovery conformance gap rather than persistence retry exhaustion.

jackzampolin:
## What

Two Lean-first fixes in the halting/liveness class, plus the conformance fences for them.

### 1. Multi-tool-result accumulated-turn persistence (runtime bug)

With several parallel tool calls accumulated in ONE assistant turn, the first streamed result persisted the turn but its tool-result user message reset the hook's `TranscriptTurnState` to `Idle` — so the second streamed result tripped the `hook.rs` "cannot persist streamed tool result before its assistant turn is persisted" guard and the request failed.

The Lean Transcript model adjudicates: `completeToolWithResult` never removes a reservation held by a **persisted message**, so after `persist_assistant` every remaining parallel result completes independently (new theorem `Transcript.parallel_results_complete_independently` + preservation lemmas). The guard itself is correct — it mirrors the coherence requirement that forbids results before the turn persists; the reset was the bug.

- Tool-result user messages no longer reset the turn state (only genuine user messages do).
- `persist_assistant_turn` treats `AssistantPersisted` as start-of-new-turn (text-only final turn after tool results) and is now infallible.
- Un-ignored `multiple_streamed_tool_results_share_one_accumulated_assistant_turn`; new `parallel_results_share_assistant_turn` executable case row + conformance driver (transcript cases 6 → 7).

### 2. Periodic subagent-liveness recovery sweeps (#465)

Field wedge: high-fanout parent, background subagent children die past the verifier deadline, runtime keeps reporting them active (`expired_processing_count: 16`, zero endpoint connections), parent response-wait blocks forever, queue never drains.

Every needed transition already existed (`terminalize_expired_local_child_request`, `recover_bridge_terminal_child`) but ran **only at startup**; the 5s `BackgroundCompletionObserver` tick only projected already-terminal children. Without a restart the wedge is permanent.

- **Lean**: `RecoveryCadence.periodic` + `Proofs/Recovery/Sweeps/SubagentLiveness.lean` registers two sweeps under the proven sweep contract (measure strictly decreases to zero; recovery equals the uninterrupted terminalization path):
  - `expiredSubagentChildSweep`: claimed/processing child + expired deadline → `dead` (a live executor enforces its own deadline, so an expired non-terminal row means the executor is gone)
  - `queuedDescendantSweep`: pending spawn descendant + terminal parent → `interrupted`
- **Rust**: `ToolCallLifecycle::reconcile_subagent_liveness` on the observer's reconciler tick — terminalizes expired local children, immediately projects their **background** bridges to terminal failed (foreground bridges belong to their live waiter), and interrupts queued spawn descendants of terminal parents. Scope guard: only requests referenced by an `AgentToolCall` bridge qualify — wake/steering queue rows carry spawn lineage but must survive a terminal caller (pinned by a bystander assertion).
- **Conformance** (recovery sweep cases 19 → 23): all four cases from #465 —
  1. expired processing/claimed child terminalizes (`expired_*_child_to_dead`)
  2. dead child projects the parent bridge to terminal failed (`tool_running_child_dead_to_failed` + driver assertions)
  3. queued descendant of terminal parent interrupts; lineage-only bystander survives
  4. status convergence: the expired-processing measure (mirror of the CLI `expired_processing_count`) reaches zero in one tick and reconciliation is a no-op thereafter

## Gates

- `lake build` clean, zero sorrys
- `cargo test -p defra-agent`: 671 lib + 201 conformance + all integration binaries green
- `cargo test -p defra-agent-cli`: green

Closes #465.


iverc (approved):

jackzampolin:
Two follow-up commits since the PR opened:

**f1d7573a — self-review findings applied** (7 finder angles + adversarial verification):
- Fixed a flag-gated cross-deployment liveness gap: the queued-descendant sweep looked the parent up filtered by the local `agent_did`, so a pending child of a replicated *remote* terminal parent was skipped forever. The parent is now resolved by `request_id` alone (a visible terminal row releases the child regardless of owning DID; an unreplicated parent still conservatively skips). Regression case added.
- Scoped the 5s tick server-side (`child_request_id != ""` on the bridge scan, `caused_by_parent_tool_call_id != ""` on the pending scan) instead of full-table scans.
- Model-fenced the wake/steering scope guard: Lean `QueuedDescendantRow` now carries the `bridgeLinked` discriminator in its stale predicate.
- Refuted during verification (no change): the foreground-waiter race (the waiter detects the dead child and fails its own bridge cleanly), the `on_completion_call` reset concern (generator/consumer strictly serialized), and double projection (`AlreadyProjected` guard).

**e3866a4b — all four #446 structure-fence gaps closed.** The #446 backlog proper was verified consumed (its closing comment was right); the fence gaps were the real residue:
- **InferenceCall**: new DB-backed conformance witness — seeds each generated slot-accounting case as real `InferenceCall` rows, reads them back, reconstructs running counts exactly as admission does, and pins the S7 capacity bound over the persisted projection. Promoted to the dataset's registered ledger consumer (two-sided swap).
- **RuntimeReconcile**: conformance consumer extended to all 6 cases + a completeness pin.
- **ManagedExec**: case-shape consumer moved to its own module home (id unchanged).
- **StreamingResponse**: stale gap — the model is fenced in `streaming_compaction.rs` (incl. a daemon-driven idle-timeout case) and the idle-deadline precondition is a registered boundary; #450 landed the configurability the gap was pending on.

The structure fence's only remaining declared gap is Skills (#460). Gates: `lake build` zero-sorry clean; `cargo test -p defra-agent` (202-test conformance binary + 671 lib + integration) and `-p defra-agent-cli` fully green.

jackzampolin:
**6a3828fc — the mirror is now 1:1.** Every Lean model under `Proofs/` maps to its own conformance module (the structure fence's table reads as a clean index):

- `tooling_slots_queue_command.rs` dissolved into `command_policy.rs`, `fleet.rs`, and moves into `tool_execution.rs` (preflight/retry) and `request_lifecycle.rs` (R4a queue/deadline rows, which project from Request/Session executables).
- `transcript_background.rs` split into `transcript.rs` + `background.rs`.
- Two new shape consumers give MCPHealth (`mcp_health.rs`, pinning the K=1 healthy→evicted collapse and the K≥2 future rows) and Process (`process.rs`, machine completeness + 25-case partition) their own conformance-binary homes; the behavior drives stay in their registered unit suites.
- PairingReconcile + ReversePairingHandlers documented as deliberately co-homed (one two-node scenario harness exercises both).
- Wrapper test names (= ledger consumer ids) unchanged → zero ledger churn.

Also batched the periodic sweep's child-liveness reads into one `_in` query per tick (review follow-up), filed #477 for the periodic-sweep registry generalization, and closed #445 (verified sealed by `sanitize_history_for_provider` + the PromptAssembly fence + the registered boundary).

Gates: `lake build` clean, full `defra-agent` (11 binaries incl. the 204-test conformance binary) + `defra-agent-cli` green.

jackzampolin:
**829666c7 — full-branch multi-agent review pass.** Seven finder angles + adversarial verification over the complete 5-commit diff.

**Headline:** the removed-behavior auditor reconstructed both deleted files from the diff and verified the mirror split was **lossless** — every function, helper, struct, and test from the 3,144 redistributed lines is accounted for, and the ledger swap left no dangling references. **Zero correctness bugs survived verification.**

Confirmed cleanups applied in this commit:
- The queued-descendant bridge-existence scope guard is now ONE batched `_in` query per tick (`load_bridged_child_ids`) — combined with the earlier child-liveness batching, the 5s reconciler tick issues **no per-row queries at all** (bridge scan + child batch + pending scan + cached parent lookups + bridge batch).
- The inference slot-accounting drive uses one EmbeddedNode for all 11 cases (was 11), rows namespaced by `call_id` prefix — ~10–30s off the suite.
- `load_child_liveness_row` → `load_request_liveness_row` (it loads the PARENT in the queued-descendant sweep since the cross-deployment fix).
- Two compiler-verified unused imports dropped; structure-fence doc refreshed; `bridgeLinked` doc in SubagentLiveness.lean now states it's derived at sweep time via the bridge-existence query, not persisted.

Refuted with constructive evidence (no change): two "compile-breaking missing mod" claims (disproven by the passing 204-test build), DefraDB default-limit truncation on unlimited queries (contradicts every existing sweep's reliance on full results), `_ne`/NULL semantics divergence (Rust-side guards deliberately retained on both query paths), serde-default `request_id` key collision (an empty key is never consulted).

Noted for later (declined now, same family as #477): the structure fence verifies a model's home file exists but not that it consumes that model's accessors — a possible future hardening.

Gates: `lake build` clean; full `defra-agent` (11 binaries, 204-test conformance binary) + `defra-agent-cli` green.

Issue #475, opened 2026-06-10
Codex shim: messages queued during an active turn are never flushed


jackzampolin:
## Symptom

In the Codex TUI connected to the shim (`defra-agent codex` or stock `codex --remote`), a user message typed **while a turn is in progress** is rendered as queued (`›` history line) but is never dispatched after the turn completes. The shim event log shows no `turn/start` for it, and no AgentRequest document is created — the message is silently lost. Messages typed while the TUI is idle dispatch fine, and the stuck queue persists even across later successful turns.

## Repro (observed during demo-rework e2e, 2026-06-10)

1. `defra-agent server` + `defra-agent codex` (branch `demo-mac-local-inference`, but the shim path is unchanged from main)
2. Submit a prompt; while the turn is working (or within ~a second of completion), type a second prompt and press Enter
3. Second prompt shows in history, queued; after turn completes it never runs. Subsequent prompts then also queue forever once the TUI is in this state.

Shim event log evidence: turns 5 and 6 ran (`turn/started`/`turn/completed`), then three queued user messages produced zero incoming requests over 20+ minutes on an idle machine.

## Suspected cause

Codex's TUI flushes its queued-message buffer on some app-server signal that the shim does not emit (its `turn/completed` carries `turn_items: []`, and the item event sequence differs from the real app-server). Needs a diff of the real `codex app-server` turn-completion notification stream vs the shim's.

## Impact

Demo-critical UX: users naturally type follow-ups while the agent works; those messages disappear silently. Workaround: retype the message once idle (it shows queued but a newly typed message dispatches), or restart `defra-agent codex`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #408, opened 2026-06-05
Add run timeline and adapter projection interop coverage

jackzampolin:
## Summary

First #407 projection/adapters implementation. Runtime documents stay the source of truth; a reusable run timeline projection is reconstructed over them, and adapter-facing projections export OpenAI/Codex-style traces, LangGraph state/history, and multi-agent task/delegation shapes — with redaction modes and DefraDB-ACP-gated row/content filtering.

Includes real native-capture import/export roundtrip coverage for Dockerized LangGraph, AutoGen AgentChat, CrewAI, and Microsoft Agent Framework scenarios: the E2E suite validates actual `defra-agent trace project` JSON/JSONL/eval-JSONL outputs against projection contracts and framework-specific semantic expectations.

## What is included

- Reusable `RunTimeline` DTOs and `build_run_timeline` (4-tier request-id inference, child-request discovery).
- `defra-agent trace timeline --request-id`, `trace project --projection <openai-codex|langgraph|multi-agent> --format json|jsonl|eval-jsonl`, and `trace project-schema`.
- Redaction modes (`full`, `training-safe`, `public`) exercised end-to-end; ACP fail-closed root access with row/content filtering and denied-row CLI coverage.
- `ProjectionAcpBinding` collection: schema, discovery, config import/export, desired-state validation (ApplyReconcile Lean model extended — zero `sorry`s), lifecycle metadata, policy lifecycle coverage.
- Dockerized external fixture generators (LangGraph linear/subgraph/provider, AutoGen round-robin/Swarm, CrewAI sequential/hierarchical, MSAF group chat) + native capture importer + framework verifiers checking semantic equivalence against native captures.
- An ignored CLI E2E that imports mapped native captures into embedded DefraDB, runs the real binary, and validates every export artifact; `run_docker_interop.sh` requires exactly 8 fixtures and 24 exports.
- Gated/manual `Adapter Interop` GitHub Actions workflow.
- `AgentMessage.request_id` so imported/persisted message rows retain the request boundary for child-agent projections.

## Rebase onto post-cleanup main (2026-06-10)

The branch was merged with main (69 commits: #426 owned loop, #451 native message family, #459 docs purge, #463 test-binary consolidation, Ivan's stack) and restructured to match current conventions:

- **No resurrected `docs/superpowers/`.** Contract schema snapshots → `crates/defra-agent/tests/fixtures/adapter_projections/contracts/`; envelope fixtures → `.../envelopes/`; Docker generators + runner → `scripts/adapter-interop/`.
[...]

PR #476, opened 2026-06-10
Demo rework: Mac local inference with embedded Codex TUI

jackzampolin:
## What this is

One tight getting-started story — a fresh Mac, local inference, a native Defra agent that can change your computer, driven through the Codex terminal UI with nothing else installed:

```bash
brew install llama.cpp
llama-server -hf google/gemma-4-12B-it-qat-q4_0-gguf
cargo install --profile dev-install --locked --path crates/defra-agent-cli
defra-agent init        # read-only; --write = sandboxed writes; --yolo = unrestricted
defra-agent server      # codex shim now on by default (loopback)
defra-agent codex       # embedded Codex TUI — no Codex install, no flags
```

Design spec: `docs/superpowers/specs/2026-06-10-mac-local-inference-demo-design.md`.

## Changes

- **`defra-agent codex`**: runs `codex-tui` in-process at the existing pinned rev, pointed at the shim with Codex-side approvals/sandboxing bypassed — the init-time tool preset is the real permission boundary. Probes the shim socket and gives a friendly error when the server isn't running. Requires mirroring codex-rs's `[patch.crates-io]` crossterm/ratatui forks into our workspace (cargo doesn't carry patch sections across git deps); unicode-width pinned to 0.2.1 for the ratatui fork.
- **init presets**: `--write` (renamed from `--write-tools`, which stays as a hidden alias) keeps the seatbelt-sandboxed write package; new `--yolo` package is ReadWrite + unrestricted bash on every platform, with a loud warning. No Lean change: config plumbing only, existing tool-selection conformance covers the document shapes.
- **server**: Codex endpoint on by default, loopback-bound (refuse-unspecified-bind guard unchanged); `--no-codex-shim` opts out. Test helpers auto-append `--no-codex-shim` unless a test configures the shim explicitly, so parallel test servers don't race the fixed port.
- **defaults**: ollama → llama-server (`http://127.0.0.1:8080/v1`, `google/gemma-4-12B-it-qat-q4_0-gguf`); the `ollama` preset keeps 11434. init next-steps and help text updated.
- **docs**: README "Get running" + `docs/demo.md` rebuilt around the single story (permission-preset guarantees table included); desktop/fleet/P2P content moved to `docs/operations.md`; demo manifest now points at the local backend; `scripts/install-codex.sh` deleted.

## Verification

- `cargo test -p defra-agent && cargo test -p defra-agent-cli` — full package suites green (44 suites, 0 failures; one known-flake `lean_apply_write_boundary` passed on isolated rerun during an interim run).
- Real e2e on this Mac: llama-server with Metal at 96% GPU (~55 tok/s); embedded TUI created a file via a write tool call and read it back; out-of-tool-root write rejected by the runtime with the exact tool-call error; `defra-agent chat` one-shot fallback works.
- Found pre-existing shim bug while testing, filed as #475 (messages typed during an active turn queue forever and are silently dropped). Not introduced or worsened by this PR; demo-relevant.


jackzampolin:
Addressed the multi-agent review findings (7 finder angles → 1-vote verify; 7 confirmed):

- **Shim failures no longer kill the server** (`08d1c0cf`): bind/validation errors (port taken, behavior missing an inference profile — including homes provisioned from `demo/manifest`) degrade to a disabled Codex endpoint with guidance, and the runtime keeps serving. Regression tests cover both the occupied-port and missing-behavior paths; the old refuses-to-start test was updated to the new contract (`107365e6`).
- **Ollama preset is coherent again**: model default is now per-preset — `--backend-preset ollama` resolves to `hf.co/google/gemma-4-12B-it-qat-q4_0-gguf` (same model, valid Ollama pull form), verified end-to-end.
- **IPv6 shim addresses are bracketed** in the advertised URL (`ws://[::1]:9292/`), so the printed `--remote` hint is actually usable.
- **`DEFAULT_CODEX_REMOTE` is pinned** to `DEFAULT_CODEX_SHIM_PORT` by a unit test; the two can no longer drift silently.
- **Yolo's tool profile derives from Write** via struct update — the surfaces cannot diverge by edit-miss.
- rustfmt (`f991e3fc`) — this was the CI `rust-and-cli` failure.

Deferred as follow-ups rather than PR blockers: a `codex` cargo feature to gate the ~368-package codex-tui dependency tree out of non-demo builds, and read-side tolerance for unknown `tool_package` values in init.json (old binary reading a `--yolo` home fails to parse it). Review also confirmed two non-issues: the embedded TUI's Arg0 paths are correct for remote mode, and the default-on shim adds no meaningful local attack surface beyond the already-unauthenticated GraphQL endpoint.

Full local gates: `cargo test -p defra-agent && cargo test -p defra-agent-cli` — 44 suites, 0 failures.

iverc (approved):

jackzampolin:
Coordinator review: scope and content look right — CLI/docs/demo only, runtime untouched, no-Lean-change claim holds, codex deps pinned at the existing rev with the patch-section constraint documented, full suites green.

One convention fix before merge: `docs/superpowers/specs/2026-06-10-mac-local-inference-demo-design.md` resurrects the `docs/superpowers/` tree that #459 deliberately purged — the #408 rework explicitly avoided recreating it. Per the curation principle (design rationale → issue comments; durable operator docs → `docs/`): either fold what's durable into `docs/demo.md` and drop the spec file, or post it as a comment on the demo-rework tracking thread and delete it from the tree. Happy to squash-merge as soon as that's relocated.

Issue #494, opened 2026-06-13
Refactor: CodexThreadProjection should derive from loaded tools/runtime state, not its own collection


jackzampolin:
## Problem
`CodexThreadProjection` is misnamed and over-modeled. Despite "Projection" in the name it is **not** a derived view — it's a dedicated replicated collection holding original Codex-TUI session sidecar state: `cwd`, `name`, `archived`, `loaded`, `memory_mode`, `settings_json`, `goal_json`, `rollback_user_turn`, `git_info_json`. Written/read only by the Codex shim (`crates/defra-agent-cli/src/commands/codex_shim/thread_projection/`).

## Desired
These values should be sourced from the **loaded tools / runtime state** (e.g. `cwd`/git from the loaded file-tool root + runtime, goal/settings from session/behavior state) rather than persisted in a standalone collection. Eliminate the dedicated `CodexThreadProjection` collection (or, if some fields genuinely have no other home, rename to honest `CodexSessionState` and keep it Codex-shim-local — but the goal is to derive, not store).

## Why now (context)
Surfaced during the scope-templates / filtered-replication work (PR #490). It was excluded from the `conversation` replication template because it's local Codex-UI state that a delegating peer never needs. This refactor is **separate** from #490 — do not bundle.

## Scope
- Audit each field's true source; derive where possible.
- Remove the collection + its migration + the `thread_projection/storage.rs` persistence, or downscope to a renamed local-only record.
- Update the codex_shim thread/turn projection to read from live state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #500, opened 2026-06-15
Derive Codex thread cwd from runtime state

iverc:
## Summary

- stop writing `cwd` as authoritative state in `CodexThreadProjection`
- derive Codex thread cwd from connection/runtime state, latest request metadata, thread settings, legacy projection rows, then shim default cwd
- thread list/read/search/fork/metadata routes now pass cwd hints through projection loading
- update tests to assert client-visible cwd instead of persisted projection cwd

## Notes

This is a scoped slice of #494. It removes `cwd` from the projection's owned state, but does not fully remove or rename `CodexThreadProjection` yet because it still stores Codex UI-local fields like archived/loaded/name/memory mode/settings/goal/git metadata.

Refs #494

## Tests

- `cargo test -p defra-agent-cli commands::codex_shim::thread_projection::storage::tests -- --nocapture`
- `cargo test -p defra-agent-cli --test cli_codex_shim codex_shim_protocol_turn_streams_defra_response -- --nocapture`
- `cargo test -p defra-agent-cli --test cli_codex_shim codex_shim_thread_fork_and_search_project_defra_sessions -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/defra-agent-cli/src/commands/codex_shim/handlers/thread.rs
+++ b/crates/defra-agent-cli/src/commands/codex_shim/handlers/thread.rs
@@ -12,7 +12,7 @@ use super::super::protocol::{
 };
 use super::super::thread_projection::{
     clear_codex_thread_goal, codex_thread_json, codex_thread_json_with_turns, create_codex_thread,
-    ensure_agent_session_pinning, get_codex_thread_goal, load_codex_thread,
+    ensure_agent_session_pinning, get_codex_thread_goal, load_codex_thread_with_cwd_hint,
     loaded_codex_thread_ids, resume_codex_thread, set_codex_thread_archived,
     set_codex_thread_git_info, set_codex_thread_goal, set_codex_thread_loaded,
     set_codex_thread_memory_mode, set_codex_thread_name, set_codex_thread_settings,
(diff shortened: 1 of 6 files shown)


jackzampolin:
Thanks @iverc — flagging that the codex-projection worktree (#505) took the full swing at #494 in parallel, and it subsumes this one, so wanted to explain rather than just redirect.

**The relationship:** this PR derives `cwd` only and explicitly keeps `CodexThreadProjection` for the other fields (archived/loaded/name/memory/settings/goal/git) — "Refs #494", a scoped slice. #505 does what the issue actually asked for: **removes the collection entirely** and derives every field from scoped `AgentSession` + `AgentConversation` + runtime state (`rg CodexThreadProjection` comes back clean). So the cwd-derivation you did here is a strict subset of #505 — same direction, full distance.

I checked the concern I'd have flagged on a full removal — the ApplyReconcile fence requiring Lean to cover every collection — and #505's session already audited it: CodexThreadProjection isn't in the config-import/desired-state path, so no Lean change is needed, and lean-proofs is green. So the full removal is clean.

**Carry-over to verify in #505:** your cwd precedence chain (connection/runtime → latest request metadata → thread settings → legacy projection rows → shim default) is a nice fallback ladder — worth confirming #505 preserves that ordering rather than collapsing it, since the legacy-rows fallback matters for already-provisioned homes.

Proposing we close this as superseded once #505 lands. Would value your review of #505 — you know the shim thread routes better than anyone.

iverc:
Closing as superseded by #505, which removed CodexThreadProjection entirely and carried over the durable cwd recovery behavior.

PR #501, opened 2026-06-15
feat(desktop): skills surface — view, manage, and bind skills

jackzampolin:
## Summary

- add desktop Skill deletion end-to-end through desktop-core, Tauri, TS API/actions, and the Skills panel
- remove deleted skills from behavior `skill_refs` / `skill_excludes` so stale bindings do not silently reactivate later
- add desktop-core coverage for skill save, behavior binding, delete, and binding cleanup
- update stale desktop test fixtures for `AgentMessageRow.reasoning`
- update desktop P2P test/live fixtures for the current filtered-replication `filters` API

This keeps the original skills surface, but makes the desktop side closer to true CRUD and removes the previously noted desktop test/bin compile drift.

## Testing

- `npm run build`
- `cargo check -p defra-agent-desktop-core --tests`
- `cargo check -p defra-agent-desktop-tauri --tests`
- `cargo test -p defra-agent-desktop-core --test manage_view manage_document_saves_refresh_store -- --nocapture`
- `cargo fmt --check`
- `git diff --check`

## Notes

`cargo check -p defra-agent-desktop-tauri --tests` still emits one pre-existing dead-code warning for `AgentRequestRowLite.request_id`.


jackzampolin:
Nice work completing the CRUD — the scoped hard delete + the DB cascade that rewrites affected behaviors' `skill_refs`/`skill_excludes` (`writes.rs:343`), and the `manage_view.rs` integration test covering add → bind → delete → ref-cleanup, are all solid and green.

That integration test proves the **data layer**. What I want before this lands is true **desktop-env e2e** against a live agent node — two tests, written and **passing green in this branch**. The home for both is `LiveBridgeFixture` (`apps/desktop-tauri/src-tauri/src/runner/live_fixture.rs`), which already stands up a desktop `ClientCore` + a real agent node with P2P replication but has **no tests yet**. Note the determinism requirement up front: the fixture resolves a real backend today (`live_fixture/backend.rs`) — these tests must run hermetically in CI via a mock model wired through `LiveBackendOverride` (the `MockChatEndpoint` pattern from the CLI tests), **not** an `OPENROUTER_API_KEY`. Wiring that mock into the fixture is part of the work.

**E2E test 1 — create then delete a skill from the desktop, verified on the agent node.**
- Drive the **real desktop code path** — `ClientCore::save_skill` / `ClientCore::delete_skill` (i.e. the `desktop_skill_save` / `desktop_skill_delete` bridge commands), not raw `create_Skill`/`delete_Skill` GraphQL. The point is to exercise the CRUD you wrote, including validation and the cascade.
- Assert on the **agent node** (`remote_core`), not just the desktop store: after create, the `Skill` doc replicates through and is queryable on the node; after delete, it's gone from the node; and if it was bound to a behavior, that behavior's refs are cleaned on the node too.
- Acceptance: desktop create → `Skill` present on agent node → desktop delete → `Skill` absent on agent node + behavior refs pruned, all observed by querying the node's DefraDB.

**E2E test 2 — start a conversation with `/{skill_name}` and confirm the skill loads, verified on desktop and server.**
- Seed a behavior-scoped skill that's in the behavior's effective set, then start a conversation by sending chat content `"/{skill_name}\n<task>"` through the desktop chat path.
- **Server-side (agent node):** the `AgentRequest` for that turn has `selected_skill_ids = ["{skill_name}"]` in metadata, the `/{skill_name}` prefix stripped from `content`, and — the part that actually proves it *loaded* — the skill's instruction body appears in the rendered model request (capture it via the mock model and assert the skill body is present in the turn's system/reminder text). Parsing happens in `prompt_slash_skill_selection` (`skills.rs:275`) and injection at request time (`prompt.rs` skill reminders); assert on the durable signal, not the parser in isolation.
- **Desktop-side:** the conversation/turn snapshot reflects that the skill was selected/loaded for the opening turn.
- Acceptance: `selected_skill_ids` persisted on the node + skill body present in the model request + desktop turn view shows the skill loaded.

One honest caveat for test 2 so it's not a surprise: the runtime `/`-activation path exists from the earlier slice, but the desktop chat composer has **no** slash affordance (autocomplete/validation/feedback) — this PR doesn't touch it. The test drives `/{skill_name}` as raw content, which is the real round-trip, so the test is valid; whether to add composer UX is a separate call we can track as a follow-up.

Both tests green in this branch's CI (deterministic mock model) is the bar. Happy to pair on the `LiveBridgeFixture` mock-model wiring if that's the fiddly part.

iverc (approved):

jackzampolin:
The latest revision addresses the E2E coverage bar I wanted for the skills desktop slice.

The new live fixture coverage now exercises both important paths:
- desktop skill create/delete through the bridge/core path, verified on the live agent node with behavior refs pruned after delete
- starting a session via raw `/{skill_id}` content through `send_chat_message`, then verifying the desktop pending turn, the replicated agent-node `AgentRequest` content/metadata, and the rendered mock model request containing the selected skill body

I also ran the targeted checks locally on `59048cde`:
- `cargo test -p defra-agent-desktop-tauri live_fixture_desktop_chat_slash_skill_loads_on_agent_node -- --nocapture` passed
- `cargo test -p defra-agent-desktop-tauri live_fixture_replicates_skill_create_delete_to_agent_node -- --nocapture` passed
- `cargo test -p defra-agent-desktop-core --test manage_view manage_document_saves_refresh_store -- --nocapture` passed
- `cargo fmt --check` passed
- `git diff --check HEAD` passed

The live fixture emits noisy P2P authorization/shutdown warnings locally, but the assertions complete and both live tests pass.

PR #503, opened 2026-06-15
Capture rendered completion requests for training

### LINKED ISSUE #499: Surface the resolved completion request for training capture
## Motivation

zanzibar-rl captures one training row per agent turn and needs each row to record the **exact** prompt + tool schemas the model was conditioned on — for zero train/serve skew. That conditioning is assembled inside defra-agent (`build_request`) and is never surfaced to the embedding host; reconstructing it in policy-agent would be a re-derivation that drifts from what the model actually saw.

Expose a per-turn projection of the resolved completion request **in-memory**, at the request-build boundary.

## Key fact: the agent loop runs in the host's process

policy-agent embeds defra-agent and runs the agent loop itself: it builds a `DefraAgent` via `DefraAgent::builder()` and does `tokio::spawn(agent.run(...))` in its **own process** (zanzibar-rl `policy-agent/src/policy_runtime/lifecycle.rs:111`). The DB is only the *data* path (submit `AgentRequest`, poll `AgentResponse`); `build_request` itself executes in the host's shared Rust address space. So a simple **in-memory sink** reaches the host directly — no DefraDB persistence, no new collection, no `wait.rs` read-path changes.

## Where the request is born

`crates/defra-agent/src/agent/loop_stream.rs:519-567` (`build_request`) — the per-turn chokepoint. The resolved `rig::completion::CompletionRequest` has:
- `chat_history` = `[System(preamble), …history, …prior, prompt]` (preamble from `prompt.rs:261` `build_preamble_with_targets`; rig's `build()` leaves `preamble: None` and folds the system text into `chat_history[0]` — **read `chat_history`, not `preamble`**),
- `tools` = per-turn `ToolDefinition` (`{name, description, parameters}`),
- `tool_choice`, `temperature`, `max_tokens`, `additional_params`.

## Proposed mechanism

Add an opt-in sink to the agent builder / `LoopConfig`, fired once at the end of `build_request` (after `builder.build()`, before `model.stream`):

```rust
// on DefraAgent::builder() -> threaded into LoopConfig
on_rendered_request: Option<Arc<dyn Fn(&str /*request_id*/, usize /*turn_index*/, &CompletionRequest) + Send + Sync>>,
```

It is per-turn, opt-in (`None` for every existing caller), and behavior-neutral (observes `&CompletionRequest` after it is fully built; cannot mutate it). The `request_id` is in scope at the `run_loop_stream` call site (`agent/daemon/inference.rs:222`) and is threaded into `loop_config_for_request` (`completion_factory.rs:65`) so the sink can key by it. The embedding host (policy-agent) passes a closure that pushes the projection into its own `Arc<Mutex<HashMap<request_id, Vec<…>>>>` and drains it after the response is terminal.

Rejected: (b) attaching to the returned result — the streaming path returns no in-memory response object; (c) a new stream-event variant — the loop deliberately yields rig's foreign `MultiTurnStreamItem` unchanged (decision D3) and must not fork that contract. DB persistence is unnecessary given the in-process embedding.

## Serialization (the projection the host records)

`{ request_id, turn_index, messages, tools, tool_choice, sampling, prompt_hash, tools_hash, source }`
- `messages`/`tools` in the **provider (OpenAI) JSON shape** — the actual wire body — via the provider `TryFrom<CompletionRequest>` conversions (`providers/openai/completion/mod.rs`), not the rig-native enum.
- `prompt_hash`/`tools_hash`: SHA-256 over **canonical (sorted-key) JSON** of `messages`/`tools`, for skew detection.
- `sampling` mirrors `temperature`/`max_tokens`/`additional_params` (capture `additional_params` verbatim — it carries `enable_thinking`/`reasoning.effort`).

The host may serialize the projection itself from `&CompletionRequest`; defra-agent only needs to expose the borrow at the right moment. Whether the provider-shape conversion is done in defra-agent (preferred, single source) or by the host is an implementation detail to settle in the PR.

## Dependency

Pairs with #497 (cache-safe prompt templating): #497 renders the dynamic context, this issue captures whatever was rendered.

iverc:
## Summary

- add an in-memory rendered-request capture hook to the owned completion loop
- expose a `DefraAgent::builder().rendered_request_capture_factory(...)` path so embedding hosts can collect per-turn provider requests by `request_id`
- keep provider wire-shape conversion inside `llm::rig_compat`, preserving the rig boundary seam
- capture the point-in-time provider JSON, messages/tools/tool-choice/sampling JSON, and canonical SHA-256 prompt/tool hashes before provider streaming starts
- avoid adding a persisted `AgentRenderedRequest` collection or replication surface; the RL trainer can fold captured bytes into its existing JSONL rows

Closes #499.

## Tests

- `cargo test -p defra-agent rendered_request`
- `cargo test -p defra-agent docs::rig_vocabulary_confined_to_the_seam`
- `cargo test -p defra-agent loop_stream::tests::rendered_request_sink_runs_before_provider_stream`
- CI: `rust-and-cli` pass
- CI: `lean-proofs` pass
- CI: `CodeRabbit` pass


jackzampolin:
**Two things before this lands.**

**1. The red check (`docs::rig_vocabulary_confined_to_the_seam`) is the right alarm, not a flake.** `rendered_request.rs` reaches into rig's Layer-A provider vocabulary directly — the prod path calls `rig::providers::openai::…::CompletionRequest::try_from` / `rig::completion::CompletionRequest`, and the test module imports `rig::completion::message::*` + `rig::one_or_many`, which is what trips the marker. Per CLAUDE.md the rule is one seam (`llm::rig_compat`) for all rig↔runtime crossings, and this is the first place in the runtime that serializes a provider *request body*. The proper fix is to move the provider-shape conversion (`provider_request_json`) **into `llm::rig_compat`** and call it from here, so all provider-wire knowledge stays in the one seam — not just allowlisting the file. (Allowlisting works mechanically, but it erodes the "one seam" claim the fence exists to protect.)

**2. I don't think this should be a DefraDB collection at all.** `AgentRenderedRequest` is a single-consumer export projection in OpenAI-wire shape — nothing in the runtime ever reads it back to make a decision; it only flows outward to the zanzibar-rl trainer, which consumes JSONL files (`load_jsonl` / `--training-out *.jsonl`), not DB rows. Persisting it as a branchable + replicated collection means canonical data → doc → GraphQL → P2P gossip → `dump_jsonl`: three hops of overhead, O(turns²) storage duplicating data already in `AgentMessage`, and we now gossip the most sensitive payload in the system (full system prompt + history, wire shape) over P2P by default. This is the same call we already made deleting `CodexThreadProjection` (#494/#505): projections are reconstructed/exported from durable canonical data, not stored.

The one thing that *is* genuinely different from the Codex thread — and where the capture instinct is right — is that the rendered request can't be fully reconstructed later: it's a point-in-time snapshot of a deterministic conversion over inputs (template, tool surface, sampling, `additional_params`) that can themselves drift. So the **capture has to happen at build time**, and the sink you added is exactly the right hook. The change I'd make is where the captured bytes *land*: write the sink through to a **durable export artifact** — a JSONL appender, or feed it into the existing `run_timeline` / `adapter_projection` export seam as an on-demand `trace export-rendered` — rather than a new collection. That keeps the skew-proof build-time capture, drops the storage bloat and replication surface, and lands the data in the format the trainer actually wants.

Net: keep the sink + the provider-shape conversion (moved into the seam); drop the `AgentRenderedRequest` schema, the replication-template wiring, and the turn-state exposure in favor of a file/export sink. Happy to pair on the seam refactor if useful.

jackzampolin:
One addition that makes the rework concretely *smaller* for the in-process consumer:

The main consumer (zanzibar-rl training) embeds the agent and **already writes the per-episode `--training-out` JSONL row itself** (`persist_training_examples_jsonl`). So for that path defra-agent doesn't need to write a file *or* a collection at all — the `on_rendered_request` sink can hand the captured bytes to the embedding host **in-memory** (the host owns the `Arc`, passes the closure into `DefraAgent::builder()`, drains by `request_id`), and the host folds `rendered_requests_v1` straight into the row it's already persisting. A defra-agent-side JSONL appender would actually be *worse* here — a second file to join by `request_id`.

So the slimmed surface is just: (1) keep the `on_rendered_request` sink, (2) move `provider_request_json` into `llm::rig_compat` (point 1), and (3) *optionally* a `trace export-rendered` on the existing export seam for non-embedded (CLI / remote) consumers. For the in-proc trainer, (1)+(2) is the whole dependency — no `AgentRenderedRequest` schema, no replication template, no turn-state exposure.

jackzampolin:
## Proposal: make this a *provable projection* of durable state, not an in-loop capture

First — the renderer here is exactly right and should stay. `rig_compat::rendered_completion_request`, the `RenderedRequestSource` wire-shape enum, the canonical SHA-256 hashing, and the `RenderedCompletionRequest` shape are all keepers. The only thing I want to relitigate is **where the render runs and how the inputs are sourced**, because there's a much stronger shape available — one that turns "train directly from the DB" from a host-side convenience into a provable property of the system.

### The reframing

This feature's real goal is: *reconstruct, with fidelity, exactly what the model was asked to do, straight from durable state.* That is a load-bearing capability for this project — training accuracy depends on it. The question is whether that fidelity is **captured** (a side-channel records the bytes) or **projected** (the bytes are a deterministic function of documents we already persist). This codebase already chose the second posture everywhere else: `run_timeline.rs` and `adapter_projection.rs` reconstruct a request's entire event stream from durable rows, and `PromptAssembly` is *formally proven* deterministic (sanitization soundness, idempotence, split-stability). The rendered request should live on that same plane.

### The determinism boundary (the evidence)

If you classify every input that feeds the rendered `CompletionRequest`, almost all of it is already durable or a deterministic function of durable state:

| Input | Class | Source |
|---|---|---|
| System prompt / preamble | **durable doc** | `AgentBehavior.system_prompt` (frozen in `LayeredPromptBuilder`) |
| Model / endpoint / backend kind | **durable doc** | `AgentBehavior.model_name`, `.backend_endpoint`, `.backend_provider_kind` |
| Message history + compaction summaries | **durable doc** | `AgentMessage`, `AgentToolCall`, `CompactionEntry` |
| `<context>` message (incl. resolved `now`) | **durable doc** | persisted exactly-once before the retry loop — see `inference.rs:172-178` |
| Sampling (temp / max_tokens / top_p / top_k) | **durable doc** | `AgentBehavior.sampling.*` + `AgentRequest` overrides, merged in `completion_factory.rs` |
| additional_params / reasoning effort | **deterministic fn** | `provider_additional_params()` / `thinking_default_params()` (effort hardcoded `"medium"`) |
| tool_choice, message ordering, cache-scope, skill reminders | **deterministic fn** | `completion_factory.rs:57`, `loop_stream.rs:84-94`, `prompt.rs:154-168` |
| Native tool schemas | **deterministic fn** | `tool.definition(rag_text)` (`loop_stream.rs:613`) |

Everything in that table is `f(documents)`, and we've **already proven `f` deterministic**. So the entire reconstruction gap is a short, enumerable list of inputs that are neither a document nor a deterministic function of one:

1. **Provider wire-shape selection** — `force_openai_chat_completions()` reads `cfg!(test)` + `DEFRA_AGENT_OPENAI_CHAT_COMPLETIONS` env (`inference_http.rs:35-49`). Decides Responses vs Chat-Completions encoding. Not persisted anywhere.
2. **MCP-resolved tool schemas** *(only for behaviors that use MCP tools)* — native tool schemas are deterministic, but MCP tool definitions come from live server introspection. **Needs verification:** is the resolved MCP surface persisted at reconcile time (durable) or fetched live per call (leak)? If live, it's the one *structured* input to snapshot.
3. **Serialization code version** — byte-exactness depends on the rig render version. This hits the *capture* approach identically (it also renders via `rig_compat`), so it's not a differentiator — but we should stamp a version for reproducibility either way.

That's the whole gap. Note what's **not** on it: `ctx.now` isn't a leak — the rendered `<context>` message that consumed it is persisted exactly-once (`inference.rs:172-178`), so you recover it by *reading the durable message*, not re-rendering.

### The crux: this makes rendered-request fidelity a *provable property*

Once the leak set is captured durably, the rendered request becomes a pure projection:

> **assembly determinism (already proven) + complete leak capture ⟹ rendered request = deterministic projection of durable state.**

That is exactly the class of theorem this repo already fences in Lean. The capture-hook approach can't state it — it's faithful-by-construction but opaque, and it can't tell you *why* the captured bytes equal what was sent. The projection approach is **faithful-by-proof**: the soundness obligation is "every input is durable-or-deterministic," which is checkable and, when it holds, guarantees the training data equals the production request. For a project whose thesis is "the database is the control plane," training fidelity should be a property of the database, not a property of whoever remembered to attach a sink.

### Concrete proposal

**1. Stamp the leak set onto `InferenceCall`** (it already keys `request_id` + `call_seq` per turn and carries `backend_config_fingerprint`):
   - `provider_wire_shape: String` — the resolved `RenderedRequestSource` (reuse the existing enum).
   - `behavior_config_version: String` — cid/version of the behavior doc at call time (or reuse `backend_config_fingerprint`'s approach, extended to cover sampling + system prompt).
   - `render_version: String` — the rig/serialization version, for byte reproducibility.
   - For MCP behaviors, *if* the resolved surface isn't already durable: a `resolved_tool_schemas` snapshot. (Resolve the verification question first — it may already be persisted at reconcile.)

**2. Export becomes a projection command**, not a host callback: `trace project --rendered <session|request>`, built on the `run_timeline` / `adapter_projection` pattern — read the durable rows, replay the proven-deterministic assembly, render via the existing `rig_compat::rendered_completion_request`. Same `RenderedCompletionRequest` output, same hashes, now a pure function of durable state. ACP redaction modes come for free, same as the existing projections.

**3. (Optional spike) Time-travel reads.** The literal "documents as-of the request's commit" path may already work: DefraDB is MerkleCRDT and supports `cid:`/`_version` time-travel query syntax, and `node.execute()` takes an arbitrary GraphQL string — this repo just never passes a temporal query. A half-day spike confirms whether the passthrough accepts it. If yes, even `behavior_config_version` reconstruction is free and we stamp almost nothing. If no, the per-call stamps above make export faithful without it (we freeze the resolved scalars at call time, which is strictly sufficient).

### Migration from this PR

~80% of the code here is exactly what the projection needs — only the *placement* and *plumbing* change:

- **Keep:** `rig_compat::rendered_completion_request`, `RenderedRequestSource` + `for_behavior_provider`, the canonical-JSON SHA-256 hashing, the `RenderedCompletionRequest`/`RenderedRequestContext` types, and the unit tests.
- **Move:** invoke the renderer at **projection time** (the `trace project --rendered` path) instead of via the in-loop `on_rendered_request` hook.
- **Drop:** the `RenderedRequestCaptureFactory` host-callback plumbing threaded through `builder → agent → daemon → runtime → loop_config`, and — importantly — the **fail-closed coupling to the hot path** (`loop_stream.rs:217-225`), where a capture failure aborts a live inference turn with a `ProviderError`. A training-data path should never be able to take down a production request; as a projection it structurally can't.

If we want the durable stamps to feed a host's JSONL trainer in-stream as well, that's a thin reader over the new `InferenceCall` fields + the projection — no hot-path coupling required.

Happy to spike the time-travel passthrough and/or sketch the `InferenceCall` schema delta + `trace project --rendered` surface if this direction lands. The determinism table above is the soundness argument; if we agree the leak set is complete, the rest is mechanical — and we get training-from-the-DB as a proven projection instead of a best-effort capture.

PR #504, opened 2026-06-15
Add cache-safe prompt context templates

### LINKED ISSUE #497: Prompt templating for dynamic node/agent context (cache-safe)
## Motivation

A defra-agent behavior (e.g. zanzibar-rl's `policy_agent`) needs to inject **dynamic context** into each request — acting seat / agent did, a live collection summary, time, and other node state — so the model reasons against the current world.

Today that dynamic context is concatenated into the system prompt at build time. Two problems:

1. **Prompt caching.** A system prompt that changes per request (seat, collection counts, time) blows the prefix cache every request.
2. **Train/serve fidelity.** Whatever assembles the prompt at serve time must be identical to what assembled it during training-data capture, or the model is trained and served on different conditioning.

## Proposal

A templating layer that distinguishes, by message role, what may vary per request:

- **system** — must stay byte-stable across requests. Only static or **run-constant** substitutions allowed (filled once at runtime start, then frozen). This keeps the cacheable prefix stable.
- **user / context** — per-request substitution is fine. Dynamic values (seat, collection summary, time, node state) render here.

Requirements:
- Templates declared by the behavior; the defra-agent runtime fills them per request.
- A clear constraint/validation that system-role templates reject per-request-varying values (cache-safety guard).
- Expose the **rendered** result so downstream consumers (training-data capture) can record exactly what the model saw.

## Why it matters

zanzibar-rl is consolidating all prompt/behavior into the Rust behavior (the same binary serves and is trained), and capturing the rendered prompt + tool schemas into each training row for zero train/serve skew. This templating layer is the cache-safe mechanism for the dynamic portion of that prompt.

iverc:
## Summary

Closes #497.

Adds behavior-level prompt context templates with a cache-safe split:

- `system_context_template` renders once into the behavior preamble using stable run context like `agent.did`, `behavior.id`, and `model.name`
- `request_context_template` renders per request as a system reminder using dynamic context like `request.id`, `request.metadata`, and `time.utc_now`
- validation rejects `request.*` and `time.*` references in system-context templates so dynamic values cannot leak into the cached preamble

This also wires the new fields through AgentBehavior schema/document rows, config import/export, desired-state manifests, CLI behavior writes, and startup migrations for existing databases.

## Testing

- `cargo check -p defra-agent`
- `cargo check -p defra-agent --tests`
- `cargo check -p defra-agent-cli --tests`
- `cargo test -p defra-agent --lib prompt_context`
- `cargo test -p defra-agent --test e2e_runtime agent_behavior_migration`
- `cargo test -p defra-agent-cli writes_principal_and_per_doc_dirs_with_sidecars`
- `cargo test -p defra-agent-schemas`
- `cargo check -p defra-agent-protocol`
- `git diff --check`


jackzampolin:
Thanks for this, @iverc — closing the loop because a second take on #497 landed in parallel from the prompt-templating worktree (#506), and we're going to converge on that one. Wanted to give you the full reasoning rather than just redirecting.

**Why #506 is the one we'll land:** #497 changes what the runtime feeds the provider, which is PromptAssembly territory — so by the foundation flow it has to start in Lean (model → conformance → Rust), not land as Rust-first. #506 carries the `Proofs/PromptAssembly/Template.lean` model + Executable mirror + a conformance fence (`prompt_template.rs`) that proves the cache-stability/system-invariance property; this PR implements the same idea without that formal anchor. For a provider-input change that's the line the project holds.

**One substantive thing to double-check in your approach** (and a reason the placement matters): you render `request_context_template` as a *system reminder* per request. If that lands in the system-role region, it reopens the exact prefix-cache break #497 exists to prevent — the per-request dynamic block needs to be outside the cacheable system prefix. #506 renders it as a turn-1 `<context>` **user** message for that reason. Worth confirming on your side regardless.

**What's worth carrying over from here:** your two-field split naming — `system_context_template` (run-once, stable) vs `request_context_template` (per-request) — reads cleaner than a single field + catalog. I've asked the #506 session to look at adopting that naming. The migration/desired-state wiring you did is also a good reference.

Mind giving #506 a review with that lens? Proposing we close this one as superseded once #506 is in — but your call, and happy to discuss if you see something #506 is missing.

iverc:
Closing as superseded by #506, which landed the prompt templating implementation with the Lean PromptAssembly model, conformance fence, and cache-safe request context rendering.

PR #505, opened 2026-06-15
Remove CodexThreadProjection from Codex shim

### LINKED ISSUE #494: Refactor: CodexThreadProjection should derive from loaded tools/runtime state, not its own collection

jackzampolin:
## Summary
- remove the `CodexThreadProjection` collection/schema/registry entries and derive Codex thread records from scoped `AgentSession` + `AgentConversation`
- move shim-only thread state into the server-side sidecar and derive cwd/git metadata/token usage from runtime data
- migrate shim tests to durable runtime collections and cover the create-before-first-turn name path, git derivation, and token usage notifications

Closes #494.

## Tests
- `cargo build -p defra-agent`
- `cargo build -p defra-agent-cli`
- `cargo test -p defra-agent`
- `cargo test -p defra-agent-cli --test cli_codex_shim codex_shim_protocol_turn_streams_defra_response`
- `rg "CodexThreadProjection|CODEX_THREAD_PROJECTION" crates -n` (clean)


iverc:
Took a review pass focused on the overlap with #500 and the cwd carry-over concern.

I pushed `6d090ab3` to keep the useful cwd recovery path without bringing back `CodexThreadProjection`:

- prefer the server-side sidecar cwd override when present
- fall back to latest scoped `AgentRequest.metadata.codex_shim.cwd`
- then fall back to thread settings cwd / shim default
- warm the sidecar during thread record assembly so subsequent turns and git derivation use the recovered cwd

I did not add a legacy `CodexThreadProjection` fallback because this PR's design doc explicitly accepts retiring pre-upgrade projection rows, and keeping that collection load path would undercut the deletion.

Local checks:

- `cargo test -p defra-agent-cli commands::codex_shim::thread_projection::storage::tests -- --nocapture`
- `cargo check -p defra-agent-cli --tests`
- `cargo fmt --check`
- `git diff --check`

CI is rerunning on the pushed commit. Assuming it goes green, I agree this supersedes #500.

iverc (approved):
Approved after the cwd recovery follow-up and the rerun.

The final shape matches the direction I wanted from #500 while preserving this PR's fuller removal of `CodexThreadProjection`: cwd is derived from durable runtime data when possible, sidecar state stays process-local, and CI is green.

PR #506, opened 2026-06-15
Implement cache-safe prompt templating

### LINKED ISSUE #497: Prompt templating for dynamic node/agent context (cache-safe)

jackzampolin:
## Summary
- add the Lean PromptAssembly template model and Rust conformance mirror
- add the runtime-owned template catalog, parser-backed reads guard, and fail-closed system prompt rendering
- add request_context_template across schema/config/desired-state/CLI surfaces and render it as a persisted turn-1 <context> user message
- thread node.* and ctx.now into trigger, manual, and CLI task render paths

## Verification
- lake build Proofs.PromptAssembly.Template
- cargo test -p defra-agent
- cargo test -p defra-agent-cli
- git diff --check

Closes #497


iverc (approved):
Reviewed this against #497 and the overlap with #504.

This looks like the right landing path for prompt templating: it has the Lean `PromptAssembly.Template` model, Rust conformance coverage, parser-backed system-template read validation, and the per-request context is rendered as a user `<context>` message before the prompt instead of being inserted into the cacheable system prefix.

I also think the narrower schema choice is reasonable: keep the existing `system_prompt` as the stable/run-constant template surface, and add `request_context_template` for dynamic per-request context rather than introducing another system-context field.

No blocking issues from my pass.

PR #507, opened 2026-06-15
Add periodic recovery sweep registry

### LINKED ISSUE #477: Generalize periodic recovery sweeps behind a registry (Lean cadence dimension)
## Context

PR #473 added the first **periodic** recovery sweeps (#465): `RecoveryCadence.periodic` in the Lean sweep contract (`Proofs/Recovery/Contract.lean`), `Proofs/Recovery/Sweeps/SubagentLiveness.lean`, and `ToolCallLifecycle::reconcile_subagent_liveness` wired into `BackgroundCompletionObserver::run_reconcilers` (the 5s tick in `background_completion.rs`).

## Gap

The Lean side now has a cadence *dimension* (the registry distinguishes `startup` vs `periodic` per sweep), but the Rust side hardcodes the single periodic call inside the background-completion observer. The next periodic sweep means re-opening `run_reconcilers()` and re-plumbing by hand, and there is no mechanical check that every Lean sweep declared `periodic` actually has a live Rust tick call site (today that pairing is pinned only by the per-sweep-id cadence map in `tests/conformance/recovery_sweeps.rs`).

## Proposal (Lean-first)

1. Keep the Lean `registeredRecoverySweeps` registry as the source of truth; consider emitting a `(sweepId, cadence, rustFunction)` projection in the contract JSON (most of this already exists via `RecoverySweepCase`).
2. Add a Rust periodic-recovery registry — a list of `(sweep id, async fn)` entries the daemon tick iterates — so adding a sweep is one registry entry on each side.
3. Conformance: assert every Lean sweep with `cadence = periodic` appears in the Rust registry (and vice versa), replacing the hand-maintained cadence map.

Not urgent while there is exactly one periodic sweep; becomes worth it the moment a second one lands.

iverc:
## Summary

- add a Rust registry for periodic recovery sweep executors
- route the background completion observer through the registry instead of hardcoding subagent liveness recovery
- add conformance coverage that compares Lean `cadence = periodic` sweep metadata against the Rust registry, including sweep IDs and Rust function names

This keeps the current subagent liveness behavior unchanged while making future periodic recovery sweeps explicit and test-covered. The registry maps both Lean subagent liveness sweep IDs to the single Rust reconciler so the daemon does not run the same database pass twice per tick.

Fixes #477

## Testing

- `cargo check -p defra-agent --tests`
- `cargo test -p defra-agent --test conformance generated_recovery_sweep_cases_drive_startup_recovery_contract -- --nocapture`
- `cargo test -p defra-agent --test conformance subagent_liveness_reconciliation_converges_expired_processing_to_zero -- --nocapture`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/defra-agent/src/background_completion.rs
+++ b/crates/defra-agent/src/background_completion.rs
@@ -1105,17 +1105,20 @@ impl BackgroundCompletionObserver {
     }
 
     async fn run_reconcilers(&self) -> Result<()> {
-        // #465: terminalize expired subagent children (and project their
-        // background bridges), and interrupt queued descendants of terminal
-        // parents, on the live tick — startup recovery alone leaves a
-        // mid-run wedge in place until the next restart.
-        let liveness = crate::tool_call_lifecycle::ToolCallLifecycle::reconcile_subagent_liveness(
+        for run in crate::periodic_recovery::run_periodic_recovery_sweeps(
             self.node.as_ref(),
             &self.local_did,
         )
-        .await?;
-        if !liveness.is_noop() {
-            tracing::debug!(?liveness, "subagent liveness reconciliation applied");
+        .await?
+        {
+            if !run.is_noop() {
+                tracing::debug!(
+                    sweep_ids = ?run.metadata.sweep_ids,
+                    rust_function = run.metadata.rust_function,
+                    outcome = ?run.outcome,
+                    "periodic recovery sweep applied"
+                );
+            }
         }
         let unclaimed =
             reconcile_unclaimed_cross_deployment_spawns(self.node.clone(), &self.local_did).await?;
--- a/crates/defra-agent/src/lib.rs
+++ b/crates/defra-agent/src/lib.rs
@@ -49,6 +49,7 @@ pub mod meta_tools;
 pub mod migration;
 pub mod native_executor_status;
 pub mod oneshot;
+pub mod periodic_recovery;
 pub mod prompt;
 pub(crate) mod registry;
 pub mod retry;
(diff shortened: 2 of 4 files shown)


jackzampolin:
Clean, and it's properly Lean-first — verified the cross-check is genuine: `RecoveryCadence` is an independent dimension in `Contract.lean` and `SubagentLiveness.lean` declares `cadence := .periodic` on its own, so `assert_periodic_recovery_registry_matches_lean` is real Lean↔Rust cross-validation, not a restatement. Behavior is preserved (one executor → same reconciler, same noop-gated logging, no double DB pass), and the N:1 sweep-ids→reconciler mapping is modeled and dedup-guarded well. The registry earns itself at N=1 by replacing the hand-maintained cadence map with a mechanical check. Two nits, neither blocking:

1. `rust_function` is a hand-typed string on both the Lean and Rust sides, not bound to the actual symbol — a rename of `reconcile_subagent_liveness` would leave both stale-but-agreeing and the conformance check would still pass. A `stringify!`-anchored guard (or a comment flagging the hazard) would close that gap.
2. Minor surface: the conformance test only needs `periodic_recovery_sweep_metadata()`; `run_periodic_recovery_sweeps` + `PeriodicRecoverySweepRun`/`PeriodicRecoverySweepOutcome` look crate-internal and could be `pub(crate)`.

LGTM to merge once the green check is fresh.

jackzampolin (approved):
No blocking findings.

The registry preserves the existing periodic subagent-liveness behavior while moving the live tick call behind an explicit periodic recovery executor list. The Lean/Rust cross-check is meaningful: Lean owns `cadence = periodic` independently, and the Rust metadata has to match both the sweep IDs and function label, including the N:1 mapping from the two subagent-liveness predicates to the one reconciler pass.

Local checks I completed:

- `cargo check -p defra-agent --tests`
- `cargo fmt --check`

GitHub is also green for `rust-and-cli`, `lean-proofs`, and CodeRabbit. The earlier nits about the hand-typed `rust_function` label and exported surface are still non-blocking.

jackzampolin:
Follow-up on local verification: approval still stands, but I did hit a local hang while trying to run the focused conformance target:

```sh
cargo test -p defra-agent --test conformance generated_recovery_sweep_cases_drive_startup_recovery_contract -- --nocapture
```

It compiled successfully and started `running 1 test`, then printed:

```text
test generated_recovery_sweep_cases_drive_startup_recovery_contract has been running for over 60 seconds
```

After several more minutes it was still idle with no additional output, so I stopped the local run. GitHub CI is green for `rust-and-cli` and `lean-proofs`, so I did not treat this as a merge blocker in the review, but if this reproduces locally for others we should fix the hang in this PR rather than normalize it.

Issue #518, opened 2026-06-17
Stage live Playwright desktop E2E from browser to runtime


jackzampolin:
## Context

PR #516 adds a deterministic Chromium/Bombadil desktop UI harness backed by an in-memory `DesktopApiAdapter`. That is useful as a fast browser/UI regression gate, but it should not become a separate fake-backed E2E silo.

The existing `test:live*` suite already starts the real `defra-agent-desktop-tauri` `bridge_runner` through Cargo and drives the desktop runtime through the live bridge, but it renders through Vitest/Testing Library rather than a real browser. We need a staged path that combines those strengths: browser-level Playwright coverage all the way down to the live nodes/runtime.

## Goal

Add a live Playwright desktop E2E path that drives the real React shell in Chromium while using the existing live bridge/runtime infrastructure underneath. Keep the deterministic Playwright harness as the fast CI browser gate.

## Staged deliverable PRs

1. **Harness hygiene and shared contracts**
   - Make the Playwright `browserLogs` fixture automatic so console errors and page errors fail browser tests by default.
   - Tighten small invariant helpers where needed, for example counting actual primary-surface elements rather than selector categories.
   - Document the distinction between deterministic browser tests, live bridge tests, and the future live browser tests.

2. **Backend selection for the browser harness**
   - Refactor `tests/ui-harness/harness.tsx` so the browser harness can choose between deterministic and live backends.
   - Keep `harness.html?scenario=...` behavior unchanged for deterministic tests.
   - Add a narrow live-backend bootstrap contract, likely via a query mode such as `harness.html?backend=live`, env/config injection, or a small local test-control endpoint.

3. **Playwright live bridge fixture/global setup**
   - Reuse the existing `LiveBridgeRunner`/`bridge_runner` startup path instead of creating a second live runtime launcher.
   - Start the live bridge before a live Playwright project, expose the runner base URL/config to the browser harness, and tear it down reliably after the suite.
   - Preserve live diagnostics: stdout/stderr tails, request diagnostics, and useful artifact paths on failure.

4. **First live browser smoke flow**
   - Add a small Playwright live project/script that proves the full stack: real browser -> React shell -> desktop bridge -> runtime/node path.
   - Cover the minimum high-signal journey first: fleet visible, open chat, submit one request, observe completion in the browser, open operations/config enough to prove those surfaces still render against live data.
   - Run this in the existing live-smoke workflow rather than the fast deterministic `desktop-ui` CI job.

[...]

PR #520, opened 2026-06-17
Expand desktop UI coverage and stage live browser harness

iverc:
## Summary
- Expands deterministic desktop UI coverage across responsive layout, config, sidebar, operations, and sad-path flows.
- Makes Playwright browser console/page-error capture automatic.
- Adds a live-browser harness seam with `backend=live&bridgeUrl=...` so future Playwright tests can drive the real bridge/runtime path.
- Shares the bridge-runner HTTP adapter between existing live bridge tests and the browser harness.
- Fixes the live `bridge_runner` compile break from the updated DefraDB `add_replicator` signature.
- Documents the desktop test layers and ownership boundaries.

## Test Plan
- `npm run format:check`
- `npm run build`
- `npm run test:ui:unit`
- `npm run test:ui:e2e`
- `npm run test:ui:fuzz -- --time-limit 30s`
- `cargo check -p defra-agent-desktop-tauri --bin bridge_runner`
- `cargo fmt --all --check`

## Notes
This keeps live browser runtime coverage as a staged seam only. The normal PR-gating browser suite still runs against the deterministic adapter; the real live Playwright project can build on this without making fast UI CI depend on inference/runtime startup.


jackzampolin (changes requested):
Thanks for expanding the deterministic desktop UI coverage here. I think this is close, but I would not merge it yet in the context of #518.

One blocking harness-hygiene issue: `browserLogs` is still an opt-in Playwright fixture in `apps/desktop-tauri/tests/playwright/desktopTest.ts`, so the new specs that only request `{ page }` do not actually fail on browser `console.error` or `pageerror`. Since #518's first staged deliverable explicitly calls for making browser console/page-error capture automatic, this PR should switch that fixture to an auto fixture before we count it as part of that stage.

Suggested shape:

```ts
browserLogs: [
  async ({ page }, use, testInfo) => {
    ...
  },
  { auto: true },
],
```

I verified the PR locally and the deterministic suite is otherwise in good shape:

- `npm run format:check`
- `npm run build`
- `npm run test:ui:unit`
- `npm run test:ui:e2e`
- `npm run test:ui:fuzz -- --time-limit 20s`
- targeted `tsc` check for `tests/ui-harness/desktopHarness.ts`

So my only merge blocker is making that browser log guard automatic.

jackzampolin (approved):
Approving — my one blocker is resolved.

`browserLogs` is now an auto fixture (`{ auto: true }`) in `apps/desktop-tauri/tests/playwright/desktopTest.ts`, so specs that only request `{ page }` are still guarded: any `pageerror` or `console.error` fails the test via `expect(unexpected).toEqual([])`, with the full console log attached as an artifact on failure. That's exactly the #518 first-stage deliverable (automatic browser console/page-error capture). 👍

Thanks for the quick turnaround.

PR #522, opened 2026-06-17
Spike DefraDB CID time-travel reads for rendered-request projection

### LINKED ISSUE #519: Spike: DefraDB time-travel reads via node.execute passthrough (for rendered-request projection)
### Context

Coming out of the #503 reframing (rendered-request fidelity as a [provable projection](https://github.com/sourcenetwork/defra-agent/pull/503#issuecomment-4730934767) of durable state). To reconstruct a rendered provider request from durable rows, we sometimes need an input document's state **as of the request's commit** — e.g. behavior config (system prompt, sampling defaults) that may have changed between the inference call and the export.

DefraDB is MerkleCRDT-backed: every document has a full commit DAG, and DefraDB natively supports time-travel query syntax (`cid:` / `_version`). This repo, however, only ever calls `node.execute(graphql_string)` with current-state queries — it never passes a temporal query. Since `execute` takes an arbitrary GraphQL string, time-travel reads **may already work today** with no new `defra_node` API; we just don't use them.

### The spike

Confirm whether a time-travel query passed through the pinned `defra_node` passthrough resolves:

1. Write a doc, mutate it (≥2 commits), then query an earlier state via DefraDB `cid:` / `_version` syntax through `node.execute()`. Does it return the historical state, or error/ignore the temporal arg?
2. If it works: confirm we can resolve "behavior doc state as of commit X" given a request's creation/claim point — i.e. map a request to the relevant commit and read config-as-of-then.
3. If it doesn't: note what's missing in the passthrough and whether it's an upstream `defra_node` (defradb.rs) ask.

### Outcome

- **If yes** → the rendered-request projection needs to stamp almost nothing extra; config reconstruction is free via time-travel, and we only durably capture the genuine leaks (provider wire-shape, MCP-resolved tool schemas — see #503 comment).
- **If no** → the projection relies on per-call stamps on `InferenceCall` (resolved scalars frozen at call time), which is strictly sufficient on its own. Time-travel becomes a nice-to-have / upstream item rather than a dependency.

Either way this is **not a blocker** for the #503 projection direction — it only changes how much we stamp vs. reconstruct.

### Pointers

- `node.execute()` usage: see query construction in `crates/defra-agent/src/trace.rs` and `crates/defra-agent/src/run_timeline.rs`
- defradb.rs pin: workspace `Cargo.toml`
- Determinism boundary + leak set: #503 PR comment

iverc:
## Summary

Closes #519.

Adds focused e2e coverage proving that the pinned DefraDB `EmbeddedNode::execute` path supports CID time-travel reads needed by rendered-request projection.

The test verifies that:

- `_version { cid ... }` exposes a composite commit CID for a freshly written document
- `_commits` returns the same composite commit CID
- after the document is updated, a normal query sees the latest value
- querying `Collection(cid: ["<historical-cid>"])` reconstructs the earlier value

## Why

The rendered-request training path should be a projection from durable database state, not a hot-path capture of in-memory provider requests. This spike confirms that projection code can use DefraDB commit CIDs through the same `node.execute` GraphQL surface used by the runtime.

This does not implement the full rendered-request projection yet; it confirms the key database primitive and gives us a regression test for it.

## Test Plan

- `cargo fmt --check`
- `cargo test -p defra-agent --test e2e_runtime node_execute_supports_cid_time_travel_projection_reads -- --nocapture`

### DIFF
--- a/crates/defra-agent/tests/e2e_runtime.rs
+++ b/crates/defra-agent/tests/e2e_runtime.rs
@@ -9,6 +9,8 @@ mod support;
 mod adapter_projection_external_fixtures;
 #[path = "e2e_runtime/agent_behavior_migration.rs"]
 mod agent_behavior_migration;
+#[path = "e2e_runtime/defradb_time_travel.rs"]
+mod defradb_time_travel;
 #[path = "e2e_runtime/document_config_bootstrap.rs"]
 mod document_config_bootstrap;
 #[path = "e2e_runtime/event_source_subscription_factory_smoke.rs"]
--- a/crates/defra-agent/tests/e2e_runtime/defradb_time_travel.rs
+++ b/crates/defra-agent/tests/e2e_runtime/defradb_time_travel.rs
@@ -1,0 +1,206 @@
+//! Spike/regression coverage for rendered-request projection (#519).
+//!
+//! Rendered completion requests should be reconstructable as projections from
+//! durable database state. This test confirms that the pinned DefraDB node
+//! accepts CID time-travel GraphQL through `EmbeddedNode::execute`, which lets a
+//! future projection read behavior/config documents as they existed when the
+//! model call was made.
+
+use std::sync::Arc;
+
+use defra_agent::defra_node::EmbeddedNode;
+use defra_agent::graphql::escape_graphql_string;
+use serde_json::Value;
+use tempfile::TempDir;
+
+const PROBE_SCHEMA: &str = r#"
+type RenderedTimeTravelProbe {
+    probe_id: String @index(unique: true)
+    label: String
+}
+"#;
+
+struct ProbeDb {
+    node: Arc<EmbeddedNode>,
+    _tempdir: TempDir,
+}
+
+async fn probe_db() -> ProbeDb {
+    let tempdir = tempfile::Builder::new()
+        .prefix("defra-agent-time-travel-probe-")
+        .tempdir()
+        .expect("tempdir");
+    let node = Arc::new(
+        EmbeddedNode::builder()
+            .data_path(tempdir.path())
+            .build()
+            .await
+            .expect("embedded node"),
+    );
+    node.add_schema(PROBE_SCHEMA)
+        .await
+        .expect("add probe schema");
+    ProbeDb {
+        node,
+        _tempdir: tempdir,
+    }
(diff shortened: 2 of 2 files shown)


jackzampolin:
Reviewed + ran it locally on `0c0131fe`: compiles, and `node_execute_supports_cid_time_travel_projection_reads` passes (1 passed; 0.07s). The primitive is real — `Collection(cid: [...])`, `_version { cid }`, and `_commits` all resolve through the same `node.execute` surface the runtime already uses, no new `defra_node` API and no upstream ask. 👍 to merge **as a spike** — it gives us a regression fence on the database floor that the actual feature depends on.

Worth being explicit about what this is and isn't, so it doesn't get mistaken for the feature:

**This proves step 1 of #519 (raw time-travel read works). It does not do step 2 (map a request → the commit that was live at call time), and it doesn't build the projection.** The behavior we actually want — materialize a past request and have it *automatically* pull the right historical versions of each contributing document and reassemble the rendered wire request, behind the scenes — is the materialized-view feature, and that's still ahead of us.

I've opened **#523** to track that feature. The remaining work:

1. **Request→commit resolution (the crux).** Decide (A) stamp the source-doc composite CIDs onto `InferenceCall` at render time — exact/deterministic — vs (B) time-correlate via the already-durable `ctx.now` against each doc's `_commits` DAG — no new writes but needs a defined at-or-before lookup the spike never exercised. #519 leaned (B)-as-free; the lookup itself is unproven, and (A) is strictly sufficient alone.
2. **Durably capture the leak set** — provider wire-shape + MCP-resolved tool schemas (everything else is reconstructable; `ctx.now` already durable). Per the #503 comment.
3. **The projection function** — reassemble from time-travel reads + the leak set.
4. **Materialize-on-read wiring** through `run_timeline.rs` / `adapter_projection.rs` / CLI `trace`, so it's behind-the-scenes, not a separate call.
5. **Fidelity as a provable property** — PromptAssembly model fences `projection(request) ≡ rendered-at-call-time`, conformance test beyond the spike. This is THE crux of #503.

Per the foundation flow, #523 starts in Lean (the resolution strategy + the determinism property), then drives conformance, then implements — this spike is the floor it stands on. Suggest merging this and moving the design conversation to #523.

PR #525, opened 2026-06-18
Gate embedded Codex TUI behind a CLI feature

iverc:
## Summary
- Add a default-on `codex-tui` feature to `defra-agent-cli`
- Make the embedded `defra-agent codex` TUI dependencies optional
- Document the lean CLI build path with `--no-default-features`

## Why
The embedded Codex TUI is useful, but it pulls in a large terminal UI dependency tree. Headless/server-style CLI builds do not need that weight.

Default builds keep existing behavior. Lean builds can now skip `codex-tui`, `ratatui`, and `crossterm` while keeping the server shim and `chat` command available.

## Test Plan
- `cargo check -p defra-agent-cli --no-default-features`
- `cargo test -p defra-agent-cli commands::codex`
- `cargo fmt --all --check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,119 @@
+SHELL := /bin/sh
+
+CARGO ?= cargo
+LAKE ?= lake
+NPM ?= npm
+
+DESKTOP_DIR := apps/desktop-tauri
+PROOFS_DIR := crates/defra-agent/proofs
+FUZZ_TIME ?= 30s
+
+.DEFAULT_GOAL := help
+
+.PHONY: help
+help:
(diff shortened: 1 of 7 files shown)


jackzampolin:
Okay, I love this. I feel like we're about to develop a huge number of different build types, and also we have a number of tests and different ways to run the tests, especially with the desktop end-to-end tests. And then the agents have an end-to-end test. Can I make a brief request on this PR that you create a Makefile that starts a way of organizing these different builds and test things? I think it'll be good to have a central way to run those.

PR #530, opened 2026-06-19
Store ChatGPT OAuth credentials in DefraDB

jackzampolin:
## Summary
- add an `OAuthCredential` schema/row and store ChatGPT OAuth tokens as DefraDB documents scoped by `agent_did` + provider
- add `defra-agent codex-login` for ephemeral Codex OAuth login, DB upsert, and redacted output
- rework ChatGptCodex runtime auth to read/refresh the credential document and inject a fresh bearer per request
- update `codex-auth-probe`, `diagnose`, model discovery, and docs for DB-backed credentials instead of `~/.codex` / `DEFRA_CODEX_HOME`
- block `defra_query` from selecting or filtering on OAuth token fields

## Verification
- `cargo test -p defra-agent`
- `cargo clippy -p defra-agent -p defra-agent-cli --all-targets` (passes with existing repo warnings)


crates/defra-agent/src/chatgpt_oauth_refresh.rs
```diff
@@ -0,0 +1,200 @@
...
+        .await
+        .map_err(|error| {
+            ChatGptAuthProblem::Other(format!("ChatGPT token refresh request failed: {error}"))
+        })?;
+
+    let status = response.status();
+    if !status.is_success() {
+        let body = response.text().await.unwrap_or_default();
+        if status == reqwest::StatusCode::UNAUTHORIZED {
+            return Err(ChatGptAuthProblem::Expired);
```
jackzampolin, line 73: **Lossy refresh error-mapping (medium).** All 401s collapse to `Expired`, discarding the `refresh_token_reused` / `refresh_token_invalidated` distinction (and with no `tracing` of status+body — the canonical codex impl logs both). Also confirm transient (network/5xx) failures don't map to a *terminal* auth error: a transient blip should fall back to the still-valid cached token, not surface `re-run codex-login`. Preserve the reuse/invalidated signal and log it (codes/status only, never token material).

crates/defra-agent/src/chatgpt_oauth_refresh.rs
```diff
@@ -0,0 +1,200 @@
...
+            .unwrap_or_else(|| refresh_token.to_string()),
+        id_token: refreshed.id_token,
+        account_id: id_claims.account_id,
+        is_fedramp: id_claims.is_fedramp,
+        plan_type: id_claims.plan_type,
+        access_token_expires_at,
+    })
+}
+
+pub fn decode_id_token_claims(id_token: &str) -> IdTokenClaims {
```
jackzampolin, line 116: **Untested + verify two correctness points (medium/low).** `decode_id_token_claims`/`jwt_payload` are only exercised on the happy nested-claims path — add malformed/missing-segment cases and assert no panic on attacker-influenced input (base64**url**, padding). Separately: if the refresh response omits `refresh_token`, the code must **preserve the prior** refresh token (codex's `persist_tokens` does) — dropping it bricks the credential.

crates/defra-agent-cli/src/commands/codex_login.rs
```diff
@@ -0,0 +1,102 @@
...
+pub(crate) async fn codex_login(args: CodexLoginArgs) -> Result<()> {
+    let (access, home_dir) =
+        resolve_config_access(args.home.as_deref(), args.graphql.as_deref(), true).await?;
+    let agent_did = resolve_agent_did(Some(&home_dir), args.agent_did.as_deref())?;
+    let provider = normalize_provider(&args.provider);
+    let synthetic_home =
+        std::env::temp_dir().join(format!("defra-agent-codex-login-{}", Uuid::new_v4()));
+    let mut opts = ServerOptions::new(
+        synthetic_home.clone(),
+        args.client_id.unwrap_or_else(|| CLIENT_ID.to_string()),
```
jackzampolin, line 21: **`--client-id` override is write-only (low).** A non-default client_id chosen at login is stored implicitly but the runtime refresh path (`chatgpt_oauth_refresh`) always uses `codex_login::CLIENT_ID`. A mismatch makes every refresh fail. Either drop the override, or persist the client_id on the credential and thread it through refresh.

crates/defra-agent-cli/src/commands/codex_login.rs
```diff
@@ -0,0 +1,102 @@
...
+    let token_data = auth
+        .get_token_data()
+        .context("ChatGPT login did not expose token data")?;
+    let credential = defra_agent::chatgpt_codex::OAuthCredential::from_login_token_data(
+        &agent_did,
+        &provider,
+        &token_data,
+        chrono::Utc::now(),
+    );
+    let mutation = defra_agent::chatgpt_codex::oauth_credential_upsert_mutation(&credential);
```
jackzampolin, line 74: **Bare auto-committed mutation (low).** Uses `access.execute(&mutation)` rather than the committed path — no DefraDB Update event is emitted on commit (the #228/#970 stopgap pattern). If a watcher needs to observe the new credential, use the execute-committed path other config writes use.

crates/defra-agent-cli/src/commands/config/backend.rs
```diff
@@ -61,6 +61,7 @@ pub(super) async fn backend_discover_models(args: BackendDiscoverModelsArgs) ->
         target.provider_kind,
         &target.endpoint,
         target.api_key.as_deref(),
+        None,
```
jackzampolin: **`backend discover-models` is broken for ChatGptCodex (medium).** This always passes `None` as the credential, so `discover_models` bails for a ChatGptCodex backend. Fix needs `agent_did` + `load_oauth_credential(...)` resolved here (the resolver already exists in `codex_auth_probe.rs`) and passed in — an args-source decision, so I left it for a focused follow-up rather than half-wiring it.

crates/defra-agent-protocol/src/row.rs
```diff
@@ -663,6 +663,35 @@ pub struct InferenceBackendRow {
     pub probe_status: Option<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct OAuthCredentialRow {
```
jackzampolin, line 667: **`Debug` over plaintext token fields (nit, rising stakes).** `OAuthCredentialRow` derives `Debug` while holding `access_token`/`refresh_token`/`id_token`. Harmless today (verified never printed), but once encryption lands a stray `{:?}` would leak plaintext at the decryption boundary — prefer a manual `Debug` that redacts the token fields now.

crates/defra-agent-cli/tests/cli_diagnose.rs
```diff
@@ -81,6 +81,10 @@ async fn diagnose_works_from_local_home_without_server() -> Result<()> {
             .and_then(Value::as_bool),
         Some(true)
     );
+    assert!(
+        output.pointer("/checks/chatgpt_auth/ok").is_some(),
```
jackzampolin, line 85: **Assertion too weak (medium).** `is_some()` passes for `ok:true`, `ok:false`, or a malformed object. Assert `ok` is a bool, and that `guidance` is present when `ok:false` — and cover both the doc-present (`ok:true`) and doc-absent cases.

crates/defra-agent/src/defra_query/mod.rs
```diff
@@ -259,7 +259,7 @@ mod truncation_tests {
         let mut value = json!({
             "count": 42,
             "active": true,
-            "score": 3.14,
+            "score": 2.5,
```
jackzampolin, line 262: **Unrelated edit bundled (nit).** The `3.14 → 2.5` fixture change is unrelated to OAuth-in-DB (flagged on the prior revision too). Split it into its own commit/PR.

jackzampolin (commented):
## Ultracode review follow-up — 3 blockers fixed, remaining findings inline

A multi-agent adversarial review (7 dimensions, findings verified by a refute pass) ran against the prior HEAD. **The dedicated secret-leak hunt came back clean**: the `defra_query` guard holds, the token-bearing `Debug` derives are never printed, and the credential is in zero replication templates so nothing broadcasts it. Plaintext-at-rest is accepted (parity with `~/.codex`; encryption is the named next slice).

### Fixed in `ba6959d1` (the 3 merge blockers)
1. **`agent_did @immutable`** — defradb hard-rejects a non-immutable replication-filter field, so without this the v1 filtered-replication scoping (the only secret-containment control) silently fails closed.
2. **`codex-auth-probe` is now read-only** — removed the second, uncoordinated refresh writer that could race the runtime and trigger provider refresh-token **reuse-detection → credential revocation**.
3. **Write-back can no longer strand the rotated token** — the rotated credential becomes the in-memory source of truth *before* persisting; persist is retried with backoff; ultimate write failure serves-from-memory + logs rather than re-refreshing with the already-consumed token.

Gate after fix: `cargo fmt`/`clippy` clean, `cargo test -p defra-agent` and `-p defra-agent-cli` green.

### Remaining (non-blocking) — see inline comments, plus these structural ones:
- **`is_owner` is hardcoded `true`** (chatgpt_codex.rs:965). Correct for the single-node demo, but it's the Sharp-Edge-A gate: multi-node/fleet serving must not ship until `is_owner` is derived from a real ownership source (the deployment for `(agent_did, behavior)`) **and** single-writer refresh is enforced cross-process — otherwise every replica refreshes the same token → reuse-revocation.
- **Replication scoping is not actually wired.** `OAuthCredential` is in no `p2p_reconcile/templates.rs` collection set, so the spec's "replicates only to `agent_did`" is currently unimplemented (safe today — it just doesn't replicate, and the non-owner branch can't receive rotations). When it's wired, it must use a **PeerDid-scoped** template (never `Unscoped`/`Replicate` like `InferenceBackend`), plus a test asserting `OAuthCredential` is never placed in an unscoped template — to prevent an InferenceBackend-by-analogy open-broadcast.
- **Missing test suite** (spec-mandated): `DbCredentialBearer` (fresh / near-expiry-refresh+write-back / non-owner-never-refreshes / 401→actionable / re-read-after-lock), `refresh_chatgpt_token` request-shaping + error-mapping, `from_login_token_data`, and malformed-JWT decode. The encryption fast-follow will rewrite these exact paths, so landing the harness now protects the refresh-token-reuse property through that rewrite.

I held off on the `backend discover-models` ChatGptCodex fix — it needs `agent_did` + credential-doc resolution plumbed into the discovery command (an args-source decision), more than a one-liner, and it currently degrades to an actionable error, not a silent break. Flagged inline.

iverc (approved):

Issue #576, opened 2026-06-29
`defra-agent demo` command — make the two-node fleet demo runnable from a release


jackzampolin:
## Summary

Promote the two-node fleet demo from a Makefile + shell-script prototype into a first-class **`defra-agent demo`** subcommand that a user can run **straight from a downloaded release binary** — no repo checkout, no `make`, no `bash`.

## Background

The demo currently lives as Makefile targets driving shell scripts:

- **#569** — `make demo-p2p-two-node` / `scripts/demo-p2p-two-node.sh`: brings up two isolated runtimes, joins them through the signed `network-control` invite flow, installs the conversation data plane, and proves request + conversation replication.
- **#570** — `make demo-desktop-two-node` / `scripts/demo-desktop-two-node.sh`: layers the Tauri fleet UI on top, seeds an isolated desktop home, tightens the tool surface, and wires a subagent.

These scripts are a faithful prototype of the orchestration the command should own, but they assume a source checkout, GNU make, bash/jq/curl, and a dev desktop toolchain. The end goal is: **download the release for your platform → `defra-agent demo` → watch a two-node agent fleet delegate work over the replicated control plane.**

## What the command should own

Port the orchestration the shell scripts do today:

1. Spin up two isolated nodes (Orchestrator + Observer) with isolated homes/ports.
2. Create the signed network, grant membership, join via a `network-control` invite.
3. Install the bidirectional conversation data plane and wait for reconciled replicators.
4. Configure the tool surfaces (drop `defra_query`; enable cross-node delegation).
5. Launch the desktop fleet UI pointed at both runtimes; teardown on exit.

## Decisions baked in (from demo-direction review)

- **Cross-node delegation, blessed for the demo.** Orchestrator delegates to a `worker` on the **Observer** node (not same-node self-spawn). This means flipping `subagent_allow_cross_deployment=true` on both selections — acceptable for local/trusted-fleet loopback ahead of ACP. (Mechanism verified: the gate at `background_tools.rs` rejects remote spawns only when the flag is off; `cli_fleet_delegation_live.rs` proves cross-node works with it on.)
- **No mock inference backend.** The normal path is user-provided real inference (URL/preset/model passthrough); ship a couple of suggested prompts crafted to reliably elicit `spawn_subagent`. (Removing the mock means delegation firing is model-dependent, so prompt quality matters; both nodes must reach the backend since the Observer now executes.)

## Transition plan

[...]

PR #585, opened 2026-06-30
feat(demo): pair + delegate — 2-node cross-node delegation (stage 2)

iverc:
Stage 2 of the `defra-agent demo` epic (#576), stacked on #584.

- `pair`: brings up node B (worker), enrolls it via a signed network-control invite, installs the conversation data plane both ways, waits for replicators.
- `delegate`: enables cross-node subagent delegation (background await + the allow-cross-deployment flag on both selections; orchestrator targets the worker). Asking the orchestrator to delegate runs the child on node B and replicates back.
- Shell tracks fleet state (node A + optional node B); `status` shows both; teardown stops both nodes. Backend is persisted so node B reuses it.

Verified: pair (2-node replication) + delegate (child runs on the worker node) via a test-only tool-calling backend.

### DIFF
--- a/crates/defra-agent-cli/src/commands/demo/mod.rs
+++ b/crates/defra-agent-cli/src/commands/demo/mod.rs
@@ -14,8 +14,11 @@ use serde_json::Value;
 use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::process::{Child, Command};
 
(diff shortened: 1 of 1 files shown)


jackzampolin:
Superseded by #586, which is now retargeted to main and contains the combined pair/delegate + desktop demo stack for review/acceptance.

PR #594, opened 2026-07-01
desktop: fix #583, gate the tauri crate in CI, and harden the UI test harness

### LINKED ISSUE #583: Desktop transcript test doesn't support Lean case parallel_results_share_assistant_turn
## Symptom

`cargo test -p defra-agent-desktop-tauri` →
`session_snapshot_transcript_rendering_consumes_generated_transcript_cases` fails:
- `assert_eq!(cases.len(), 6)` is now `left: 7, right: 6` — Lean emits **7** transcript cases.
- Bumping the count to 7 then panics in the loop: `unsupported Lean transcript case "parallel_results_share_assistant_turn"` (the test helper `transcript_contract_store` / the per-case match at `apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_state.rs:~1213`).

## Cause

A 7th Lean transcript conformance case (`parallel_results_share_assistant_turn` — parallel tool results sharing an assistant turn) was added on `main`, but the **desktop** transcript structure-fence test wasn't taught to build/render it. Pre-existing on `main`; not caught by CI (the tauri crate isn't `cargo test`-ed in CI).

## Fix

In the desktop transcript test harness:
- [ ] Add a `transcript_contract_store` arm for `parallel_results_share_assistant_turn` (build the store for parallel tool results sharing one assistant turn).
- [ ] Bump the `cases.len()` count guard 6 → 7.
- [ ] Confirm the per-case assertions (message/tool-call counts, hidden tool-result rows, rendered tool groups) hold for the new case — i.e. the desktop renderer groups parallel tool results under the shared assistant turn as the Lean case expects.

Owner: whoever added the `parallel_results_share_assistant_turn` Lean transcript case (transcript/streaming + workflow orchestration). Surfaced while repairing the `AgentToolCallRow` tauri-test compile break (#582), which had been masking this runtime failure.

## References
- #582 (the compile fix that unmasks this)
- `apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_state.rs:492` (the failing test)

iverc:
Desktop-perfection epic, stages 1–2. Closes #583.

## Stage 1 — correctness + close the CI hole
- **Fix #583** — the desktop transcript snapshot test consumed only 6 of the 7 Lean transcript conformance cases; the 7th (`parallel_results_share_assistant_turn` — three parallel tool calls sharing one assistant turn) was missing, so `cargo test -p defra-agent-desktop-tauri` was red. Wired the case into the test fixtures (the renderer already groups parallel tool calls, so this is a test-only fix).
- **Close the CI hole** — the tauri crate was never `cargo test`'d in CI, which is exactly why #583 shipped red. Added `cargo test -p defra-agent-desktop-tauri --lib` to the `rust-and-cli` job (PR smoke + main).
- **Guard the fleet row buttons** — Playwright spec clicks the three per-row action buttons directly (chat/config navigate; repair disabled while healthy). Previously only the agent name was exercised.
- **Cleanup** — removed a never-read field the new gate surfaced.

## Stage 2 — harden the UI test harness
- **Scenario coverage** (`desktop-async-states.spec.ts`) for the three previously-undriven harness scenarios: `loading` (delayed-load convergence, no error), `save-error` (banner + no Saved chip + button recovery + failure scoped to behavior saves), `backend-health-error` (Backends-tab failure is isolated; other ops tabs still work).
- **Responsive reachability** (`desktop-layout.spec.ts`) — locks in that the fleet row action buttons stay reachable at 390px (via the scrollable `.fleet-table-wrap`, without widening the page). A future `overflow: hidden`/clip regression flips it red.
- **Component units** — `FleetRow` (three action buttons, DID-passing handlers, repair enable/disable logic) and `AddPeerForm` (manual submit vs `/status` fetch→parse→submit discovery).

Verified: tauri 109 lib tests; desktop **136 unit / 111 e2e** (+21); fmt/prettier clean; non-flaky repeat. Two adversarial review passes — the 2 confirmed test-quality findings were fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,10 +103,15 @@ jobs:
           cargo test -p defra-agent-cli --test cli_server server_startup_with_iroh_p2p_reports_runtime_connectivity -- --nocapture --test-threads=1
           cargo test -p defra-agent-cli --test cli_status -- --nocapture --test-threads=1
           cargo test -p defra-agent-cli --test cli_reconciliation reconciled_runtime_sends_generation_two_tools_and_completes_tool_loop -- --nocapture --test-threads=1
+          cargo test -p defra-agent-desktop-tauri --lib
 
       - name: Run full library and integration tests
         if: github.event_name != 'pull_request'
         run: cargo test -p defra-agent --lib --tests
+
+      - name: Run desktop Tauri crate tests
+        if: github.event_name != 'pull_request'
+        run: cargo test -p defra-agent-desktop-tauri --lib
 
       - name: Run full CLI test suite
         if: github.event_name != 'pull_request'
--- a/apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_state.rs
+++ b/apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_state.rs
@@ -494,7 +494,7 @@ fn session_snapshot_transcript_rendering_consumes_generated_transcript_cases() {
     let cases = lean_transcript_cases();
     assert_eq!(
         cases.len(),
-        6,
+        7,
         "desktop transcript rendering should consume every generated Lean transcript case"
     );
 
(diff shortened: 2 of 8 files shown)


jackzampolin (approved):
Reviewed locally. No blocking findings from the diff; targeted desktop unit/e2e checks passed. The only remaining item I saw is to let the pending rust-and-cli CI job finish green before merge.

Issue #580, opened 2026-06-30
Desktop tool-configuration redesign (unified tool-policy)


jackzampolin:
## Context

PR #568 landed the unified tool-policy model (SP1–SP3 + Phase B + bash). On the desktop, **Phase B fixed the backend save/load data-loss** — the desktop-core mutation builder + read queries + the tauri bridge (request/view/snapshot) now carry every tool-selection field (`tool_policy_version`, `defra_query_collections`, `write_tools`, `subagent_default_await_mode`, `orchestration_enabled`), and the TS types were added — so a configured query allowlist no longer silently reverts on save.

**Deliberately deferred:** the editable **panel UI** itself. The data is plumbed end-to-end and round-trip-tested in Rust, but `ToolSelectionConfigPanel.tsx` does not yet expose controls to *edit* these fields, and there is no surfacing of the policy version or the `ToolSurfaceExplanation` trace. It was deferred because verifying frontend changes needs `npm ci` + the visual/Playwright suite.

## Why a redesign (not just a control)

The current desktop tool panel predates the unified policy. The model is now category-complete (`Capability` / `BashPolicy` / `EndpointScope` over ~11 categories, behavior ⊓ operator-ceiling, with a `requested → ceiling → runtime → effective` explanation). The panel should be redesigned around that model rather than bolting on individual fields.

## Scope

Immediate (the deferred slice):
- [ ] Editable `defra_query_collections` allowlist control in `ToolSelectionConfigPanel.tsx` (the field whose revert was the data-loss bug). Honor preserve-on-absent (`Option<Vec>` on the bridge).
- [ ] Read-only **policy version** facts row (`toolPolicyVersion ?? "legacy (unversioned)"`).
- [ ] Keep `write_tools` **preserve-only / non-editable** in the UI — editing raw `WriteToolDecl` JSON risks bricking the fail-closed runtime loader (see PR #568 review). Display only.

Redesign (the larger effort):
- [ ] Surface the full unified-policy surface per category (file/bash/cli, mcp/meta, defra_query + collections, subagents + targets + await-mode + cross-deployment, background, memory, session-history, context-budget, orchestration, skills).
[...]

PR #595, opened 2026-07-01
feat(desktop): editable defra_query allowlist + tool-policy facts (#580)

iverc:
Immediate slice of #580 (desktop tool-configuration for the unified tool-policy). The #568 model, bridge, and TS types already carry these fields end-to-end; this exposes the UI. Frontend-only — no type or Rust-bridge changes.

- **Editable `defra_query_collections` allowlist** in the Tool Selections panel (textarea, one collection per line; empty = all collections). Saved through the existing preserve-on-absent path — an empty list clears it, and the bridge emits `null`, never `[]`.
- **Read-only policy-version row** (`toolPolicyVersion ?? "legacy (unversioned)"`).
- **Display-only managed write-tools row** — decodes each JSON `WriteToolDecl` into a friendly `name → collection` rather than a raw JSON blob. `write_tools` is apply-managed and never sent from the UI.

Tests: unit (allowlist edit round-trips; save payload asserts `defraQueryCollections` and never `writeTools`/`toolPolicyVersion`; write-tools decoding; legacy-version fallback) + e2e round-trip (edit → save → reopen → persisted; display-only facts survive). The harness `saveToolSelectionConfig` now mirrors the bridge's preserve-on-absent.

**Deferred (larger scope of #580):** surfacing the `ToolSurfaceExplanation` trace needs a new Rust bridge command (none exists today) + a snake→camel view — a follow-up.

Verified: build (tsc) + 129 unit + 93 e2e green, non-flaky; adversarial review (its one confirmed finding — a raw-JSON write-tools display — is fixed by the decoder).

### DIFF
--- a/apps/desktop-tauri/src/components/config/ToolSelectionConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/ToolSelectionConfigPanel.tsx
@@ -148,6 +148,7 @@ export function ToolSelectionConfigEditor({
   const [subagentBackgroundEnabled, setSubagentBackgroundEnabled] = useState(false);
   const [crossDeploymentSpawnTimeoutSeconds, setCrossDeploymentSpawnTimeoutSeconds] =
     useState("");
+  const [defraQueryCollections, setDefraQueryCollections] = useState("");
   const ceiling = normalizeToolCeiling(toolCeiling);
   const fileToolsDisabledByCeiling = ceiling === "MetaOnly";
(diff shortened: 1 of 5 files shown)


jackzampolin (approved):
Reviewed the desktop save path and tests. The editable defra_query allowlist is scoped correctly, write_tools/tool_policy_version stay preserve-only, and the mutation path still serializes empty string lists as null rather than [].

Issue #575, opened 2026-06-29
Declarative P2P pairings do not support cross-deployment subagent topology


jackzampolin:
## Summary

The declarative P2P pairing templates available to fleet config do not currently appear to reproduce the topology used by the cross-deployment subagent live tests. Observed on v0.6.0 with Amy as coordinator and `workstation-1-host` as target.

## Evidence

- Amy and workstation-1 were directly P2P connected and had bidirectional `conversation` pairings.
- A background `spawn_subagent` to `workstation-1-host` created an Amy-side bridge row, but it expired as `no_peer_claimed_spawn`.
- Switching both pairings to `backup` made the rows less filtered, but caused target-side P2P authorization/direction errors including `access denied: peer ... not authorized for collection ...` and `Dropping GossipSub message outside accepted replication direction ...`. It also left stale foreign `AgentRequest` rows visible on the target.
- Reverting to `conversation` restored the safe filtered pairing state.
- The target runtime needed a service restart before pairing trust changes were reflected, which suggests `paired_peer_dids` is loaded into a runtime snapshot and not reconciled live.

## Expected behavior

Fleet config needs a durable declarative subagent delegation topology, or equivalent runtime reconciliation, that can:

- deliver coordinator-owned parent/bridge docs to the target,
- return target-owned child/result docs to the coordinator,
- avoid unfiltered `backup` replication,
- keep P2P authorization and replication direction consistent, and
- refresh paired-peer trust without a manual restart, or make that restart requirement explicit and observable.

## Reference

The live tests appear to create the working path with explicit one-way replicators in both directions before agent boot, plus pairing rows for trust. The fleet needs a persistent/declarative equivalent of that pattern before cross-deployment subagent delegation can be rolled out safely.

jackzampolin:
Closing as completed/superseded by #578. The broad gap in this issue -- declarative cross-deployment subagent topology -- now has dedicated subagent-coordinator/subagent-host scope templates and declarative PeerPairingDesired-driven wiring. Remaining desired-state-manifest ownership and pre-boot seeding work stays tracked in #607.

PR #597, opened 2026-07-01
fix(subagents): expose background bridge before child materialization (#593)

### LINKED ISSUE #593: Background subagent spawn can return a running child ID that is not visible to list/read
## Summary

A live Amy session on 2026-07-01 exposed a bad subagent control-plane contract: `spawn_subagent` can return a background receipt with `status: running` and a `child_request_id`, but the parent immediately cannot observe that child through `list_subagents` or by an exact `AgentRequest` lookup.

This is worse than a normal failed cross-deployment spawn because the parent is told it has a running child, then all follow-up control tools have no visible handle to read, wait, steer, or cancel.

## Observed evidence

Transcript-derived sample:

- Attempt A: background spawn reported child `bcbd01b6-d8f3-44ee-9c10-e9202a761b2f`; immediate `list_subagents(status="all", limit=20)` returned `entries: []`.
- Attempt B: spawn reported:
  - `ok: true`
  - `child_request_id: 353e2ea3-da41-4f58-94a9-e479f225ccac`
  - `await_mode: background`
  - `status: running`
- Immediately after Attempt B:
  - `list_subagents(status="all", limit=20)` returned `entries: []`.
  - exact DefraDB query `AgentRequest.request_id == 353e2ea3-da41-4f58-94a9-e479f225ccac` returned `count: 0`.

Quantification from this transcript:

- `2/2` spawned child IDs were invisible to `list_subagents(all)` immediately after the reported spawn.
- `1/1` exact `AgentRequest` lookup for a reported child ID returned no row.
- Impacted follow-up tools: `read_subagent`, `wait_subagent`, `steer_subagent`, and `cancel_subagent` have no reliable handle after the parent receives the spawn receipt.

## Relationship to existing issues

Related but not duplicate of #575. #575 tracks the declarative cross-deployment topology gap where a bridge can expire as `no_peer_claimed_spawn`.

This issue tracks the stricter parent-facing contract: after returning a successful background spawn receipt, the parent must be able to observe a durable control-plane state for that `child_request_id`, even if the child is still unmaterialized, not yet replicated, or eventually fails as unclaimed.

Also related to #564 because `SubagentSource` and bridge scans are involved, but the observed failure here is the user/operator surface: a successful spawn receipt followed by an empty child listing and missing exact request row.

## Code references

- `crates/defra-agent/src/hook/persistence/message_spawn.rs:678-720` generates the `child_request_id`, starts a subagent `AgentToolCall` lifecycle, and returns the background receipt before child materialization.
- `crates/defra-agent/src/background_tools.rs:375-502` implements `handle_list_subagents`.
- `crates/defra-agent/src/background_tools.rs:457-459` silently drops a bridge when the matching child `AgentRequest` is not present in `children_by_request`.
- `crates/defra-agent/src/background_completion.rs:122-199` can eventually fail unclaimed cross-deployment spawns, but until that projection happens the operator sees no subagent in `list_subagents`.
- `crates/defra-agent/tests/e2e_subagent/subagent_convergence.rs:200-277` asserts the local happy-path contract: bridge materializes child and `handle_list_subagents` reflects it. The live Amy case needs a cross-deployment/unmaterialized regression case.

## Expected behavior

One of these should be true after `spawn_subagent` returns a successful background receipt:

- the child `AgentRequest` is visible to the parent/listing path within a bounded interval, or
- `list_subagents(status="all")` shows a bridge-level entry for the returned `child_request_id` with a state like `awaiting_child_materialization`, `awaiting_peer_claim`, or `unclaimed`, or
- the spawn result is not reported as successful/running and instead returns a terminal failure payload.

Silent disappearance from the parent control plane should not be possible.

## Done when

- Add a regression test for a background spawn where the bridge exists but the child row is absent/not yet replicated; `list_subagents(all)` must return an observable entry or an explicit diagnostic.
- Add/extend a cross-deployment e2e covering the Amy-style path: receipt returned, parent lists immediately, then either observes materialization or observes an explicit unclaimed/failure state.
- Ensure `read_subagent`/`wait_subagent` errors for an unmaterialized returned child ID explain the bridge state rather than reporting only "not owned" or disappearing.

jackzampolin:
Closes #593.

## The invariant

**A returned background child id never disappears from the parent control plane.** After `spawn_subagent` returns `ok: true, status: running, child_request_id: X`, every parent-facing control tool must observe a durable state for `X` — even while the child `AgentRequest` is unmaterialized (spawn convergence #377 creates it asynchronously via `SubagentSource`; a cross-deployment child appears only after the owning peer claims and replicates it, or never, in which case the unclaimed reconciler fails the bridge as `no_peer_claimed_spawn`).

The live Amy session in #593 hit the gap: 2/2 spawned children invisible to `list_subagents(all)` immediately after a successful receipt, leaving read/wait/steer/cancel with no handle.

## Root cause

The bridge `AgentToolCall` is persisted synchronously at spawn (`message_spawn.rs`), but every read-side surface required the child row:

- `handle_list_subagents` silently dropped any bridge whose `child_request_id` had no matching `AgentRequest` (`background_tools.rs:457`).
- `read_subagent` collapsed the missing child into "not a background subagent owned by this parent request".
- `wait_subagent`/`cancel_subagent` collapsed it into a non-retryable "not available to this parent request".
- `steer_subagent` resolved it to `NotAuthorized`.

## Fix shape: a projection, not a new state

`awaiting_child_materialization` is a read-side **projection** of (bridge `await_mode = background` ∧ bridge non-terminal ∧ child row absent). It is never persisted and adds no transition to the Background bridge state machine — once the child materializes the projection collapses back to the bridge lifecycle state, so the local happy path is byte-identical (the optional `diagnostic` field is skipped when absent).

**Lean first** (`proofs` builds clean, zero `sorry`s): new R4c witness `r4c.list_subagents.unmaterialized_child_visible` (`ContractCases/Types.lean` + `Contracts/Json/BackgroundWork.lean`) pins the contract: listed under both `all` and the default `running` filter with the exact projected status string, read is never faked-terminal, wait is retryable. The conformance consumer (`tests/conformance/background.rs`, 6→7 cases) ties the witness string to the runtime constant.

**Runtime**:
- `list_subagents`: a background bridge without a child row now yields a bridge-level entry — status `awaiting_child_materialization` while the bridge runs (matched by `running` and `all` filters), or the bridge terminal state (e.g. `failed` after the unclaimed projection) under `terminal`/`all`; `name`/`behavior_id` come from the resolved bridge args, plus a `diagnostic` naming the bridge and the unclaimed deadline.
- `read_subagent`: returns the projected state (`terminal: false`, empty transcript, `diagnostic`) instead of not-authorized; a terminal bridge without a child reads as that terminal state — never a faked success.
- `wait_subagent`/`cancel_subagent`: when the caller owns a spawn bridge for the id, the error explains the bridge state and is `retryable` while the bridge is non-terminal.
- `steer_subagent`: new `AwaitingMaterialization` resolution → retryable `service_unavailable` explaining the bridge state.
- Ownership is enforced in the bridge probe (filtered by the caller's `request_id`), so lineage rejection (`r4c.list_subagents.lineage_rejects`) is unchanged — verified in the regression test.

## Tests

- **Red-first regression** (`subagent_convergence.rs::unmaterialized_background_child_stays_observable_in_list`): bridge targets a remote DID so the local `SubagentSource` never materializes the child — the durable analog of the Amy cross-deployment gap. Failed on the old code at exactly the reported symptom (`entries: []`); now asserts list (both filters), read, steer, and non-owner rejection. 3× stable.
- Unit tests pin the retryable-until-bridge-terminal explanation and the filter semantics of the projected status.
- Existing convergence happy paths (materialized child lists normally, foreground completion) unchanged and green.
- Gates: full `lake build` green; focused subagent e2e + conformance + unit suites green (3× stable); fmt clean; clippy introduces no new warnings. Full `cargo test -p defra-agent` is running locally at PR-open time — result will be confirmed in a comment.

## Notes

- A full 2-node cross-deployment e2e of this path belongs to the #575 declarative-pairing work; this PR pins the parent-facing contract with a deterministic single-node reproduction of the same gap.
- Model-facing tool descriptions for `list_subagents`/`read_subagent` document the new projected status so orchestrating models keep polling instead of concluding the child vanished.

### DIFF
--- a/crates/defra-agent/proofs/Proofs/Conformance/ContractCases/Types.lean
+++ b/crates/defra-agent/proofs/Proofs/Conformance/ContractCases/Types.lean
@@ -410,6 +410,40 @@ structure SteerInterruptComposes where
   queueInterruptedRequestId : String
   deriving Repr
 
+/-- #593: after a successful BACKGROUND spawn receipt the returned
+`child_request_id` must never disappear from the parent's control plane,
+even while the child `AgentRequest` is not yet materialized (spawn
+convergence #377 materializes it asynchronously via `SubagentSource`, and a
+cross-deployment child may replicate later or never be claimed).
+
+Boundary note: `awaiting_child_materialization` is a read-side PROJECTION of
+(bridge `await_mode = background` ∧ bridge non-terminal ∧ child row absent).
+It is never persisted as a bridge `lifecycle_state` and adds no transition
+to the Background bridge state machine (`Proofs/Background/Transition.lean`);
+once the child materializes, the projection collapses back to the bridge
+lifecycle state and the happy path is unchanged. -/
+structure UnmaterializedChildVisible where
+  callerRequestId : String
+  bridgeToolCallId : String
+  childRequestId : String
+  /-- The child `AgentRequest` row is absent in this scenario. -/
+  childMaterialized : Bool
+  /-- Persisted bridge state stays `running`; the projection never mutates it. -/
+  bridgeLifecycleState : String
+  /-- `list_subagents` entry status projected for the missing child. -/
+  listedStatus : String
+  /-- The handle must be visible under `status="all"`. -/
+  listedUnderAllFilter : Bool
+  /-- The projection is non-terminal, so the default `running` filter shows it. -/
+  listedUnderRunningFilter : Bool
+  /-- `read_subagent` reports the projection instead of not-authorized. -/
+  readLifecycleState : String
+  /-- Never fake a terminal outcome for an unmaterialized child. -/
+  readTerminal : Bool
+  /-- `wait_subagent` explains the bridge state with a retryable payload. -/
+  waitRetryable : Bool
+  deriving Repr
+
 end R4cWitnesses
 
 structure TranscriptCase where
--- a/crates/defra-agent/proofs/Proofs/Conformance/Contracts/Json/BackgroundWork.lean
+++ b/crates/defra-agent/proofs/Proofs/Conformance/Contracts/Json/BackgroundWork.lean
@@ -79,6 +79,27 @@ def r4cSteerAppendPreservesLineageJson
     ++ "\"queue_policy\":" ++ jsonString witness.queuePolicy
     ++ "}"
 
+def r4cUnmaterializedChildVisibleJson
+    (witness : R4cWitnesses.UnmaterializedChildVisible) : String :=
+  "{"
+    ++ "\"witness\":"
+      ++ jsonString "r4c.list_subagents.unmaterialized_child_visible" ++ ","
+    ++ "\"caller_request_id\":" ++ jsonString witness.callerRequestId ++ ","
+    ++ "\"bridge_tool_call_id\":" ++ jsonString witness.bridgeToolCallId ++ ","
+    ++ "\"child_request_id\":" ++ jsonString witness.childRequestId ++ ","
+    ++ "\"child_materialized\":" ++ boolString witness.childMaterialized ++ ","
+    ++ "\"bridge_lifecycle_state\":"
+      ++ jsonString witness.bridgeLifecycleState ++ ","
+    ++ "\"listed_status\":" ++ jsonString witness.listedStatus ++ ","
+    ++ "\"listed_under_all_filter\":"
+      ++ boolString witness.listedUnderAllFilter ++ ","
+    ++ "\"listed_under_running_filter\":"
+      ++ boolString witness.listedUnderRunningFilter ++ ","
+    ++ "\"read_lifecycle_state\":" ++ jsonString witness.readLifecycleState ++ ","
+    ++ "\"read_terminal\":" ++ boolString witness.readTerminal ++ ","
+    ++ "\"wait_retryable\":" ++ boolString witness.waitRetryable
+    ++ "}"
+
 def r4cSteerInterruptComposesJson
     (witness : R4cWitnesses.SteerInterruptComposes) : String :=
   "{"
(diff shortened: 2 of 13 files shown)


jackzampolin:
**Full-suite confirmation + review response.**

Full `cargo test -p defra-agent` against `304c1478`: **1,387 passed / 0 failed** across all 11 binaries (876 lib, 258 conformance incl. the new witness pin, 93 e2e_subagent, 41 lifecycle, 51 runtime, 14 triggers, 8 workflow, rest auth/property).

**Review finding (medium) — fixed in `2c33a5d3`:** `wait_subagent`/`cancel_subagent` applied the bridge-state fallback on *any* `load_authorized_child_edge` error whenever a bridge row existed, which could mask child-present corruption (e.g. missing `behavior_id`, linkage mismatch) or query failures as retryable materialization lag. They now gate on `authorization_lookup_error` exactly like read/steer — the fallback fires only for resolution failures (child absent/unlinked); everything else preserves the original non-retryable error.

**Test gap — closed in the same commit:**
- `wait_subagent_explains_unmaterialized_child_bridge` / `cancel_subagent_explains_unmaterialized_child_bridge`: hook-level, remote-DID bridge, assert `retryable: true` and that the message names the bridge.
- `wait_subagent_preserves_error_for_corrupt_materialized_child`: child row present with empty `behavior_id` → original `has no behavior_id` error preserved, `retryable: false`, explicitly asserted **not** masked as materialization lag.
- `authorization_lookup_error_excludes_child_present_corruption`: unit test pinning the classifier boundary.

Clean full-suite rerun against the amended tree is in flight; will confirm here.

jackzampolin:
Confirmed: clean full `cargo test -p defra-agent` against the amended tree (`2c33a5d3`) — **1,391 passed / 0 failed** (877 lib incl. the new classifier unit test, 258 conformance, 96 e2e_subagent incl. the 3 new wait/cancel boundary tests, rest unchanged).

PR #599, opened 2026-07-01
feat(desktop): chat-driven Code mode — diffs, terminal output, boundary header

iverc:
Stacks on #595. Stage 4 of the desktop-perfection epic: a chat-driven **Code** experience.

The desktop had no coding surface — agents' file edits and command runs rendered as opaque args/result JSON blobs. This adds one, shaped for the multi-agent reality: an agent codes **on its own host**, in **its own tool root**, within **its permission/sandbox boundary**; the desktop directs and observes. No local editor, no new bridge commands — pure frontend over data already in the transcript and deployment views.

**Code-aware transcript rendering** (benefits chat too)
- `edit_file`/`write_file` render as diffs (±lines, path, created-vs-edited from the real `defra_fs` envelope, `×N` for `replace_all`).
- `bash`/`bash_unrestricted` render as terminal blocks: full shell-joined command line, exit-code badge, sandbox/network chips, and the runtime's `stdout:`/`stderr:` framing parsed into separate streams. Timeouts and `ok:false` envelopes badge as failed even with a null exit code.
- Only completed, uncancelled calls are projected; errored/running/cancelled calls keep the generic disclosure (with denial + cancel-cause UX), so a failed edit never reads as an applied diff.

**Code mode + boundary header**
- A `Code` entry on the sidebar peer card opens a code view reusing the chat surface, headed by the agent's boundary: **agent host** (peer identity — not a possibly agent-relative GraphQL endpoint), **working directory** (`fileToolRoot`), file/bash/command-network modes.
- The boundary is resolved from the **selected behavior's** tool selection (falling back to the deployment default); if none resolves, it honestly reports files & bash off. Copy notes the host operator's tool ceiling may clamp the persisted selection.

**Tests**: a `coding` harness scenario with envelope fixtures that mirror the runtime byte-for-byte (`render_command_output` framing, `EditFileMetadata`/`WriteFileMetadata`); 18 unit tests + 2 e2e journeys (×3 viewports).

Deferred (needs new bridge commands): live file tree, remote-host ceiling read, working-dir scoping control from the Code header.

Verified: build (tsc) + 144 unit + 99 e2e green, non-flaky; adversarial review pass — all 21 confirmed findings fixed (envelope parsing, failure semantics, boundary honesty, fixture fidelity).

### DIFF
--- a/apps/desktop-tauri/src/App.css
+++ b/apps/desktop-tauri/src/App.css
@@ -15,6 +15,7 @@
 @import "./styles/sidebar/nav.css";
 @import "./styles/sidebar/conversations.css";
 @import "./styles/chat.css";
+@import "./styles/code-context.css";
 @import "./styles/fleet/layout.css";
 @import "./styles/fleet/table.css";
 @import "./styles/mcp-health.css";
--- a/apps/desktop-tauri/src/App.tsx
+++ b/apps/desktop-tauri/src/App.tsx
@@ -1,6 +1,7 @@
 import { useState } from "react";
 
 import { ChatWorkspace } from "./components/ChatWorkspace";
+import { CodeContextHeader } from "./components/code/CodeContextHeader";
 import { ConfigWorkspace } from "./components/ConfigWorkspace";
 import { FleetDashboard } from "./components/fleet/FleetDashboard";
 import { Sidebar } from "./components/Sidebar";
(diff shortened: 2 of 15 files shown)


iverc:
Review follow-ups deferred from b8b3527 (the envelope-honesty fix):

- `splitCommandStreams` splits at the first `\nstderr:\n`; stdout containing a literal `stderr:` line leaks the rest of stdout into the stderr pane. Format is ambiguous client-side — worth a comment or a smarter frame.
- Test fixtures say "mirror the runtime's envelopes exactly" but omit `argv`/`cwd`/`duration_ms`/`timeout_ms`/`sandbox`/`*_truncation` fields the runtime always serializes.
- The only-completed-uncancelled projection gate (Transcript.tsx) has no test locking it.
- `write_file` over an existing file renders an all-`+` diff labeled "edited" — prior content loss isn't visible.
- Diff/terminal rendering has no line cap; a huge `write_file` content can stall the render.
- `bashModeLabel` maps a nonexistent `ReadWrite` mode; Back from Configure exits Code mode to Chat.

Issue #608, opened 2026-07-02
Epic: desktop client feature completeness and UI/UX perfection


iverc:
Bring the desktop client to Claude-Desktop-for-macOS quality: every element works, every screen holds up at desktop/laptop/narrow, and the runtime's capabilities are actually operable from the UI.

Grounded in a full multi-agent audit (10 areas, 133 file:line findings). Workstreams:

- **WS0 baseline**: rebase desktop stack onto main (#594 fix + CI gate); regenerate stale visual baselines.
- **WS1 broken (P0, verified)**: Lineage/Interrupt console.log stubs; MCP Probe discards result; Delete Skill dead in packaged app (`window.confirm` under WKWebView); config editors wiped by background snapshot refresh; transcript links navigate the webview away; onboarding CTA dead-ends without a running runtime; fleet rows show global state as per-agent facts; code-mode stdout/stderr marker split.
- **WS2 design system + platform**: missing/unenforced tokens, unloaded brand fonts, type/motion/focus scales, token lint; macOS shell pass (menu, shortcuts, window-state, titlebar); error banner + error boundary.
- **WS3–WS6 screen polish**: chat (streaming UX, copy/highlighting, retry, composer), code mode (diff/terminal quality, live output, approvals), fleet home (guided onboarding, health visibility, peer remove/edit), config (dirty-state, validation, deletes, one paradigm), operations rail (attention badges, liveness, drawer-scale chrome).
- **WS7 completeness**: pairing management UI, network/P2P state view, trace timeline/projection, real ops liveness, cross-node lineage, conversation management, backend editor parity, session fork, context budget.
- **WS8 QA ratchet**: click coverage for fleet actions/skills/add-peer, chat state journeys, baseline-regen policy.

Folds in remaining scope of #580, #518, and the desktop slices of #576 and #368. Supersedes #531 (complete — can be closed).

jackzampolin:
Closing this broad desktop audit as stale after the major desktop packaging, mobile, transcript, tool-call, context, and long-session UX work that landed afterward. The 2026-07 audit no longer describes the current product accurately; remaining desktop defects and missing capabilities should stay or be filed as focused issues against the current implementation.

PR #602, opened 2026-07-01
polish(desktop): onboarding hero, sticky fleet actions, quieter chat chrome

iverc:
Stacks on #599. Home-screen and chat polish from the desktop-perfection visual audit (P1 items).

- **Onboarding hero** — the empty state leads with one-click "Connect Local Agent"; the remote form (connection JSON, DIDs, multiaddrs) moves behind a "Connect a remote agent instead…" disclosure, so first-run isn't a wall of operator plumbing.
- **Fleet row actions** — chat/config/repair buttons grow 24→32px with visible borders and hover states; the actions column is now **sticky**, so at narrow widths the buttons stay pinned and visible instead of scrolling off-screen (the previously-locked scroll-reachability contract is preserved — `.fleet-table-wrap` still owns overflow for the data columns).
- **Quieter chat chrome** — the composer's "42 rows / 21.0 KB" store internals demote to a hover tooltip; the cryptic collapsed "‹Ops" rail handle now reads "‹Operations".
- **Operations drawer** — fully opaque when open (transcript no longer ghosts through the top edge).

CSS/copy-level changes only; no behavior or testid changes. Verified: build + 144 unit + 99 e2e green; screenshots reviewed across desktop/laptop/narrow (all three action buttons visible at 390px).

PR #609, opened 2026-07-02
fix(desktop): wire ops drawer row actions, surface MCP probe results

iverc:
First slice of #608 (WS1 broken elements + WS0 baselines):

- **Lineage** row action in the Background tools panel now focuses the lineage tab on the row's parent request (was a `console.log` stub).
- **Interrupt** row action runs the real preview → cascade-dialog interrupt flow, shared with the composer (was a `console.log` stub).
- **MCP Probe** renders the live probe outcome (status · latency · error) per row instead of discarding the result; probe transport failures show on the row, not the global banner.
- Regenerated the stale visual baselines for the current stack (were last touched by #550).

Fences: `backgrounded-tools-actions.test.tsx` (action wiring + disabled-without-handlers), `McpHealthPanel` container probe tests. Unit 158 passed, e2e 120 passed, visual 3×3 passed.

Stacked on #602.


iverc:
Review notes (two small fixes pushed: lineage-override reset on deployment switch, probe-error clamp):

- The backgrounded-tools snapshot is not agent-scoped (`desktop_operations_snapshot` filters only `await_mode == background`; `AgentToolCall` has no `agent_did`), so on a multi-agent desktop foreign rows appear with enabled Interrupt/Lineage that dead-end in raw "root request not found" banners. Cross-agent interrupt is fenced today only because the preview walk is DID-filtered — `desktop_interrupt_request` itself would latch any request id. Worth a follow-up: per-row agent DID in the snapshot envelope + disabled actions for foreign rows, so the fence is design rather than coincidence.
- Test gaps: ChatWorkspace glue is unfenced (a refactor wiring row-Interrupt straight to `interruptRequest` — skipping the scoped preview — would stay green); probe tests don't cover resolve-with-unreachable-status or assert `is-error` styling.

PR #610, opened 2026-07-02
fix(desktop): external links open in OS browser; in-app delete confirmation

iverc:
Second WS1 slice of #608:

- **External links**: a document-level guard routes absolute http(s)/mailto anchor clicks to the OS browser (opener plugin under Tauri, `window.open` in plain browsers) and suppresses any other anchor navigation — a bare markdown link click previously replaced the whole webview UI with no way back.
- **Delete Skill**: replaced `window.confirm` (always returns false under wry's WKWebView — the button silently no-oped in the packaged app) with a reusable in-app `ConfirmDialog` built on the existing dialog styles.

Fences: `external-links.test.ts` (guard policy incl. relative/unknown schemes), `skill-config-delete.test.tsx` (dialog flow, confirms `window.confirm` is never called). Unit 164 passed, e2e 120 passed, visual 3×3 passed.

Stacked on #609.

### DIFF
--- a/apps/desktop-tauri/src/App.tsx
+++ b/apps/desktop-tauri/src/App.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 
 import { ChatWorkspace } from "./components/ChatWorkspace";
 import { CodeContextHeader } from "./components/code/CodeContextHeader";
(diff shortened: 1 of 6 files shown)


iverc:
Review fixes pushed:

- Link guard moved to capture phase — a component-level `stopPropagation()` (dialog click-swallows, row handlers) could previously switch off the app's navigation boundary for agent output.
- Empty `href` now suppressed: react-markdown rewrites `javascript:`/`data:` destinations to `href=""`, whose default click action reloads the whole webview (agent-controllable state wipe).
- ConfirmDialog: document-level Escape + Tab trap (was backdrop-scoped — one Tab past Confirm made Escape dead and Enter hit the hidden Save button under the overlay).
- Tests: hostile schemes (`javascript:`/`data:`/`file:`///`//host`), empty-href, and stopPropagation-bypass click paths.

Deferred nit: focus not restored to trigger on close (matches existing CascadeCancelDialog behavior).

PR #611, opened 2026-07-02
fix(desktop): preserve in-progress config edits across snapshot refreshes

iverc:
Third WS1 slice of #608. The bridge emits `client-updated` on every store/P2P-health change; the shell re-fetches the snapshot, every config view object gets a fresh identity, and each editor's reset effect re-fired — discarding whatever the operator had typed (any schedule fire, streaming response, or gossip event destroyed mid-edit form state).

All ten config editors now key their reset effects on the document **id** (plus real secondary keys like the selected task id / profile-id set), so forms rehydrate on selection change but survive background refreshes.

Fence: `config-edit-preservation.test.tsx` — typed values survive a full object-tree replacement; switching documents still resets. Unit 166, e2e 120, visual 3×3 green.

Stacked on #610.

### DIFF
--- a/apps/desktop-tauri/src/components/config/AgentConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/AgentConfigPanel.tsx
@@ -68,7 +68,8 @@ export function AgentConfigEditor({
   useEffect(() => {
     setDisplayName(agent.displayName ?? "");
     setEditingDisplayName(false);
-  }, [agent]);
+    // Id-keyed: background snapshot refreshes must not wipe in-progress edits.
+  }, [agent.agentDid]);
 
   const defaultBehaviorId =
     agent.defaultBehaviorId ??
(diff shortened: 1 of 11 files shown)


iverc:
Review fixes pushed:

- BehaviorConfigEditor: the profile-id-set dep re-fired the FULL reset, so any remote profile add/remove still wiped in-progress behavior edits (the bug class this PR fixes). Reset now keys on the behavior id only; the default-profile fallback lives in its own effect keyed on the (sorted — was order-dependent) profile-id set, and keeps a valid current pick.
- AgentConfigPanel: the read-only title froze at hydration, hiding remote renames for the panel's lifetime, and a later Save resurrected the stale name. Title now renders the live document value; the edit buffer seeds on pencil-click and only an in-progress rename overrides on save.
- Fence extended with the behavior-editor profile-churn case.

Remaining gap (fine as follow-up): the fence covers 3 of 10 panels and no new-document case; ToolSelectionConfigEditor has the same residual wipe class via toolServiceIdKey (remote MCP service registration wipes tool-selection edits).

PR #613, opened 2026-07-02
fix(desktop): honest per-agent facts in fleet rows

iverc:
Fifth WS1 slice of #608. Three fleet-row elements showed desktop-global state as per-agent facts:

- **Last update** rendered the desktop client's single global P2P probe timestamp — identical in every row, so a dead remote agent could show "12s ago". Now driven by the deployment's own `runtime.updatedAt` reconciler heartbeat ("unknown" when absent).
- **Tool-ceiling tooltips** claimed "Server ceiling: X" from the local machine's init.json on every row. The claim now appears only on rows sourced from the local runtime (`local-standard`/`server-status`); remote rows omit it.
- **Repair** was labeled "Repair {agent} P2P" per row but invoked the client-wide repair (and disabled every row's button at once). It is now a single fleet-level "Reconnect P2P" control in the header, shown only when a peer or the P2P probe is actually unhealthy.

Fences: fleet-row tests (per-agent heartbeat, ceiling scoping, no row repair), fleet-dashboard tests (header control visibility + fire), updated Playwright fleet-actions and narrow-reachability specs. Visual baselines regenerated. Unit 171, e2e 120, visual 3×3 green.

Stacked on #612.

### DIFF
--- a/apps/desktop-tauri/src/components/fleet/FleetDashboard.tsx
+++ b/apps/desktop-tauri/src/components/fleet/FleetDashboard.tsx
@@ -55,6 +55,16 @@ export function FleetDashboard({
   const [peerFormError, setPeerFormError] = useState<string | null>(null);
   const [localRuntimeError, setLocalRuntimeError] = useState<string | null>(null);
   const hasDeployments = deployments.length > 0;
+  // The repair command re-dials the desktop client's P2P connections as a
+  // whole, so it lives here at fleet level — a per-row placement would lie
+  // about its scope. Shown only when something actually needs repairing.
+  const needsP2PRepair =
+    deployments.some(
+      (deployment) => !deployment.dialSucceeded || Boolean(deployment.lastError),
+    ) ||
+    (p2pHealth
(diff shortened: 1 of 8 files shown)


iverc:
Review fixes pushed (branch also rebased onto main to drop the stale pre-rebase #612 commit):

- `isLocalRuntimeSource` accepted `server-status`, but the peer directory stamps ANY graphql-bearing peer with it (Add-connection form, `init --status-endpoint`, the demo's worker node) — so remote rows still claimed the local machine's init.json tool ceiling, the lie this PR removes. Now `local-standard` only; the negative fence uses `server-status` (was `manual`, which nothing emits).
- `runtime.updatedAt` is change-gated (RuntimeStatusHandle::update persists only on field change; the 1s executor observer writes only on deltas), not a heartbeat — a healthy idle agent ages identically to a dead one. Reworded the tooltip to say it's a state change, not liveness. Proper fix is a periodic AgentRuntime heartbeat write in the runtime — follow-up.

Clean checks for the record: runtime rows genuinely per-agent; future clock skew clamps to "0s ago"; header Reconnect predicate grounded in continuously-maintained peer state, no hidden-while-dead hole found.

PR #614, opened 2026-07-02
fix(desktop): design-token integrity, bundled brand fonts, layer conformance

iverc:
First WS2 slice of #608 — the design-system foundation fixes, each previously silent:

- **Missing tokens defined**: 18 declarations referenced `--font-mono`/`--font-body`/`--space-9`/`--space-10`, which never existed — per `var()` semantics the declarations invalidated silently (collapsed backend-health padding/gaps, un-monospaced IDs and durations). Now defined in tokens.css; the inconsistent per-sheet mono stacks unify on `--font-mono`.
- **Brand fonts actually ship**: tokens.css declared Inter and Funnel Display but nothing loaded them — every user silently fell back to system fonts. Now bundled via `@fontsource-variable/*`, imported from App.css so both the app entry and the test harness get them (an earlier main.tsx-only import produced pixel-identical screenshots — the harness mounts App directly).
- **Cascade layers repaired**: backend-health.css and backgrounded-tools.css had no `@layer` (unlayered rules outrank every layer; their declared layers sat empty); cancel-ux.css squatted in the top-precedence `utilities` layer. All three now sit in their proper layers.
- **Fence**: `design-system-conformance.test.ts` — fails CI on any fallback-less `var()` referencing an undefined token, any top-level rule outside `@layer`, and any layer name missing from App.css's declared order.

Visual baselines regenerated (real fonts change every screen). Unit 175, e2e 120, visual 3×3 green.

Stacked on #613.


iverc:
Review notes (comment fix pushed; branch was rebased onto main earlier to drop stale stack commits):

- Clean on the high-stakes checks: no layer-precedence regressions found from the cancel-ux demotion or the two newly-layered sheets; fonts bundle locally (no CDN); token values consistent with the scale.
- Latent fence gap, fine as follow-up: the conformance test collects all `--x:` definitions into one global union with no scope/media awareness — a token defined only inside a media query passes the fence but still silently invalidates its uses outside it. Also only `src/styles` + `App.css` are scanned; a co-located `.css` elsewhere under `src/` would be unchecked.

PR #615, opened 2026-07-02
polish(desktop): tokenized type scale

iverc:
Second WS2 slice of #608. The app had 24 distinct font-size values (18 px sizes incl. four fractional, plus mixed rem) hard-coded across area sheets, with hierarchy carried by dozens of local overrides.

- Adds a 10-step `--text-*` scale to tokens.css, honest to real usage (dense operator UI lives at 10–13px).
- Sweeps every `font-size` declaration onto the nearest step — most map pixel-identically; ~15 declarations shift by ≤1px (fractional and off-scale values rounded); three 26px declarations (markdown h1, conversation-rename input, MCP empty-state emoji) round up +2px to `--text-3xl`/28px — the 22→28 gap has no nearer step.
- Fence extended: `design-system-conformance.test.ts` now fails any raw `font-size` value outside the token scale.

Heading-role normalization (base h2 = 21px vs chat 28 / ops-rail 16 / sidebar 18) is deliberately NOT in this slice — which element gets which size is per-screen hierarchy design, owned by the WS3–WS6 polish passes. This slice only makes the vocabulary they'll use.

Visual baselines regenerated (sub-pixel rounding shifts). Unit 176, e2e 120, visual 3×3 green.

Stacked on #614.


iverc:
Review notes (fence hardening pushed as 5e2c1d5; PR body corrected):

- The sweep claim was off: three 26px declarations shift **+2px** to `--text-3xl`/28px (markdown h1, conversation-rename input, MCP empty emoji) — the 22→28 gap has no nearer step. Code is fine (baselines capture it); the body now discloses it.
- Fence gaps closed: the longhand check missed `FONT-SIZE:` (case) and the last declaration of a block (no `;`); the `font:` shorthand could smuggle a raw size entirely — now only `font: inherit` passes.

Clean for the record: scale monotonic, all 10 steps used, rem conversions land within 1px at the 16px root, no px line-height mismatches, no meaningful hierarchy collisions.

PR #616, opened 2026-07-02
polish(desktop): spacing token sweep + raw-literal ratchets

iverc:
Third WS2 slice of #608, deliberately value-preserving:

- Sweeps every exact-match px value inside spacing declarations (`padding`/`margin`/`gap` and variants) onto the `--space-*` scale — 71+ raw values down to 40, all remaining being hairline nudges (1–3px) and one-offs owned by the per-screen passes. Pixel-identical by construction: visual verification passes against the unchanged baselines.
- Adds **ratchet fences** to `design-system-conformance.test.ts`: raw spacing px (ceiling 40), raw `rgb()` literals (90), bespoke `box-shadow`s (38) may only go down — new raw literals fail CI, and each polish pass lowers the ceilings.

Shadow/color tokenization proper is left for the per-screen passes where the visual decisions belong; the ratchets stop the bleeding now.

Unit 179, e2e 120, visual 3×3 green (no baseline changes).

Stacked on #615.

### DIFF
--- a/apps/desktop-tauri/src/styles/backend-health/backend-health.css
+++ b/apps/desktop-tauri/src/styles/backend-health/backend-health.css
@@ -471,7 +471,7 @@
   }
 
   .backend-health__empty-body code {
-    padding: 1px 6px;
+    padding: 1px var(--space-2);
     background: var(--color-surface-strong);
     border-radius: var(--radius-sm);
     border: 1px solid var(--border-subtle);
--- a/apps/desktop-tauri/src/styles/backgrounded-tools.css
+++ b/apps/desktop-tauri/src/styles/backgrounded-tools.css
@@ -19,18 +19,18 @@
   .background-tools-panel .chip-row {
     display: flex;
     flex-wrap: wrap;
-    gap: 6px;
+    gap: var(--space-2);
     align-items: center;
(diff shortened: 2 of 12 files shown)


iverc:
Review fixes pushed (plus earlier: rgb() ratchet was at 91/90 after the sticky-cell fix on main — simplified that background to `var(--color-bg)`, count now 89):

- Box-shadow ratchet counted every `box-shadow:` declaration including future `none`/`var(--shadow-*)` values — tokenizing shadows could never lower the pressure, and the first conformant token use would fail CI at the zero-slack ceiling. Now only bespoke values count.
- Spacing ratchet: logical longhands (`margin-inline-start` etc.) escaped the fence entirely (raw-px smuggling path); `scroll-padding-*` miscounted as `padding-*`. Both fixed.
- Verified clean: the sweep is pixel-identical (sampled 15+ swept declarations against the token values, incl. multi-value shorthands); remaining 40 raw spacing values are genuine hairlines/one-offs.

Known friction, deliberate: negative hairline nudges (`margin-top: -2px`) count against the zero-slack ceiling and force either a token rewrite or ceiling review — that's the ratchet working, just be aware.

PR #617, opened 2026-07-02
polish(desktop): motion tokens, global reduced-motion, focus-ring integrity

iverc:
Fourth WS2 slice of #608, value-preserving like the last:

- **Motion tokens**: `--motion-fast/medium/pulse` replace every hard-coded duration (all 30 sites; same values, so pixel-identical — visual passes against unchanged baselines).
- **Reduced motion globally honored**: one `prefers-reduced-motion` override in base.css kills all animation/transition durations; the single per-sheet override it obsoletes is removed.
- **Focus outlines never removed**: the two `outline: none` declarations (ops-rail collapsed/close buttons, which masked focus with only a subtle background tint) now get a real `:focus-visible` ring.
- **Fences**: durations must use `--motion-*` tokens; `outline: none|0` fails CI.

Unit 181, e2e 120, visual 3×3 green (no baseline changes).

Stacked on #616.

### DIFF
--- a/apps/desktop-tauri/src/styles/backend-health/backend-health.css
+++ b/apps/desktop-tauri/src/styles/backend-health/backend-health.css
@@ -100,7 +100,7 @@
     border: 1px solid var(--border-subtle);
     border-radius: var(--radius-lg);
     overflow: hidden;
-    transition: border-color 120ms ease;
+    transition: border-color var(--motion-fast) ease;
   }
 
   .backend-health__row[data-state="available"] {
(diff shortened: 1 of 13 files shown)


iverc:
Review fixes pushed:

- Focus fence bypasses closed: `outline: 0px`, `outline-style: none`, `outline-width: 0`, `!important`, and case variants all passed the old regex — the exact regression class the fence claims to stop.
- Reduced-motion override now includes `scroll-behavior: auto !important`: the transcript panel's `smooth` scroll made streaming auto-follow animate continuously for reduced-motion users.

Verified clean: the 0.01ms (not 0s) duration kill keeps transitionend/animationend firing — no listener in src depends on real durations; all 30 swept duration sites are value-identical; the removed per-sheet override is fully covered by the global one.

PR #618, opened 2026-07-02
fix(desktop): app-level error boundary

iverc:
WS2 item 12 of #608. Any render exception previously white-screened the entire desktop app — no message, no recovery.

`ErrorBoundary` now wraps the app shell (inside `App`'s default export, so the app entry, the UI-test harness, and any future mount all get it). The fallback is a styled panel: plain-language message, collapsible error details, and a Reload action. Hook errors in `useDesktopShell` are covered too since the boundary sits above the shell component.

Fence: `error-boundary.test.tsx` (happy path renders children; a throwing child renders the fallback with details and a working Reload). Unit 183, e2e 120, visual 3×3 green (no baseline changes).

Stacked on #617.

### DIFF
--- a/apps/desktop-tauri/src/App.tsx
+++ b/apps/desktop-tauri/src/App.tsx
@@ -3,6 +3,7 @@ import { useEffect, useState } from "react";
 import { ChatWorkspace } from "./components/ChatWorkspace";
 import { CodeContextHeader } from "./components/code/CodeContextHeader";
 import { ConfigWorkspace } from "./components/ConfigWorkspace";
+import { ErrorBoundary } from "./components/ErrorBoundary";
 import { FleetDashboard } from "./components/fleet/FleetDashboard";
 import { Sidebar } from "./components/Sidebar";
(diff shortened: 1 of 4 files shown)


iverc:
Review fixes pushed:

- Details panel showed only `error.message` — stack and component stack went to a console the packaged app can't show, making crash reports undebuggable. Both now render in the collapsed `<pre>` (whose CSS already anticipated them).
- `role="alert"` + autofocus on Reload: the fallback swap previously dropped focus to `<body>` with no screen-reader announcement.
- `text-align: left` on the panel (inherited centering made wrapped error text ragged-center inside left-aligned boxes).
- Fences: a new test asserts App actually wraps the shell in the boundary (deleting the wrapper previously passed the whole suite); the location mock restores in `finally`.

Known scope gap, worth a follow-up note: React boundaries don't catch event-handler or async/promise errors, and no `window.onerror`/`unhandledrejection` handler exists — a rejecting Tauri invoke inside `useDesktopShell`'s effects still surfaces nowhere. The body's "hook errors are covered" holds only for render-phase throws.

PR #648, opened 2026-07-07
feat(retry): per-completion retry in the owned loop (#631)

### LINKED ISSUE #631: Runtime never retries failed completion requests — transient provider 400s and backend outages silently kill agentic runs
## Summary

When a completion request fails mid-run — HTTP 400 from the provider or a transport error — the runtime marks the AgentResponse `error` and stops. There is no retry, no backoff, no re-submission. For scheduled/agentic runs this means the entire run's work is discarded (no `post_status`), and the failure is only visible if someone reads run history.

## Production evidence (amygdala fleet)

- **2026-07-01 backend outage window (~20 min):** 17 scheduled steward runs died with `agent stream failed: ProviderError: Http client error` — every steward whose slot landed in the window lost its run entirely. One retry after 30s would have saved most of them.
- **d4f raw-escape 400s (~20% of tool-heavy calls, vLLM raw-mode array params):** `400 Bad Request: Expecting ',' delimiter` kills runs at the *final* post_status step — 3 fleet hits on 2026-07-06 alone, each discarding a completed multi-minute run. The trigger is a single invalid escape the model emitted in tool-call args; a retry gives the model a fresh sample and nearly always succeeds (operator re-kicks succeed ~80%+ first try).
- Session-poison persistence from garbled args is fixed (#589/#590), but that covers ingest/egress normalization — the *live request* that 400s still terminates the run.

## Proposed behavior

For scheduled/task runs (idempotent from the runtime's perspective — the conversation transcript is the state):

1. **Transport errors / 5xx / connection failures:** retry the completion with exponential backoff (e.g. 3 attempts, 5s base), same transcript.
2. **Provider 400s:** retry once or twice — for sampling-induced 400s (invalid escapes in emitted tool args) a fresh sample usually clears it. If the 400 persists, run the #589-style transcript repair (re-serialize/sanitize the last assistant tool-call message) and retry once more before giving up.
3. Surface retries in run history (`retry_count`, last error) so operators can distinguish flaky-but-recovered from dead.

Interactive sessions could adopt the same with a tighter budget.

## Why now

The fleet's steward swarm runs hourly checks against one backend; at the measured 400 rate, roughly 1 in 25–30 tool-heavy runs dies silently. Under multi-agent fan-out (the swarm demo path) exposure multiplies with subagent count — this is currently the largest single source of lost agent work in the fleet.

Related: #438 (owned provider layer — natural home for this), #589/#590 (arg normalization, closed).

### LINKED ISSUE #638: Daemon-level inference retry can re-execute tools on text-less runs (side-effect replay)
## Summary

The daemon's inference attempt loop (\`agent/daemon/inference.rs\`) gates whole-run retry on \`!processor.has_observable_activity()\`. That predicate is:

- \`assistant_turn.has_content()\` — drained at every turn boundary (taken on ToolResult / FinalResponse), and
- \`!streamed_text.trim().is_empty()\` — accumulates **visible text only**, across turns.

For a run whose turns emit **only tool calls and no visible text** (common with tool-heavy behaviors and some models), both are false when a later completion fails with a retryable classification. The outer loop then re-invokes \`run_loop_stream\` from scratch with the original history — and every tool in the run **re-executes** (fresh call ids, real side effects: writes, posts, subagent spawns).

## Trigger conditions

1. Multi-turn run where no turn produced visible assistant text before the failure point.
2. A completion fails with a retryable classification (transport / 5xx / vLLM parse-signature 400) and it is not a liveness timeout.
3. \`retry_policy.max_retries > 0\` (default 3).

## Why it hasn't bitten loudly

Any visible text anywhere in the run permanently disables the outer retry (\`streamed_text\` is never reset), which is also why the 26 retryable production failures in #631's 48h window all died at attempt=1 — the gate is simultaneously too strict for recovery and too loose for safety.

## Fix

Resolved structurally by #631's redesign: retry moves inside the owned completion loop at the per-completion boundary (re-issue only when the current turn has no effects or was durably closed), and the daemon-level whole-run retry path is removed. The no-tool-re-execution guard becomes a Lean obligation in the new CompletionRetry model.

Related: #631 (retry design), #512 (where the had_observable_activity gate was reviewed in another context).

jackzampolin:
## Summary

Implements per-completion retry inside the owned loop instead of retrying the whole daemon run. Each provider attempt is admitted independently and projected as its own `InferenceCall` row, with `run_timeline` deriving `RetrySummary` from the attempt rows.

Closes #631.
Closes #638.

## Triage

The deterministic vLLM 400 sample was from the pre-#601 fleet binary. The post-#601 binary shows zero occurrences, so there is no new #601 gap here. The parse-400 repair path is retained as a regression fence and is covered by the prod payload tape.

## Mechanisms

- Pre-stream retry: retries transport/5xx/rate-limit failures with deadline-aware backoff, resamples parse-signature 400s once, then repairs deterministic repeated parse failures.
- Mid-stream before effects: retracts streamed assistant output and resamples when no tool effect has occurred.
- Mid-stream after tool effects: closes the current turn with the executed tool results and continues, avoiding tool replay.
- Daemon retry removal: removes the old whole-run inference retry path, closing the tool replay hazard from #638.

## Formal Model

- Added the CompletionRetry Lean model and Rust mirror/conformance fence.
- N1: reissue/repair phases perform no effects.
- N2: same-turn visible output only decreases when paired with a retraction.
- N3: once effects occur, retry continuation preserves them and does not rerun tools.
- N4: retry actions are bounded by request budget/deadline.
- N5: terminal failures stop retrying and preserve terminality.

## Acceptance

- Backend restart cluster recovers with one failed row and one completed row per request.
- Tight deadlines fail cleanly before sleeping into a retry that cannot fit.
- Interactive defaults spend only the short interactive retry budget.
- The deterministic prod 400 tape resamples once, repairs, and recovers.
- A first stream poll error now follows pre-stream retry policy, covering adapters that surface HTTP status failures after stream construction.

## Validation

- `cargo fmt -p defra-agent`
- `git diff --check`
- `cargo test -p defra-agent --lib failure_class_parse_signature_can_come_from_classified_reason`
- `cargo test -p defra-agent --lib first_stream_poll_parse_400_uses_pre_stream_retry_policy`
- `cargo test -p defra-agent --test e2e_runtime completion_retry_tape`
- `(cd crates/defra-agent/proofs && lake build)`
- `cargo test -p defra-agent`

Lean passed with existing warnings in unrelated Background proof files.


jackzampolin:
## Ultracode review — 2 of 4 confirmed findings fixed in this branch, 2 handed back

Ran an 8-dimension multi-agent review (adversarial verify + completeness critic). 4 confirmed findings (1 HIGH, 2 MEDIUM, 1 LOW), 3 refuted. I've pushed fixes for the two that are clean, verifiable correctness bugs; the other two need an author call, detailed below.

### Fixed in this branch

**HIGH — one-shot retract corrupted the durable transcript.** `run_loop_to_text`'s `TurnRetracted` arm was a bare `continue` and never reset its accumulator (`loop_stream.rs`), while the daemon's `StreamProcessor` does reset (accumulator + streamed-text truncate + tail rewind). Since `oneshot.rs` drives `run_loop_to_text` with a *persisting* hook, a mid-stream failure with no tool executed → retract → resample concatenated the retracted partial into the persisted assistant message (`"Based on" + "The answer is 42"` → `"Based onThe answer is 42"`). The returned string stayed correct, masking it, but the durable transcript — which feeds future history and training capture — was corrupted, violating the design's "exactly one turn per turn index" invariant on the one-shot path.
Fix: reset the accumulator on `TurnRetracted`, mirroring `StreamProcessor`. Fenced by `run_loop_to_text_retract_persists_only_the_resample` (verified red against pre-fix: it persisted the concatenation).

**MEDIUM — retry backoff could trip the liveness timeout and defeat the retry.** The backoff sleep now runs inside the loop generator, spanning the next `stream.next()` poll, which the daemon wraps in `timeout(liveness_timeout, …)`. A backoff step longer than `stream_liveness_timeout_secs` fired a spurious `"stream liveness timeout"` terminal failure — killing the very retry the PR adds, in the acceptance tape's own backend-restart scenario, under a plausible config (the test harness itself uses ~45–60s liveness against the 120s ladder step; shipped defaults happen not to collide). Three independent review agents converged on this.
Fix: the daemon now reads `AttemptFailed{backoff, will_retry:true}` and extends the *next* poll's liveness budget by that backoff (`inference.rs`), so a legitimate backoff is never misread as a dead stream.

### Handed back — need an author decision (not fixed)

**MEDIUM — repair re-sanitizes `new_messages` only, not loaded history.** `repair_provider_input` (`loop_stream.rs`) iterates only run-threaded `new_messages`. The deterministic-400 the repair path exists to fence originates in the *input transcript* = loaded `history` (borrowed `&[Message]` at the repair point). So repair wouldn't fix its own motivating case. I did **not** fix this because the correct fix threads mutable history through the proven assembly path (`build_request`), which I won't restructure blind on a formally-verified PR — and the design's own triage says this 400 class is dead post-#601 (egress normalization covers all completion inputs incl. threaded history). Decision for you: extend repair to cover history, **or** document it as belt-and-suspenders for a #601-covered class and add a poisoned-*history* fixture proving whichever path is responsible.

**LOW — resample budget diverges from the Lean model.** Rust effectively caps resamples at the transport-ladder length (`completion_retry.rs`); the Lean model treats `resampleRetries` as independent, and no conformance case catches the divergence (e.g. `retry_max_resample=2` with a length-1 `retry_backoff_ms`). Not fixed because reconciling it is a Lean-first semantics call (clamp Rust to match Lean, or extend the model + a conformance row) that's yours to make — a blind Rust clamp risks encoding the wrong intent.

### Refuted (3) — one worth a 5-minute Lean check
Correctly killed: interactive-inherits-the-minutes-ladder (has its own quick path), rate-limit-hint-jittered-below-window. The third is in the sensitive theorem-absorbs-conclusion class — "N1 (`n1_reissue_requires_no_open_effects`) presupposes its goal; the real fence is `reissue_inv_preserved`, and no lemma establishes `ReissueInv` at the initial state." The adversarial verifier refuted it as not-real, but given how much this repo cares about that failure mode, worth the Lean owner confirming `ReissueInv` is genuinely established at init. Cheap insurance, not a blocker.

jackzampolin:
## Second ultracode pass (after e2d4da47) — found my own FIX-B was incomplete; fixed + fenced in 85cdf072

Ran a re-review weighted toward adversarially stressing the two fixes from the last round. It caught a real gap in FIX-B, triangulated by 4 independent dimensions.

### Confirmed + fixed (85cdf072)
**FIX-B was asymmetric.** Last round I extended the daemon liveness budget by `AttemptFailed{backoff}` (pre-stream + close-and-continue), but the sibling **retract-and-resample** path yields `TurnRetracted`, which carried no backoff and fell into the ZERO arm. So a retract backoff longer than `stream_liveness_timeout` was still misread as a dead stream → spurious terminal "stream liveness timeout" → defeated the resample. Exactly the class FIX-B was meant to close, left open on the sibling path.
Fix: added `backoff` to `TurnRetracted`, populated at the retract emission, and the daemon now reads it via an or-pattern (`AttemptFailed{will_retry:true} | TurnRetracted`), so both mid-stream retry paths extend the liveness budget symmetrically.

**FIX-B was also unfenced** (the critic's other point — reverting it kept the suite green). Now fenced by `backoff_longer_than_liveness_timeout_recovers`: it drives a real agent through the daemon with `liveness=1s` and a backoff exceeding it, and I verified it **red-then-green** — reverting the extension leaves the request `lifecycle_state=failed` with a spurious liveness timeout instead of `completed`. Full package gate green.

### Still open — one author decision (not fixed)
**Compaction & title generation now inherit the scheduled retry ladder with `deadline: None`.** This is a PR-wide interaction the critic surfaced (not from the two fixes): before this PR, `run_loop_to_text` (rig `Agent::prompt`) didn't retry, so a transient provider error during inline compaction/title failed instantly. Now those paths inherit `scheduled_default()` = [5s,30s,120s], so an inline compaction can block ~155s uninterruptibly on a flapping backend. It doesn't corrupt anything, but it's a real, user-visible behavior change. Recommend bounding it — e.g. a minimal/no retry budget (or a deadline) for the compaction/title execution origin — before merge. That's your call on intended semantics.

### Refuted (1)
Interactive-inherits-the-minutes-ladder — refuted again (interactive has its own quick path), consistent with round 1.

Merge-readiness (critic): mergeable-with-caveats; residual risk is the compaction/title retry-inheritance above. The two previously-deferred items (repair-history MEDIUM, resample-vs-Lean LOW) still stand as handed back.

jackzampolin:
Bounded compaction and title generation to **no retry budget** (6f256ba0), per agreement — this closes the last open item from the re-review.

- `CompletionRetryPolicy::no_retry()` (empty ladder, no resample, no repair).
- `DefraCompactor::new` forces it even when handed a `scheduled_default` config; title generation forces it too (it already retries at its own `generate_title_with_fallback` layer).
- Fence `compaction_fails_fast_without_retrying_on_transient_error`: a transient provider error yields exactly one provider call and `compact()` returns within a 3s timeout — verified **red-then-green** (reverting the override makes `compact()` block in the 155s ladder and the test fails at the timeout). Full package gate green.

All findings from both ultracode passes are now addressed in-branch. Remaining handed-back items (author's call, non-blocking): the repair-history MEDIUM and the resample-vs-Lean LOW.

jackzampolin:
The two handed-back items now have trackers so they don't block the merge:
- #652 — repair path re-sanitizes only `new_messages`, not loaded history (MEDIUM; likely #601-covered, decision: extend repair vs document + poisoned-history fixture)
- #653 — resample budget capped at transport-ladder length diverges from the Lean model, unfenced (LOW; Lean-first: clamp Rust or extend the model + conformance row)

The optional Lean sanity check (confirm `ReissueInv` is established at the initial state for N1) is noted here rather than filed — it's a 5-minute confirmation, not tracked work.

Issue #647, opened 2026-07-07
Server backend configuration


nasdf:
On first launch we should configure and store available backend servers. 

When no backend servers are configured:
- If no local server is detected, the user can choose to launch a local server or configure a remote backend. 
- If a local server is detected we should automatically connect to it

When backend servers are saved:
- with a single backend: stored automatically connect to it. 
- with multiple backends: prompt the user to select one.

Additionally we should add a UI menu to switch between backends and add new ones.

jackzampolin:
Closing as an outdated desktop connection and onboarding sketch. The desktop and fleet architecture has changed substantially since this was filed; any remaining server-selection or first-launch defect should be captured as a concrete issue against the current UI rather than retained through this old umbrella.

Issue #640, opened 2026-07-07
Inference backend probes report false-healthy for dead endpoints — last_probe never set, metric constant


jackzampolin:
## Summary

Inference backend probing reports healthy for endpoints that are hard down. The probe status appears to be a constant, not a measurement: `probe_status="healthy"` with `last_probe=null`, and the `defra_agent_backend_probe_status` metric is constant-1.

## Production evidence (amygdala fleet, v0.6.3)

- spark-1 (backend `sparks-cluster`, `100.74.68.88:8000/v1`) has been physically offline since 2026-07-06 21:43Z.
- ~16h later, **all 16 reachable runtimes** still report that backend as `enabled=true, probe_status="healthy", last_probe=null`.
- During the 2026-07-01 outage window, runtimes routed requests to dead backends for ~3h with the metric pinned at 1 throughout (noted at the time; this issue is the tracker).

Nothing currently routes to `sparks-cluster` by default (manual failover only), which is why this is silent — but the failover path is booby-trapped: any operator or behavior change that selects it inherits a "healthy" dead endpoint, and dashboards/alerts built on the metric can never fire.

## Proposed behavior

1. Actually execute the probe on its schedule; set `last_probe` on every attempt.
2. Connect failure / non-2xx / timeout ⇒ `probe_status="unhealthy"` (with hysteresis if desired).
3. `defra_agent_backend_probe_status` reflects the measured state per backend.
4. Optional: behaviors skip enabled-but-unhealthy backends during selection, so registry hygiene lapses degrade gracefully.

A probe subsystem that cannot report unhealthy is worse than none — it converts "no data" into "false confidence" (#631's retry logic would also benefit from a truthful signal for fail-fast vs backoff decisions).

jackzampolin:
2026-07-09 triage update:

Closing as completed by PR #644 (`47ba687c`, merged 2026-07-07).

Current main has:
- scheduled per-runtime backend probing;
- K-failure demotion and single-success promotion in `backend_health`;
- measured unhealthy state merged into admission/snapshot availability;
- Prometheus metrics overlaid with measured health, so dead endpoints report value `0` with a measured status label;
- `last_probe`/probe freshness surfaced through the measured-health path and documented in `docs/backends.md`.

If this reappears in production, open a new issue against the measured-health/prober path rather than reopening the old startup-only false-healthy bug.

PR #650, opened 2026-07-08
feat(cli): backend onboarding flow — shared core + onboard command (#647)

iverc:
First slice of the #647 server flow.

Promotes the demo's ad-hoc backend detection into a first-class, tested `onboarding` module and adds an interactive `defra-agent onboard` command.

- `onboarding.rs`: a live `/models` health probe, local-server detection, and the #647 decision tree as a pure, unit-tested function — one stored backend (or a detected local one) auto-connects, multiple prompt for a pick, none-plus-detected adopts, none-plus-nothing offers to start a local server (prints instructions to re-run) or configure a remote one.
- `onboard` command: applies that tree and binds the chosen backend and its model to the default behavior. Writes go through `resolve_config_access`, so it works against a running server or offline and never opens a second node on the server's data dir. Non-interactive escape via `--backend-preset` / `--inference-url` / `--api-key` / `--model` for headless and CI.
- `serve`: when the default behavior is not runnable, it now points at `onboard` instead of leaving a silent "degraded".
- The demo's probe/detect is deduped onto the shared module.

Verified: `cargo check --workspace --all-targets` clean, fmt clean, 8 new unit tests (decision tree + the behavior-bind input builder, including GraphQL-escaping of interpolated values).

Deferred to follow-ups: the desktop backend menu (647b), starting a default agent on server startup, and honest backend probes (#640) so a required backend can be enforced as actually working.

Stacked on #635 (which carries the pre-existing main compile repairs); retarget to main once that lands.

### DIFF
--- a/crates/defra-agent-cli/src/cli/args.rs
+++ b/crates/defra-agent-cli/src/cli/args.rs
@@ -158,6 +158,41 @@ pub(crate) enum Command {
     },
     #[command(about = "Interactive, self-contained fleet demo (single node -> paired fleet)")]
     Demo(DemoArgs),
+    #[command(
+        about = "Configure and connect an inference backend for an initialized agent (#647)"
+    )]
+    Onboard(OnboardArgs),
+}
+
+#[derive(clap::Args)]
+pub(crate) struct OnboardArgs {
+    #[arg(long, help = "Agent home directory (defaults to the standard home)")]
+    pub(crate) home: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "GraphQL endpoint of a running server. Defaults to auto-detecting a live server for this home, else operating on the home offline"
+    )]
+    pub(crate) graphql: Option<String>,
+    #[arg(
+        long,
+        help = "Non-interactive: configure this backend preset (openai, openrouter, ollama, llama-cpp, vllm)"
+    )]
+    pub(crate) backend_preset: Option<BackendPresetArg>,
+    #[arg(
+        long,
+        help = "Non-interactive: configure this backend base URL, e.g. http://127.0.0.1:8080/v1"
+    )]
+    pub(crate) inference_url: Option<String>,
+    #[arg(
+        long,
+        help = "Model name to bind (defaults to the detected/preset model)"
+    )]
+    pub(crate) model: Option<String>,
+    #[arg(
+        long,
+        help = "API key stored in the backend document. Prefer an env-var backend for secrets"
+    )]
+    pub(crate) api_key: Option<String>,
 }
 
 #[derive(clap::Args)]
--- a/crates/defra-agent-cli/src/cli/args.rs
+++ b/crates/defra-agent-cli/src/cli/args.rs
@@ -158,9 +158,7 @@ pub(crate) enum Command {
     },
     #[command(about = "Interactive, self-contained fleet demo (single node -> paired fleet)")]
     Demo(DemoArgs),
-    #[command(
-        about = "Configure and connect an inference backend for an initialized agent (#647)"
-    )]
+    #[command(about = "Configure and connect an inference backend for an initialized agent")]
     Onboard(OnboardArgs),
 }
 
--- a/crates/defra-agent-cli/src/cli/args.rs
+++ b/crates/defra-agent-cli/src/cli/args.rs
@@ -183,14 +183,19 @@ pub(crate) struct OnboardArgs {
     pub(crate) inference_url: Option<String>,
     #[arg(
         long,
-        help = "Model name to bind (defaults to the detected/preset model)"
+        help = "Model to bind. With --backend-preset/--inference-url: required unless the preset has a safe default (same rule as init). On the detect path: an optional override of the server's advertised model. Ignored when rebinding an already-stored backend"
     )]
     pub(crate) model: Option<String>,
     #[arg(
         long,
-        help = "API key stored in the backend document. Prefer an env-var backend for secrets"
+        help = "API key stored in the backend document. Prefer --api-key-env-var for secrets"
     )]
     pub(crate) api_key: Option<String>,
+    #[arg(
+        long,
+        help = "Environment variable name the runtime reads the API key from (kept out of the stored document; mirrors init)"
+    )]
+    pub(crate) api_key_env_var: Option<String>,
 }
 
 #[derive(clap::Args)]
--- a/crates/defra-agent-cli/src/commands/onboard.rs
+++ b/crates/defra-agent-cli/src/commands/onboard.rs
@@ -36,6 +36,12 @@ use crate::{
 
 const DEFAULT_MODEL: &str = "gpt-4.1-mini";
 const DEFAULT_MAX_CONCURRENT: i64 = 8;
+/// A backend onboarding just probed successfully (detected-local path): safe to
+/// mark runnable now instead of waiting for the server's startup re-probe.
+const HEALTHY_PROBE_STATUS: &str = "healthy";
+/// A backend onboarding did not measure (remote / preset / scripted): the
+/// startup probe (#640) decides its health.
+const UNKNOWN_PROBE_STATUS: &str = "unknown";
 
 pub(crate) async fn onboard(args: OnboardArgs) -> Result<()> {
     let home_dir = resolve_home_dir(args.home.as_deref());
@@ -79,7 +85,8 @@ pub(crate) async fn onboard(args: OnboardArgs) -> Result<()> {
             .model
             .clone()
             .unwrap_or_else(|| DEFAULT_MODEL.to_string());
-        let backend_id = persist_backend(&access, &agent_did, &resolved, &model).await?;
+        let backend_id =
+            persist_backend(&access, &agent_did, &resolved, &model, UNKNOWN_PROBE_STATUS).await?;
         bind_default_behavior(&access, &default_behavior_id, &backend_id, Some(&model)).await?;
         return report(&backend_id, "configured", &default_behavior_id);
     }
@@ -103,7 +110,12 @@ pub(crate) async fn onboard(args: OnboardArgs) -> Result<()> {
                 None,
                 BackendResolutionMode::Init,
             )?;
-            let backend_id = persist_backend(&access, &agent_did, &resolved, &model).await?;
+            // detect_local_backend already probed this endpoint's /models, so
+            // record it healthy now — the behavior is runnable without waiting
+            // for the server to re-probe the same live server on next startup.
+            let backend_id =
+                persist_backend(&access, &agent_did, &resolved, &model, HEALTHY_PROBE_STATUS)
+                    .await?;
             bind_default_behavior(&access, &default_behavior_id, &backend_id, Some(&model)).await?;
             println!(
                 "Detected a local inference server at {} — connected.",
@@ -124,8 +136,14 @@ pub(crate) async fn onboard(args: OnboardArgs) -> Result<()> {
                 // bound — the user re-runs `onboard` once the server is up.
                 None => Ok(()),
                 Some((resolved, model)) => {
-                    let backend_id =
-                        persist_backend(&access, &agent_did, &resolved, &model).await?;
+                    let backend_id = persist_backend(
+                        &access,
+                        &agent_did,
+                        &resolved,
+                        &model,
+                        UNKNOWN_PROBE_STATUS,
+                    )
+                    .await?;
                     bind_default_behavior(&access, &default_behavior_id, &backend_id, Some(&model))
                         .await?;
                     report(&backend_id, "configured", &default_behavior_id)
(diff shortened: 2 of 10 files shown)


crates/defra-agent-cli/src/commands/onboard.rs
```diff
@@ -0,0 +1,371 @@
...
+            None,
+            None,
+            args.api_key.as_deref(),
+            None,
+            BackendResolutionMode::Init,
+        )?;
+        let model = args
+            .model
+            .clone()
+            .unwrap_or_else(|| DEFAULT_MODEL.to_string());
```
jackzampolin: **[bug]** Non-interactive onboarding (`--backend-preset` / `--inference-url`) defaults the model to the hardcoded `DEFAULT_MODEL` (`"gpt-4.1-mini"`). That disagrees with the flag help ("defaults to the detected/preset model") and with `init`'s `resolve_init_model_name`, which uses `BackendPresetArg::default_model_name()` (e.g. Ollama's `hf.co/google/...` tag, llama-cpp's GGUF path) and refuses presets that have no safe default. `defra-agent onboard --backend-preset ollama` therefore binds a model Ollama does not serve; OpenRouter/vLLM silently get an OpenAI model name instead of requiring an explicit `--model` as init does.

**Suggestion:** Mirror `resolve_init_model_name`: prefer `--model` (reject empty after trim), else `args.backend_preset.and_then(BackendPresetArg::default_model_name)`, else error for presets/URLs with no safe default (or fall back to `DEFAULT_INIT_MODEL_NAME` only where init would). Drop the standalone `gpt-4.1-mini` default unless it is intentionally OpenAI-only and gated on that preset.

crates/defra-agent-cli/src/commands/onboard.rs
```diff
@@ -0,0 +1,371 @@
...
+        bind_default_behavior(&access, &default_behavior_id, &backend_id, Some(&model)).await?;
+        return report(&backend_id, "configured", &default_behavior_id);
+    }
+
+    let configured = query_configured_backends(&access).await?;
+    let detected = detect_local_backend().await;
+    match plan_backend_onboarding(&configured, detected) {
+        // Reusing an existing configured backend: keep its behavior model_name.
+        BackendPlan::AutoConnect { backend_id } => {
+            bind_default_behavior(&access, &default_behavior_id, &backend_id, None).await?;
```
jackzampolin: **[suggestion]** After a normal `defra-agent init`, there is always exactly one enabled `InferenceBackend`, so interactive `onboard` takes `BackendPlan::AutoConnect`, rebinds the existing backend, and reports `"action": "connected"` without probing reachability or offering reconfiguration. `serve`'s new hint tells operators to run `onboard` when the default behavior is not runnable; for the common case (seeded init backend that is unbound/wrong, or an enabled-but-unusable row) this path is effectively a no-op success and never reaches `AdoptDetected` / `OfferLaunchOrRemote`.

**Suggestion:** When AutoConnect would fire, still validate the stored backend (live `/models` probe, or at least surface that nothing changed and print how to reconfigure with flags / disable the old backend). Alternatively, if the default behavior is already bound to that backend, treat onboard as "reconfigure" and offer the launch/remote path when the endpoint is unreachable.

crates/defra-agent-cli/src/commands/onboard.rs
```diff
@@ -0,0 +1,371 @@
...
+                BackendResolutionMode::Init,
+            )?;
+            // detect_local_backend already probed this endpoint's /models, so
+            // record it healthy now — the behavior is runnable without waiting
+            // for the server to re-probe the same live server on next startup.
+            let backend_id =
+                persist_backend(&access, &agent_did, &resolved, &model, HEALTHY_PROBE_STATUS)
+                    .await?;
+            bind_default_behavior(&access, &default_behavior_id, &backend_id, Some(&model)).await?;
+            println!(
```
jackzampolin: **[suggestion]** The AdoptDetected path writes a human message with `println!` to stdout immediately before `report` → `print_json`. Elsewhere (e.g. `serve`) operator chatter goes to stderr so stdout stays a single JSON document. Scripted consumers that pipe `onboard` stdout will see mixed text+JSON on this path.

**Suggestion:** Use `eprintln!` for the human "Detected a local inference server..." line (and other interactive chatter that is not a prompt), keeping `print_json` as the only stdout payload when a machine report is emitted.

crates/defra-agent-cli/src/commands/onboard.rs
```diff
@@ -0,0 +1,371 @@
...
+        let resolved = resolve_backend_config_with_preset(
+            args.backend_preset,
+            args.inference_url.as_deref(),
+            None,
+            None,
+            args.api_key.as_deref(),
+            None,
+            BackendResolutionMode::Init,
+        )?;
+        let model = args
```
jackzampolin: **[suggestion]** Empty `--model` is accepted: `args.model.clone()` is `Some("")`, so the empty string is persisted and bound instead of falling back or erroring. `init` explicitly rejects empty `--model-name`.

**Suggestion:** Normalize with `trim`/`non_empty` before use; `bail!` on whitespace-only `--model`, consistent with init.

crates/defra-agent-cli/src/commands/serve.rs
```diff
@@ -343,6 +343,16 @@ pub(crate) async fn serve(args: ServeArgs) -> Result<()> {
...
         eprintln!("defra-agent server is running local-only. Press Ctrl-C to stop.");
     }
+    // The runtime keeps serving even when the default behavior can't run; make
+    // that visible up front rather than leaving a silent "degraded" that fails
+    // only at chat time. Point at the backend flow (#647) without asserting the
+    // cause is a backend — the reason string carries the actual diagnosis.
+    if let Some(reason) = unavailable_behaviors.get(&default_behavior_id) {
+        eprintln!("The default behavior ({default_behavior_id}) is not runnable: {reason}");
+        eprintln!(
+            "If it needs an inference backend, configure one with `defra-agent onboard`, then restart."
```
jackzampolin: **[suggestion]** The degraded-behavior hint says to run `defra-agent onboard`, then restart. Config writes go through the live server's GraphQL when it is up (`resolve_config_access`), and the runtime control watcher reconciles document changes (including backend/behavior bindings) without a process restart. Requiring restart is often unnecessary and may discourage the correct "onboard while server is up" path.

**Suggestion:** Soften the wording to something like "run `defra-agent onboard` (against this home/server); if the behavior is still unavailable, restart the server so startup probes re-run."

crates/defra-agent-cli/src/prompt.rs
```diff
@@ -0,0 +1,57 @@
+//! Terminal prompt plumbing shared by the interactive `onboard` command and the
+//! `demo` session. One owned [`StdinLines`] reader is threaded through a whole
+//! interactive flow so successive prompts never fight over buffered stdin —
```
jackzampolin, line 3: **[suggestion]** Module docs claim the prompt plumbing is "shared by the interactive `onboard` command and the `demo` session", but demo still uses its own duplicate helpers in `commands/demo/util.rs` (`prompt_line`, `prompt_secret`, `non_empty`, `StdinLines`). Only probe/detect was shared via `onboarding.rs`. The duplication will drift (e.g. stty secret handling) under future fixes.

**Suggestion:** Point demo at `crate::prompt` and delete the duplicated helpers from `demo/util.rs`, or narrow the prompt module docs to "used by onboard" until demo is migrated.

crates/defra-agent-cli/src/cli/args.rs
```diff
@@ -158,6 +158,39 @@ pub(crate) enum Command {
...
+    )]
+    pub(crate) backend_preset: Option<BackendPresetArg>,
+    #[arg(
+        long,
+        help = "Non-interactive: configure this backend base URL, e.g. http://127.0.0.1:8080/v1"
+    )]
+    pub(crate) inference_url: Option<String>,
+    #[arg(
+        long,
+        help = "Model name to bind (defaults to the detected/preset model)"
```
jackzampolin: **[nit]** `--model` help says it "defaults to the detected/preset model", but the non-interactive path ignores preset defaults (Issue 1), and AutoConnect/Select/OfferLaunchOrRemote ignore `--model` entirely except AdoptDetected.

**Suggestion:** Document the real rules: required or preset-default when using backend flags; optional override only on the detect path; ignored when rebinding an existing stored backend.

crates/defra-agent-cli/src/onboarding.rs
```diff
@@ -0,0 +1,195 @@
+//! Shared inference-backend onboarding: local-server detection, live health
+//! probing, and the first-launch decision tree from #647.
+//!
+//! Consumed by the interactive `onboard` command, the `demo` setup, and the
```
jackzampolin: **[nit]** Docs say the shared core is consumed by "the interactive `onboard` command, the `demo` setup, and the `serve` no-backend hint". `serve` only prints a string pointing at onboard; it does not call `plan_backend_onboarding` / `detect_local_backend`. Slightly overstates sharing.

**Suggestion:** Adjust the comment to "consumed by onboard and demo; serve points operators at onboard when degraded."

crates/defra-agent-cli/src/commands/onboard.rs
iverc: Fixed in 9400518e. `resolve_onboard_model` now mirrors `resolve_init_model_name`: explicit `--model` wins (trimmed, rejected if empty), else the preset's `default_model_name()`, else an error for presets with no safe default, else `DEFAULT_INIT_MODEL_NAME` for a bare `--inference-url`. Verified end-to-end: `onboard --backend-preset ollama` binds `hf.co/google/gemma-…` (not gpt-4.1-mini); `--backend-preset openrouter` errors requiring `--model`. Unit tests added for each branch.

crates/defra-agent-cli/src/commands/onboard.rs
iverc: Reworked in 9400518e — and the first attempt at this exposed a real hazard your suggestion's safe reading avoids. AutoConnect now runs a best-effort `/models` probe via `connect_stored_backend`: reachable → recorded healthy (immediately runnable); not reachable → it does NOT silently reconfigure/overwrite (an earlier version did, which could destroy a working authenticated remote backend since the probe is unauthenticated) — it binds the configured backend and prints exactly how to reconfigure (flags, or disable + re-run). So no more no-op success, and no data loss.

crates/defra-agent-cli/src/commands/onboard.rs
iverc: Fixed in 9400518e. All human/prompt output now goes to stderr (the "Detected a local inference server…" line, the reconfigure/warn lines, and `prompt()`/`prompt_secret()` themselves). `print_json` is the only stdout writer, so piping onboard's stdout yields a single JSON document on every report-emitting path.

crates/defra-agent-cli/src/commands/onboard.rs
iverc: Fixed in 9400518e. `validated_model_override` (detect path) and `resolve_onboard_model` (flag path) both trim and `bail!` on a whitespace-only `--model`, matching init. Unit-tested.

crates/defra-agent-cli/src/commands/serve.rs
iverc: Softened in 9400518e. The hint now says to run `onboard` against the running server (writes go through its GraphQL and reconcile live) and only suggests a restart if the behavior stays unavailable (to re-run startup probes).

crates/defra-agent-cli/src/prompt.rs
iverc: Fixed the real way in 9400518e: migrated `demo/util.rs` to `pub(super) use crate::prompt::{…}` and deleted the duplicate `prompt_line`/`prompt_secret`/`set_terminal_echo`/`non_empty`/`StdinLines`; `demo/mod.rs` uses `crate::prompt::stdin_lines()`. One implementation now, so stty/secret handling can't drift.

crates/defra-agent-cli/src/cli/args.rs
iverc: Updated in 9400518e to the real rules: required (or preset-default) with `--backend-preset`/`--inference-url`; an optional override only on the detect path; ignored when rebinding an already-stored backend.

crates/defra-agent-cli/src/onboarding.rs
iverc: Fixed in 9400518e — narrowed to "consumed by onboard and demo; serve only prints a string pointing operators at onboard."

iverc:
Thanks for the thorough review — all eight addressed in 9400518e (each thread replied inline). Summary:

- **Model resolution** now mirrors `init` (`resolve_onboard_model`): preset defaults, empty-`--model` rejected, error for presets with no safe default. Unit-tested + verified end-to-end.
- **Stored-backend connect** is best-effort and **non-destructive**. Worth flagging: my first pass at your AutoConnect suggestion *reconfigured* on a failed probe — but the probe is unauthenticated, so a healthy OpenAI/OpenRouter backend 401s and would have been silently overwritten by a detected local server. An adversarial re-review caught it. It now binds the configured backend and prints how to reconfigure (verified: an openai backend's endpoint survives with a local server up). One deliberate tradeoff: an unreachable *stored* backend is bound-with-warning rather than auto-swapped, to avoid that data-loss path — switching backends is `--inference-url`/`--backend-preset` or disable-and-rerun.
- **stdout purity**: prompts + all chatter to stderr; `print_json` is the sole stdout writer.
- **init parity**: `max_concurrent` now 2 (was 8); added `--api-key-env-var`.
- **demo dedup**: `demo/util.rs` now uses `crate::prompt`; duplicates deleted.
- Docs (args `--model`, `onboarding.rs`, `prompt.rs`) corrected.

Full CLI suite green (574) + 13 unit tests; `cargo check --workspace --all-targets` clean.

PR #655, opened 2026-07-08
Improve sub-agent target config user experience

ChrisBQu:
This PR makes some changes to improve user experience regarding the setting of config tools using subagent-target. Specifically, there are the following changes:

1. Previously, `--subagent-target` only accepted inline JSON. There are now several options available. Firstly, inline JSON still works, so it is backwards compatible. However, the user can now also prefix their input with `@` to signify a source to read from, rather than literal JSON. For examplee, `@some/path.json` would read in from a file, while `@-` would read from stdin. The contents of the source can be a singuularr JSON object, or an array of them.

2. A builder command has also been introduced. See `subagent-target-entry`. This subcommand takes `--name`, `--agent-did`, `--behavior-id`, and `--description` and formats/prints a valid JSON object to stdout. This is intended to provide users a way to easily prepare the input for the aforementioned command. **I would like to discuss how we feel about this idea as a team.**

3. The help text for `--subagent-target` has been updated to document the available forms of input, and point out the existence off the builder subcommand.

4. This PR contains unit tests for the new functionality where possible...

...and a more holistic set of tests were performed manually, by running the following script locally:

```
#!/usr/bin/env bash
# Exercises every --subagent-target input path: inline JSON, @file (single
# object), @file (array), and @- (stdin). Requires a running
# `defra-agent server --home /tmp/subagent-test` (or update the vars below
# to match your home / node).
set -euo pipefail

AGENT_DID=did:key:z6Mkv3ESeh7qdiibpsEVs7WRBQwfgRrgZYwq1ZGkSFy2AfFR
BEHAVIOR_ID=did:key:z6Mkv3ESeh7qdiibpsEVs7WRBQwfgRrgZYwq1ZGkSFy2AfFR:default
SELECTION_ID=did:key:z6Mkv3ESeh7qdiibpsEVs7WRBQwfgRrgZYwq1ZGkSFy2AfFR:default-tools
GRAPHQL=http://127.0.0.1:9191/api/v0/graphql

BIN=./target/debug/defra-agent
WORKDIR=$(mktemp -d)
trap 'rm -rf "$WORKDIR"' EXIT

show() {
  "$BIN" config tools show --graphql "$GRAPHQL" "$SELECTION_ID"
}

# check_names STEP NAME [NAME...]
# Fails unless `show` output contains exactly the given set of names (no
# extras, none missing) -- proves each path both wrote and fully replaced.
check_names() {
  local step="$1"; shift
  local output found
  output=$(show)
  # subagent_targets entries are stored as JSON-encoded strings, so quotes
  # inside them show up escaped (\"name\":\"...\"); strip the escaping
  # before matching so this works regardless of nesting depth.
  found=$(tr -d '\\' <<<"$output" | grep -oE '"name"[[:space:]]*:[[:space:]]*"[^"]*"' \
    | sed -E 's/.*"([^"]*)"$/\1/' | sort | tr '\n' ' '; true)
  local expected
  expected=$(printf '%s\n' "$@" | sort | tr '\n' ' ')
  if [[ "$found" != "$expected" ]]; then
    echo "FAIL [$step]: expected names [$expected] got [$found]"
    echo "$output"
    exit 1
  fi
  echo "PASS [$step]: subagent_targets == [$found]"
}

echo "== path 1: inline JSON via subagent-target-entry =="
ENTRY=$("$BIN" config tools subagent-target-entry \
  --name self-inline \
  --agent-did "$AGENT_DID" \
  --behavior-id "$BEHAVIOR_ID" \
  --description "inline path")
"$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target "$ENTRY" >/dev/null
check_names inline self-inline

echo "== path 2: @file with a single JSON object =="
SINGLE_FILE="$WORKDIR/single.json"
"$BIN" config tools subagent-target-entry \
  --name self-file \
  --agent-did "$AGENT_DID" \
  --behavior-id "$BEHAVIOR_ID" \
  --description "single object from file" >"$SINGLE_FILE"
"$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target "@$SINGLE_FILE" >/dev/null
check_names "file-single" self-file

echo "== path 3: @file with a JSON array of objects =="
ARRAY_FILE="$WORKDIR/array.json"
cat >"$ARRAY_FILE" <<EOF
[
  {"name":"self-array-a","agent_did":"$AGENT_DID","behavior_id":"$BEHAVIOR_ID","description":"array entry a"},
  {"name":"self-array-b","agent_did":"$AGENT_DID","behavior_id":"$BEHAVIOR_ID","description":"array entry b"}
]
EOF
"$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target "@$ARRAY_FILE" >/dev/null
check_names "file-array" self-array-a self-array-b

echo "== path 4: @- reading the same array from stdin =="
"$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target @- <"$ARRAY_FILE" >/dev/null
check_names stdin self-array-a self-array-b

echo "== error path: missing file =="
if "$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target "@$WORKDIR/does-not-exist.json" 2>"$WORKDIR/err.txt"; then
  echo "FAIL [missing-file]: expected a non-zero exit"
  exit 1
fi
grep -q "does-not-exist.json" "$WORKDIR/err.txt" && echo "PASS [missing-file]: error names the missing path"

echo "== error path: malformed JSON =="
echo "not json" >"$WORKDIR/bad.json"
if "$BIN" config tools set --graphql "$GRAPHQL" --agent-did "$AGENT_DID" --selection-id "$SELECTION_ID" \
  --subagent-target "@$WORKDIR/bad.json" 2>"$WORKDIR/err.txt"; then
  echo "FAIL [malformed-json]: expected a non-zero exit"
  exit 1
fi
grep -q "bad.json" "$WORKDIR/err.txt" && echo "PASS [malformed-json]: error names the bad source"

echo
echo "All --subagent-target paths verified."
```

### DIFF
--- a/crates/defra-agent-cli/src/cli/args.rs
+++ b/crates/defra-agent-cli/src/cli/args.rs
@@ -1159,6 +1159,30 @@ pub(crate) enum ToolSelectionCommand {
         alias = "remove"
     )]
     Rm(ConfigShowArgs),
+    #[command(
+        name = "subagent-target-entry",
+        about = "Build a single --subagent-target JSON entry from its parts",
+        after_help = "Example:\n  defra-agent config tools set --graphql <url> --agent-did <did> \\\n    --selection-id main --subagent-target \"$(defra-agent config tools \\\n    subagent-target-entry --name researcher --agent-did did:key:z... \\\n    --behavior-id did:key:z...:default)\""
+    )]
+    SubagentTargetEntry(SubagentTargetEntryArgs),
+}
+
+#[derive(Debug, clap::Args)]
+pub(crate) struct SubagentTargetEntryArgs {
+    #[arg(long, help = "Model-facing name used by spawn_subagent")]
+    pub(crate) name: String,
+    #[arg(
+        long,
+        help = "DID of the agent that owns the target behavior (local or remote)"
+    )]
+    pub(crate) agent_did: String,
+    #[arg(long, help = "Behavior id on the owning agent")]
+    pub(crate) behavior_id: String,
+    #[arg(
+        long,
+        help = "Optional human-readable description surfaced to the model"
+    )]
+    pub(crate) description: Option<String>,
 }
 
 #[derive(Subcommand)]
@@ -1467,7 +1491,11 @@ pub(crate) struct ToolSelectionUpsertArgs {
     pub(crate) clear_defra_query_collections: bool,
     #[arg(
         long = "subagent-target",
-        help = "SubagentTarget JSON entry allowed for spawn_subagent (repeatable); omit to preserve existing targets"
+        help = "SubagentTarget JSON entry allowed for spawn_subagent, e.g. \
+                {\"name\":\"researcher\",\"agent_did\":\"did:key:...\",\"behavior_id\":\"did:key:...:default\",\"description\":\"...\"} \
+                (repeatable); or @path/@- to read one entry or a JSON array of entries from a \
+                file/stdin; omit to preserve existing targets. See `config tools \
+                subagent-target-entry --help` to build a single entry from its parts."
     )]
     pub(crate) subagent_targets: Vec<String>,
     #[arg(
--- a/crates/defra-agent-cli/src/commands/config/mod.rs
+++ b/crates/defra-agent-cli/src/commands/config/mod.rs
@@ -45,6 +45,9 @@ pub(crate) async fn dispatch(command: ConfigCommand) -> Result<()> {
             ToolSelectionCommand::Rm(args) => {
                 crud::config_rm(crud::TOOL_SELECTION_SPEC, args).await
             }
+            ToolSelectionCommand::SubagentTargetEntry(args) => {
+                tools::subagent_target_entry_command(args)
+            }
         },
         ConfigCommand::Profile { command } => match command {
             InferenceProfileCommand::Set(args) => profile::inference_profile_set(args).await,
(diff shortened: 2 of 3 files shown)


jackzampolin (approved):
Really solid CLI UX improvement — approve.

The pain this fixes is real: stuffing SubagentTarget JSON onto `--subagent-target` was hostile (quoting, multi-entry setup, no structure help). Making the flag accept `@path` / `@-` with object-or-array expansion is the right move, familiar, and backwards-compatible with inline JSON. Scope is disciplined: pure CLI plumbing, no schema/runtime/Lean changes.

On the open design question: **keep `subagent-target-entry`.** Typed flags beat hand-typed parse errors, it composes cleanly with `$()`, and reusing `defra_agent::subagent_target_entry` keeps one construction path. Surface area is small enough relative to the ergonomics win, especially while people are still learning the document shape.

Thanks for the unit coverage on the pure expand/parse path and the manual matrix of input forms. Ship it.

Issue #661, opened 2026-07-08
v0.6.5: Amy hub stalls after AgentResponse persistence failure; replica processing divergence


jackzampolin:
## Summary

During Amygdala #630 stage-1 P2P rollout, enabling two same-host steward→Amy `subagent-host` legs (`coding-steward` and `x-data-steward`) stayed stable for the initial watch window, but the first Amy spawn/readback smoke request failed before any subagent/tool call and the hub then became unresponsive.

The failure happens while Amy creates the initial streaming `AgentResponse`:

```text
request_id=f58f3198-e765-4d72-9d20-d4476c6fe3e5
session_id=stage1-p2p-20260708T192713Z-2a4951c8-6346-424d-befb-7084e6003a9f
behavior_id=default
error=creating AgentResponse failed: begin_streaming_response failed: [QueryResponseError { message: "Expected array, got: Number(0)", path: None, locations: None }]
```

No `AgentToolCall` rows were created and `defra-agent subagent list --root f58f...` showed no children, so this did not reach the actual subagent dispatch path.

## Environment

- defra-agent v0.6.5 fleet-wide except known excluded hetzner-vps 0.6.4
- Amy hub: studio-1 `amy`, HTTP `:9191`, P2P `:9192`
- Batch enabled: same-host host legs only
  - `infra/agents/studio-1/coding-steward/peer-pairings/amy/object.json`
  - `infra/agents/studio-1/x-data-steward/peer-pairings/amy/object.json`
- Enable commit in Amygdala repo: `bd4763a2 p2p(#630): stage-1 enable coding/x-data host legs`
- Rollback commit in Amygdala repo: `6e16ba87 p2p(#630): rollback stage-1 host legs`

## Timeline / observations

1. Baseline before enabling:
   - #639 resume empty-stream query: `[]`
   - QUIC/restart/resume-path warning/error query: `[]`
   - studio-1 CPU: ~20.7%
   - no `defra_agent_*pending*`, `*dag*`, `*p2p*`, or `*quic*` metric exported; only backend/probe metrics matched.

2. Enabled and applied only two same-host host legs:
   - `coding-steward` apply wrote pairing doc `bae-85717128-c4bf-56d1-8720-65245df4d261`
   - `x-data-steward` apply wrote pairing doc `bae-75f13181-a014-5635-9c54-d43da133fea2`
   - per-root validation and `validate-p2p-topology` passed.
   - Amy coordinator-side diff was all `=ok`.

[...]

PR #663, opened 2026-07-08
fix(migration): scalar-Int fields patched as IntArray (Kind 5→4) + type guard + terminal-failure surfacing (#661)

jackzampolin:
## Summary

Fixes the #661 root cause and two robustness gaps it exposed. During the amygdala fleet's v0.6.5 rollout, **every `create_AgentResponse` failed** with `Expected array, got: Number(0)` and the hub then wedged.

## Root cause: migration patches typed scalar-Int fields as `IntArray`

In defradb.rs's numeric field-kind encoding, **scalar `Int` is `Kind: 4`; `[Int]` (IntArray) is `Kind: 5`**. Six migration patches used **`Kind: 5`** for fields that are scalar `Int` in the SDL:

| field | collection |
|---|---|
| `reasoning_progress_seq` | AgentResponse |
| `subagent_depth` | AgentRequest |
| `cross_deployment_spawn_timeout_seconds` | ToolSelection |
| `tool_count` | ToolServiceHealthState |
| `behavior_executor_capacity` | AgentRuntime |
| `behavior_executor_queue_depth` | AgentRuntime |

Any node that adds one of these **via the migration patch** gets an array column; a scalar write then fails with `Expected array, got: Number(0)`. `reasoning_progress_seq` is new in v0.6.5 and added *only* by patch, so it hit every node on the first request; the other five are latent landmines (nodes that created the field from the base SDL got the correct scalar Int, masking the bug).

**Fix:** all six patches → `Kind: 4` (scalar Int), matching the SDL. A test that previously asserted `Kind 5` (encoding the bug) is corrected. Added a comment documenting the numeric-Kind scalar/array trap next to the existing kind-number guidance.

Fenced by `reasoning_progress_migration_adds_scalar_int_not_array`, which runs the migration and then does the exact incident operation — a scalar `create_AgentResponse { reasoning_progress_seq: 0 }` — and asserts it succeeds (with the old `Kind:5` it fails with `Expected array, got: Number(0)`).

## Robustness fix 1: migrations guard field TYPE, not just presence

`ensure_agent_response_reasoning_progress_migration` guarded on `collection_has_field(...)` (presence only), so a field that already existed with the *wrong* kind was silently accepted and never repaired. It now compares the existing field's kind against a stable scalar-Int sibling and **fails loudly at startup** with an actionable message, rather than running broken. This turns "silent, catastrophic-on-first-request" into an obvious boot error, and would have caught this class immediately (and catches out-of-band mistyped fields, e.g. from a hand-applied schema patch). Fenced by `reasoning_progress_migration_fails_loud_on_wrong_field_kind`.

## Robustness fix 2: terminal-failure transitions no longer swallow errors

The daemon's four request-failure arms did `let _ = lifecycle.record_failure_reason(...); let _ = lifecycle.fail();` — discarding the results. A request whose terminal transition can't be persisted stays claimable and gets re-processed, invisibly. Extracted into `finalize_request_failure`, which surfaces both errors at `ERROR`. (These writes target the `AgentRequest` doc, independent of the failing `AgentResponse` write — so a node already reaches terminal state on its own doc; this just makes a failed transition loud.)

## Scope notes

- The **hub wedge itself** (CPU→98% / unresponsive) is a deeper **P2P re-claim / convergence** issue: with the pairing legs briefly enabled, the failed request replicated to co-located peers whose terminal state didn't converge (stale `processing`/`claimed` copies). That's a separate investigation, not force-fit here.
- The fleet was repaired out-of-band (fields reset to scalar Int); this PR fixes the source so re-adds and fresh nodes are correct, and the guard makes any residual mismatch fail loudly.

## Validation

`cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo test -p defra-agent` (full gate). Root-cause + guard fences both verified red-then-green.

Refs #661.

### DIFF
--- a/crates/defra-agent/src/agent/daemon.rs
+++ b/crates/defra-agent/src/agent/daemon.rs
@@ -23,6 +23,33 @@ use crate::runtime_trace::{
 use crate::streaming::DefraStreamWriter;
 use crate::watcher::AgentRequest;
 
+/// Drive a request to its durable terminal `failed` state, surfacing — not
+/// swallowing — persistence errors. These writes go to the `AgentRequest` doc
+/// (independent of any `AgentResponse` write, which may itself be failing). A
+/// request whose terminal transition can't be persisted stays claimable and
+/// gets re-processed, so a failure here must be loud rather than a silent
+/// `let _ =` (#661).
+async fn finalize_request_failure(
+    lifecycle: &mut RequestLifecycle,
+    reason: &str,
+    request_id: &str,
+) {
+    if let Err(error) = lifecycle.record_failure_reason(reason).await {
+        tracing::error!(
+            request_id,
+            error = %error,
+            "failed to persist request failure_reason; request may remain non-terminal"
+        );
+    }
+    if let Err(error) = lifecycle.fail().await {
+        tracing::error!(
+            request_id,
+            error = %error,
+            "failed to transition request to failed; request may be re-claimed and re-processed"
+        );
+    }
+}
+
 pub(super) struct BehaviorDaemon<M: CompletionModel> {
     node: Arc<defra_node::EmbeddedNode>,
     behavior: Arc<AgentBehavior>,
@@ -280,8 +307,8 @@ impl<M: CompletionModel + 'static> BehaviorDaemon<M> {
                     requested_behavior_id = %requested_behavior_id,
                     "rejecting request for unroutable behavior"
                 );
-                let _ = lifecycle.record_failure_reason(&error.to_string()).await;
-                let _ = lifecycle.fail().await;
+                finalize_request_failure(&mut lifecycle, &error.to_string(), &request.request_id)
+                    .await;
                 let response_exists = lifecycle.response_exists().await.unwrap_or(false);
                 let response_written = if response_exists {
                     self.stream_writer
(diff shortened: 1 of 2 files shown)


jackzampolin:
## P2P hub-wedge investigation — conclusion (multi-agent, adversarially verified)

Ran a deep investigation of the #661 hub wedge against the evidence bundles + code + proof surface to decide whether the P2P re-claim/convergence fix belongs in this PR. **It does not — and this PR as-is is the correct pre-test merge.**

**Mechanism: liveness/convergence gap, NOT a safety violation.** No peer ever claimed or processed the owner's replicated request — the watcher filters `agent_did == self` (`watcher/query.rs:36,:72`); every peer replica carried the owner's DID with `claimed_by=null`; both enabled legs reported `ignored_foreign_processing_count=2` (read-only, no mutation). The only divergence: the owner's terminal `failed` delta never converged to the replicas, and nothing sweeps a stale *foreign* row on a live peer (recovery is owner-scoped + startup-only).

**This PR removes the trigger for the observed divergence.** The schema fix makes `create_AgentResponse` succeed → the request reaches `completed` → that terminal delta replicates normally → peers converge. The specific stuck-state signature (owner=failed, replicas=processing) cannot recur from this cause.

**The real convergence fix is model-first and NOT small/safe for this PR** — filed as #664 (multi-replica request-state model: SAFETY `SingleClaimer` + LIVENESS `TerminalConverges`, then a convergence mechanism). Two residual items also live there: the ~6-min **unlogged wedge window** (hub emitted nothing under load — the CPU-wedge cause is genuinely undetermined) and a latent `finalize_request_failure` terminal-write recurrence path.

**Re-test gate:** stage-1 can proceed on this merge, but instrument the wedge window (hub must keep logging under P2P/gossip/HTTP degradation) and watch peer convergence / `ignored_foreign_processing_count` during the run — this is *monitor-and-instrument*, not *monitor-only*.

Follow-up: #664.

PR #665, opened 2026-07-08
feat(p2p): auto-install reciprocal conversation replicators for dapair-invited peers (server side)

noahsaso:
## Problem

When a phone pairs with a `defra-agent` server via a `dapair1-` conversation invite, the pairing is one-directional: the phone pushes its conversation docs to the server, but nothing pushes the server's `AgentResponse`/`AgentMessage` docs back. Today that requires a manual `defra-agent p2p admin replicators add` (8 conversation collections, filtered by the server's DID, targeting the phone).

A server-only fix is impossible: the defradb adapter only learns dialable shareable addresses on **outbound** dials, so the server cannot recover an address for a peer that dialed in. The phone must publish a signed `PeerEndpoint`; this PR is the server-side half that consumes it.

## Design

Two missing inputs are supplied document-driven, and the existing (proven) Layer-2 pairing engine does the rest:

1. **Intent** — `p2p pairings invite --template conversation` records an unsigned, server-local `ReciprocalConversationIntent { member_did }` row. Only conversation invites create one; `network-control` (fleet) invites never do, so the fleet no-crosswise invariant is preserved structurally.
2. **Address** — a new background reconciler joins intent DIDs with signature-verified, fresh `PeerEndpoint` rows and derives `source="reciprocal"` `DataPlanePairingDesired` rows (peer id + address from the endpoint, `agent_did` = self, `template` = conversation). Intent removal or endpoint disappearance retracts exactly the derived row.
3. **Materialization** — the Layer-2 gate now accepts `network ∪ reciprocal` endpoints, so a conversation-paired phone doesn't need fleet `NetworkMembership`. The engine's existing safeguards are unchanged: `agent_did` is forced to self, and the replicator address always comes from the live signed endpoint.

### Trust basis

The intent row is **unsigned**; its safety rests on the standing control-plane assumption that these collections are only ever written locally (the same assumption `PeerPairingDesired`/`DataPlanePairingDesired` already rely on). This is now documented on the schema, with the escalation path (server-signed intents) if inbound collection admission ever widens. The `PeerEndpoint` itself is signature-verified against the member DID and freshness-checked before it can materialize anything.

### Ownership

`DataPlanePairingDesired` gains a `source` column. The reciprocal reconciler only creates/refreshes/deletes rows with `source="reciprocal"`; peers owned by any other source (operator, legacy null) are excluded from its desired set up front (mirroring the network reconciler's blocked-set pattern), with a per-upsert guard kept as a race backstop. Operator writers now stamp `source="operator"`, so an operator taking over a peer is explicit and the reciprocal reconciler backs off permanently. Modeled in Lean as partition preservation.

### Mobile staleness window

Reciprocal endpoints use a dedicated staleness window (`DEFRA_AGENT_RECIPROCAL_STALE_MS`, default 24 h) instead of the fleet 3×-heartbeat (90 s) window. Phones stop heartbeating the moment the app backgrounds; with the fleet window every background/foreground cycle would retract the row, uninstall the replicator, and replay history on re-add. A stale-but-intended endpoint only costs replicator retries; deliberate revocation is intent removal, which retracts immediately.

## Foundation flow

- **Lean** (`proofs/Proofs/PeerRegistryDiscovery/ReciprocalConversation.lean`): models the intent × endpoint → data-plane derivation. Proves the membership characterization (`mem_deriveReciprocal`), idempotence/convergence (settled = full-row equality with the derived set), ownership safety (operator/network/registry partitions preserved), and retraction soundness for both intent removal and endpoint staling. Zero `sorry`.
- **Conformance** (`tests/conformance/reciprocal_conversation.rs`): 5 tests pinning the Rust tick to the model — join semantics, idempotence, ownership safety (including a blocked operator-owned peer), row-content convergence + write-free fixpoint, and retraction soundness.
- **Rust**: `reciprocal.rs` (pure derivation, `ReciprocalStore` trait, GraphQL store with endpoint verification, tick, reconciler loop), Layer-2 gate extension with per-sweep caching in `engine.rs`, runtime spawn in `startup.rs`, additive schema + idempotent migrations for both fresh and upgraded stores.

## Scope

Server-side contract only. Follow-ups:

- **#666** — scan-one-QR pairing (claim-at-join bearer invite). Builds on this PR unchanged; moves intent creation from mint time to claim time.
- **jackzampolin/amy#3** — client half: phone publishes the signed `PeerEndpoint` (needs sourcenetwork/defradb.rs#1098 for the shareable-address FFI), then the end-to-end two-node proof test lands here.

## Verification

- `lake build`: zero `sorry`.
- `cargo fmt --all --check`, `cargo check --workspace --all-targets`: clean.
- Unit: derivation join/defer/blank handling, tick upsert/refresh/retract/quiescence/blocked-peer, ownership classification, mutation shape (source-scoped filter, `null` not `[]`, conversation template), interval defaults/overrides.
- Conformance: 5 Lean-driven tests (above).
- Integration: two-agent CLI test asserting the `ReciprocalConversationIntent` row after a conversation invite (and none for `network-control`).


noahsaso:
Follow-up work building on this PR:

- **#666** — scan-one-QR mobile pairing via a claim-at-join bearer invite. Builds on this PR's reciprocal engine unchanged (reconciler, Layer-2 gate, ownership partitioning, staleness/retraction); the only seam that moves is the `ReciprocalConversationIntent` creation site, from invite-mint time to claim time — which also tightens the lifecycle wart where a never-consumed invite leaves a standing intent.
- Client half is tracked in jackzampolin/amy#3; the phone's dialable-address publishing needs sourcenetwork/defradb.rs#1098 (`p2p_shareable_address` FFI, closes sourcenetwork/defradb.rs#1097).

jackzampolin:
## Review notes (stack with #668)

This is in great shape — foundation flow is doing what we want: Lean (`ReciprocalConversation`) → conformance that names the theorems → pure derive/tick → GraphQL seams with escape / `null`-not-`[]`, and ownership partitions that hold under tests. The 24h phone-aware stale window and the fleet `source=operator` stamp are good product instincts.

**Merge this before #668** (bearer claim depends on the intent collection + reciprocal reconciler).

### Follow-ups OK after merge (non-blocking)

- **Revoke ≠ retract:** deleting/revoking membership does not clear `ReciprocalConversationIntent`, so a reverse conversation edge can keep living until intent delete or the 24h endpoint stale window. Document that revocation = delete intent, or cascade later.
- **v5 intent-at-mint wart:** conversation invite still records intent at mint time (standing intent if never joined). #668 moves intent to claim time for the bearer path; tightening v5 the same way is a natural follow-up once claim exists.
- Full reverse data-plane e2e (endpoint heartbeat → reciprocal desired → installed reverse replicator) still waits on client half / harness intervals — fine for this PR.

Client half (QR scan / signed `PeerEndpoint` publishing) stays out of this repo for now — separate work.

Thanks for keeping the proofs and fences tight.

noahsaso:
Thanks — agreed on merge order (#665 → #668). Follow-ups from your notes are now tracked: revoke ≠ retract is **#672** (with the interim operator contract: revocation = delete the `ReciprocalConversationIntent` row, which retracts the derived edge on the next sweep per the Lean retraction-soundness theorem); the v5 intent-at-mint tightening is a work item on #666 and is unblocked now that #668 provides the claim step; the full reverse data-plane e2e rides the client half (jackzampolin/amy#3).

PR #669, opened 2026-07-09
Fix llama.cpp/llama-server model discovery parsing

noahsaso:
## Purpose

Local OpenAI-compatible backends such as llama.cpp/llama-server were being marked unhealthy during model discovery. Their `/v1/models` response returns an OpenAI-standard `data[].id` plus a top-level `models[]` array whose entries carry `name`/`model` but **no `slug`**. The discovery parser required `slug` on every `models[]` entry, so serde rejected the whole response — even though the valid `data[].id` was right there. This blocked mobile e2e because failed discovery marks a running local llama backend unhealthy.

## Changes

- `ChatGptCodexModelRecord` now treats `slug` as optional and adds optional `id`, `name`, and `model` fields, so a non-Codex `models[]` entry can no longer fail deserialization of an otherwise valid response.
- New `identifier()` helper selects the first non-empty of `slug`/`id`/`name`/`model`; entries with no usable identifier are ignored.
- Discovered model IDs are deduplicated while preserving response order, so a llama response that repeats the same id in `data[].id` and `models[].name`/`model` returns a single model.
- Added targeted tests for the observed llama.cpp/llama-server response shape and for common `models[]` `id`/`name`/`model` variants. Existing OpenAI `data[].id` and Codex `slug` coverage is unchanged.

## Verification

- `cargo fmt --check`: clean.
- `cargo test -p defra-agent backend_provider::tests`: 5 passed (including the two new llama-shape tests), 0 failed.

### DIFF
--- a/crates/defra-agent/src/backend_provider.rs
+++ b/crates/defra-agent/src/backend_provider.rs
@@ -105,7 +105,23 @@ struct OpenAiModelRecord {
 
 #[derive(Deserialize)]
 struct ChatGptCodexModelRecord {
-    slug: String,
+    /// ChatGPT Codex returns `slug`; llama.cpp/llama-server and other OpenAI-compatible
+    /// servers commonly return `id`, `name`, or `model` under a top-level `models` array.
+    /// All fields are optional so a non-Codex `models` entry cannot reject an otherwise valid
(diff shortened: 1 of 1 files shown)


jackzampolin:
ezpz thx

PR #686, opened 2026-07-10
Fix active version migration bug

ChrisBQu:
In the `ensure_subagent_extensions_migrations` function inside `migration.rs` see that v3 through v9 blocks all concluude by setting the `active_version` value to the new version ID. However, this line was missing for v10. This PR simply adds the missing line.

Note: A previous version of this PR was much larger in scope, but the issues it addressed have been separately, and independently resolved.

### DIFF
--- a/crates/defra-agent/src/migration.rs
+++ b/crates/defra-agent/src/migration.rs
@@ -802,6 +802,7 @@ pub async fn ensure_subagent_extensions_migrations(node: Arc<EmbeddedNode>) -> R
                 v10 = %v10.version_id,
                 "ToolSelection patched with context budget flag"
             );
+            active_version = v10;
         }
     } else {
         tracing::debug!("ToolSelection collection absent; subagent patch no-op");


jackzampolin:
Chris — good find, and the diagnosis is exactly right: `ensure_subagent_extensions_migrations` was never wired into startup, which is the root cause behind this week's boot crashes. You independently landed on the same conclusion as #682, which merged about an hour after you opened this — it wires the whole orphaned chain into `ensure_all_runtime_migrations` and adds a schema-driven sweep (`ensure_bundled_schema_fields`) that additively patches any field the bundled SDL has and the live collection doesn't. That's why this branch now conflicts on `migration.rs`, and it also answers your 17-fields question by construction: the bundled schema is the migration source of truth now, so no per-field patch decisions needed (`read_only_command_allowlist` is covered by test there).

Two things from this PR are still worth keeping:

1. **The missing `active_version = v10;` fix** — main still doesn't have it. It's latent today (v10 is the last hand-written ToolSelection patch) but it's a trap for whoever appends the next one. Worth landing as a one-liner.
2. **The typed `ToolSelectionQueryError` + honest error hint** — optional now that the sweep self-heals this class, but the sweep still warns-and-skips unmapped kinds (relations etc.), so the failure mode isn't fully extinct. If we keep it, it probably wants generalizing beyond ToolSelection rather than landing as-is.

Suggest rebasing onto main and shrinking this to (1), with (2) as a follow-up discussion if you think it's worth it. Also FYI the red `rust-and-cli` here looks like a flake, not your change — the only delta from your green head is a doc comment, and the failure is `cli_status::status_and_metrics_surface_overridden_p2p_admission_knobs` dying in `init`.

ChrisBQu:
> Chris — good find, and the diagnosis is exactly right: `ensure_subagent_extensions_migrations` was never wired into startup, which is the root cause behind this week's boot crashes. You independently landed on the same conclusion as #682, which merged about an hour after you opened this — it wires the whole orphaned chain into `ensure_all_runtime_migrations` and adds a schema-driven sweep (`ensure_bundled_schema_fields`) that additively patches any field the bundled SDL has and the live collection doesn't. That's why this branch now conflicts on `migration.rs`, and it also answers your 17-fields question by construction: the bundled schema is the migration source of truth now, so no per-field patch decisions needed (`read_only_command_allowlist` is covered by test there).
> 
> Two things from this PR are still worth keeping:
> 
> 1. **The missing `active_version = v10;` fix** — main still doesn't have it. It's latent today (v10 is the last hand-written ToolSelection patch) but it's a trap for whoever appends the next one. Worth landing as a one-liner.
> 2. **The typed `ToolSelectionQueryError` + honest error hint** — optional now that the sweep self-heals this class, but the sweep still warns-and-skips unmapped kinds (relations etc.), so the failure mode isn't fully extinct. If we keep it, it probably wants generalizing beyond ToolSelection rather than landing as-is.
> 
> Suggest rebasing onto main and shrinking this to (1), with (2) as a follow-up discussion if you think it's worth it. Also FYI the red `rust-and-cli` here looks like a flake, not your change — the only delta from your green head is a doc comment, and the failure is `cli_status::status_and_metrics_surface_overridden_p2p_admission_knobs` dying in `init`.

I have reworked this PR as suggested. It's now the one-line fix that is described. Let's separate out number two, because I think it warrants discussion first, and is conceptually a different thing.

iverc (approved):

PR #694, opened 2026-07-11
Fix startup recovery on duplicate-session stores; count only real recoveries (#693)

### LINKED ISSUE #693: Startup conversation recovery: fails on duplicate session_id docs and reports failed upserts as recovered
## Summary

Startup conversation recovery has two defects, observed on a real host during
the v0.6.11 rollout (host aborted and restored per criteria):

1. **Fails on duplicate matching documents.** Recovery runs
   `upsert_AgentConversation(filter: { session_id: { _eq: "<id>" } }, …)`; on a
   store containing two `AgentConversation` docs with the same `session_id`,
   DefraDB (correctly) refuses with `cannot upsert multiple matching documents`
   and the recovery mutation fails.
2. **Reports the failures as successes.** Immediately after two such failed
   upserts, the runtime logged
   `startup: recovered stuck conversations … count=2` — the count tallies
   attempts, not successful recoveries. A completely failed recovery pass
   reports healthy.

## Evidence (v0.6.11 / 4f54a137 first boot on an upgraded v0.6.5 home)

```
ERROR query::runner::executor: Query execution failed query=mutation {
    upsert_AgentConversation(
        filter: { session_id: { _eq: "b2de8338-ffbf-4b16-920c-affc99d82785" } },
        add: { … }
    ) { _docID }
} error=cannot upsert multiple matching documents
[second identical failure for another session_id]
INFO defra_agent::agent::runtime::startup: recovered stuck conversations agent_did=… count=2
```

Fleet prevalence check (GraphQL, same day): the affected host has duplicated
`session_id`s in `AgentConversation`; two other production stores surveyed
(129 and 125 conversations) have zero — so the duplicate condition is
store-history-dependent (this host lived through a schema-broken period), but
any store that ever minted duplicates will brick recovery on upgrade, and the
false-success log (defect 2) is latent everywhere.

## Asks

1. Recovery must tolerate duplicate matches: deterministically pick the
   canonical doc (e.g. richest/latest) and recover it — ideally reconciling or
   flagging the duplicates (a startup sweep already exists for schema shape;
   duplicate-doc healing is its data-shape sibling). At minimum: recover the
   canonical doc, warn per duplicate set with the session_id.
2. `recovered stuck conversations count=` must count successful recoveries
   only; failures logged distinctly (they are the opposite of recovery).
3. A regression test with a seeded duplicate-session_id store.

## Context

Found during the fleet v0.6.11 rollout (spokes-first). The migration sweep
(#682/#688/#691) healed the host's `[Int]` schema wound correctly — this
recovery defect is the remaining blocker for that host. Rollout continues on
hosts that pass a duplicate-conversation pre-check; the affected host is held
on v0.6.5 awaiting this fix.

jackzampolin:
Fixes #693 — both defects observed on the v0.6.11 rollout (strangenas, studio-1 web-research/inference, amygdalabook):

1. **Duplicate `session_id` docs bricked conversation recovery.** `upsert_AgentConversation(filter: {session_id: …})` multi-matches on a duplicate-carrying store and DefraDB refuses it (`cannot upsert multiple matching documents`), so the conversation was never recovered.
2. **Failed attempts were logged as successes.** `recovered stuck conversations … count=` tallied attempts (`rows.len()`), so a fully-failed pass reported healthy.

## Fix (query-first, per the issue's preferred option)

- **`session/query.rs::load_conversation_document`** — reads every doc matching the session and deterministically picks the canonical one: **latest `updated_at`, `_docID` tie-break** (an unparseable timestamp sorts oldest, so a malformed twin never outranks a well-formed one). When duplicates exist it WARNs with the `session_id`, `duplicate_count`, and `canonical_doc_id`. Duplicates are **flagged, never deleted** (v1: operators wipe or a later sweep reconciles).
- **`session/conversation.rs`** — the two upsert helpers now address an existing document by `_docID` (`update_AgentConversation`), making multi-match structurally impossible. The upsert survives only for the doc-doesn't-exist case, where its atomicity still closes the concurrent-create race. This also heals the **live request path** on a duped store, which routed through the same upsert and would have failed on every turn of the wounded session.
- **`lifecycle/recovery.rs::recover_stuck_conversations`** — a duplicate-carrying store yields one stuck row per twin, so the pass now drives **one attempt per session_id** (canonical stuck row by the same recency rule) and returns `(recovered, failed)`: the recovered count reflects only writes that succeeded.
- **`startup.rs`** — `conversations_failed > 0` gets its own distinct WARN (`startup conversation recovery attempts failed`); the count in the INFO line is successes only.

**Sibling audit** (issue ask 2 / prompt step 3): request, response, tool-call, and inference-call recovery counters already increment only on success with per-failure WARNs — conversations was the lone `rows.len()` offender. No change needed there.

## Regression tests (TDD — both watched red on the exact production failure first)

Seeding a duplicate through the mutation path is impossible on a fresh store (the unique `session_id` index rejects it — production twins arrived via P2P merge, which enforces no index; indexes are also immutable to patches). The new `test_db_with_duplicate_capable_conversations` support therefore creates the collection from an **era-faithful SDL** (non-unique `session_id` index) before normal bootstrap — the same shape the affected stores actually lived through — and the recovery semantics under test (upsert multi-match refusal, `_docID`-keyed update) are index-independent.

- `recover_all_recovers_canonical_conversation_when_session_id_is_duplicated` — two twins, one session: canonical (newest) doc recovered to `active`, stale twin untouched, `conversations_recovered == 1`, WARN names the session_id and `duplicate_count=2` (asserted via a thread-scoped tracing capture). **Red before fix: count=2 with zero successful writes.**
- `recover_all_counts_only_successful_conversation_recoveries` — failure injected via a behavior_id conflict between the stuck doc and its newer twin: `conversations_recovered == 0`, `conversations_failed == 1`. **Red before fix: count=1.**

No Lean change: the Recovery model doesn't cover conversation-status repair, and this changes no legal transitions — it's write-path robustness plus counter honesty.

## Gates

- `cargo test -p defra-agent --lib --tests` — green (exit 0, full suite)
- `cargo check --workspace --all-targets` — green
- `cargo fmt --all --check` — green
- clippy: all changed files clean under local clippy 1.95; remaining findings are pre-existing new-lint noise in untouched crates (CI runs no clippy step)

### DIFF
--- a/crates/defra-agent/src/agent/runtime/startup.rs
+++ b/crates/defra-agent/src/agent/runtime/startup.rs
@@ -601,6 +601,17 @@ async fn log_recovery(node: &defra_node::EmbeddedNode, agent_did: &str, default_
                     "recovered stuck conversations"
                 );
             }
+            if report.conversations_failed > 0 {
+                // Failed attempts found stuck documents, so the pass was not
+                // a no-op — but they are the opposite of a recovery and must
+                // never inflate the recovered count (#693).
+                recovered_any = true;
+                tracing::warn!(
+                    agent_did = %agent_did,
+                    count = report.conversations_failed,
+                    "startup conversation recovery attempts failed"
+                );
+            }
         }
         Err(error) => {
             tracing::warn!(agent_did = %agent_did, error = %error, "startup recovery failed");
--- a/crates/defra-agent/src/lifecycle.rs
+++ b/crates/defra-agent/src/lifecycle.rs
@@ -243,6 +243,9 @@ pub struct RecoveryReport {
     pub requests_recovered: usize,
     pub responses_recovered: usize,
     pub conversations_recovered: usize,
+    /// Conversations whose recovery attempt failed. Failures are the opposite
+    /// of a recovery and must never fold into the recovered count (#693).
+    pub conversations_failed: usize,
 }
 
 /// Outcome of one durable request-terminal repair pass. A terminal response
--- a/crates/defra-agent/src/lifecycle/recovery.rs
+++ b/crates/defra-agent/src/lifecycle/recovery.rs
@@ -16,12 +16,14 @@ impl RequestLifecycle {
         let requests_recovered = Self::repair_terminal_requests(node, agent_did)
             .await?
             .repaired;
-        let conversations_recovered = recover_stuck_conversations(node, agent_did).await?;
+        let (conversations_recovered, conversations_failed) =
+            recover_stuck_conversations(node, agent_did).await?;
 
         Ok(RecoveryReport {
             responses_recovered,
             requests_recovered,
             conversations_recovered,
+            conversations_failed,
         })
     }
 
@@ -546,7 +548,13 @@ async fn recover_missing_response_documents(node: &EmbeddedNode, agent_did: &str
     Ok(recovered)
 }
 
-async fn recover_stuck_conversations(node: &EmbeddedNode, agent_did: &str) -> Result<usize> {
+/// Returns `(recovered, failed)`. The recovered count reflects only writes
+/// that actually succeeded — a failed attempt is the opposite of a recovery
+/// and gets its own WARN instead (#693).
+async fn recover_stuck_conversations(
+    node: &EmbeddedNode,
+    agent_did: &str,
+) -> Result<(usize, usize)> {
     let escaped_agent_did = escape_graphql_string(agent_did);
     let query = format!(
         r#"{{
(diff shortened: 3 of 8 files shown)


jackzampolin:
The minting mechanism this PR defends against (P2P merge bypassing unique indexes, + index constraints being unpatchable after the fact) is now tracked upstream as sourcenetwork/defradb.rs#1111 — this fix is the correct app-side posture regardless of which merge semantic upstream picks.

PR #698, opened 2026-07-12
Duplicate-tolerant conversation recovery that counts successes, not attempts (#693)

### LINKED ISSUE #693: Startup conversation recovery: fails on duplicate session_id docs and reports failed upserts as recovered

iverc:
Fixes #693.

## Both defects, and a third the issue didn't cover

**1. Recovery failed on duplicate `session_id` docs.** `upsert_AgentConversation(filter: { session_id })` matches every duplicate, and DefraDB refuses a multi-match upsert.

**2. Failed upserts were reported as recoveries.** `recover_stuck_conversations` returned `rows.len()` — attempts, not successes — so a fully failed pass logged `recovered stuck conversations count=2` and read as healthy.

**3. (New) The live request path was broken on the same stores.** The identical session_id-filtered upsert backs `prepare_session_with_identity` (every claimed request) and the status/title updates. A duplicate-carrying host wasn't merely un-bootable — ordinary request handling failed too, and the filtered `update_` variants silently wrote *both* duplicates. Fixing only the recovery sweep would have left the held hosts degraded.

## The formal gap that let defect 2 ship

The Lean sweep contract models `recover : Row → Row` as **total**, and proves `recoveredRows_length : recovered.length = rows.length`. "Recovered count == attempted count" was the specification — so `rows.len()` was a *faithful* implementation of it, and no fence could have caught the bug.

`Proofs/Recovery/Outcome.lean` adds the missing layer: a per-row outcome, a `{recovered, failed}` report, and the theorems that make honest accounting checkable — `report_accounts_for_every_row`, `recovered_counts_only_successes`, and **`all_failed_reports_zero`** (a pass in which every write fails reports zero recoveries — the direct anti-theorem for defect 2).

`Proofs/Recovery/Sweeps/Conversation.lean` adds the sweep, whose `Row` is the **duplicate group** (every doc sharing a `session_id`), not a single doc:
- `canonical_perm_invariant` — the canonical pick is independent of the order the store returns rows in. Load-bearing: DefraDB returns duplicates in **docID order, not recency order**, so "pick the latest" cannot be left to scan order.
- `duplicate_group_recovers` + `conversation_recover_idempotent` — the sweep is total on duplicate groups and converges (a second pass recovers nothing).
- The sweep contract pins `targetSelector = "_docID"`, which makes the session_id-filtered upsert illegal *by contract*.

Adding `agentConversation` to `PersistedRecoveryCollection` breaks the coverage theorem until the sweep is registered — that fence is why the model and the code land together.

## The fix

Every conversation write is addressed by `_docID` (update on hit, create on miss); no `upsert_AgentConversation` remains in the runtime. The canonical doc is chosen by an explicit total order mirroring Lean's `docRank` (newest `updated_at`, then richest, then greatest `_docID`). Recovery groups stuck docs by session, recovers the canonical doc, and **converges the duplicates onto it** rather than deleting them — the collection is `@branchable`/replicated, so a delete can be resurrected by a peer or fork the CRDT. `RecoveryReport` gains `conversations_failed` and `duplicate_conversation_sessions`; startup logs failures and duplicate sets distinctly from recoveries.

## Tests

The duplicate condition **cannot be seeded naively** — the shipped schema declares `session_id @index(unique: true)` and DefraDB enforces it on create. The wild stores exist because DefraDB cannot add an index to an *already-created* collection, so hosts whose collection predates the index keep duplicates permanently. `test_db_with_duplicate_tolerant_conversations` reproduces exactly that: register the pre-unique SDL first, so `ensure_runtime_schemas`' own `add_schema` no-ops on it.

- `recover_all_recovers_canonical_conversation_of_a_duplicated_session` — recovers 1 session (not 2 rows), both docs terminalized, second pass converges. Falsified: reintroducing defect 2 fails it with `left: 2, right: 1`, exactly the issue's symptom.
- `live_request_path_survives_a_duplicated_session` — drives the real lifecycle (`claim` → `prepare_session_with_identity`); pre-fix this died with `cannot upsert multiple matching documents`.
- Conformance: the Lean conversation sweep cases and the 4 outcome witnesses drive the **production** recovery pass, including the duplicate-store case.

Green: `cargo test -p defra-agent` (full package incl. 294 conformance), CLI, desktop-core, `cargo check --workspace --all-targets`, `lake build` (zero sorries).

## Acceptance stores

Validates against the acceptance cases in the issue thread (strangenas forensic specimen; `web-research-steward` / `inference-steward` on studio-1; `amygdalabook-steward`) — the seeded regression reproduces their shape (one duplicated `session_id`, zero duplicate `request_id`s). Worth noting for rollout: duplicates are *permanent* on those stores (the unique index cannot be backfilled), so the runtime tolerates them rather than assuming they get cleaned up.


jackzampolin:
**#694 merged** (7a8e9d32) — it landed defects 1&2 here: duplicate-tolerant canonical-pick recovery + counting successes not attempts. So this PR now overlaps/conflicts on the recovery path.

But **defect 3 is not covered by #694 and is worth keeping**: the live request path (`prepare_session_with_identity` on every claimed request, plus the status/title upserts) hitting the same `session_id`-filtered multi-match upsert. That's a real live-path wedge on duplicate-carrying stores (we suspect it explains some steward flakiness that was attributed to P2P).

Could you rebase onto main — reconcile 1&2 against #694's approach (query-first canonical pick, `_docID`-addressed updates), and preserve the defect-3 live-path fix? That's the part we specifically want.

jackzampolin:
Decision (Jack): **take #698 wholesale over #694** — the single-pass convergence, richness-aware canonical pick, full live-path coverage, and the Lean layer (Recovery/Outcome.lean + Sweeps/Conversation.lean) are the better design, so we want this to replace #694, not just cherry-pick defect-3.

I attempted the swap two ways (merge-main-into-698, and revert-#694-then-merge) and it needs your hand: it's a genuine **three-way reconciliation** on the conversation-upsert path, not a clean rebase. Specifics so it's fast for you:

1. **#694 is on main** (squash `7a8e9d32`). Its defects 1&2 overlap yours; taking #698 wholesale supersedes it. A `git revert 7a8e9d32` reverts clean on every recovery file **except** `session/conversation.rs`.
2. **`session/conversation.rs` now has requester-lineage work you branched before.** Main added `upsert_conversation_from_request_with_identity_and_requester_did` and a `{requester_did_field}` on the upsert `add:` (from the return-template/requester-lineage line, ~#713/#715). Your branch has `..._with_identity` without the `_requester_did` variant. **This must be reconciled, not overwritten** — your canonical-pick + `_docID`-addressing needs to carry main's `requester_did` scoping. This is the crux and why an auto-merge produces a missing-function / mismatched-caller break.
3. **`agent/runtime/startup.rs`** also carries #703's readiness-barrier changes (auto-merge keeps them; just take your recovery-logging hunks on top).
4. **`tests/conformance.rs`** import list needs the union: your `lean_recovery_outcome_cases` + main's `lean_startup_readiness_cases`.
5. Please **drop `.agents/recovery-693-plan.md`** — it shouldn't land.

Rebase onto current `main` (`8ea4be69`), reconcile (2) carefully, and it should go green. Ping me and I'll fast-track the review + merge.

jackzampolin:
Rebased and reconciled onto current `main` at `d5951f01`; refreshed head is `f4cb4ee7`.

The reconciliation keeps #698's `Recovery.Outcome` / duplicate-group sweep, richness-aware canonical selection, `_docID`-addressed live writes, single-pass duplicate convergence, and honest success/failure accounting. It also preserves current main's requester-lineage fields on conversation and recovery-response creates, startup readiness/goal/Codex changes, and drops `.agents/recovery-693-plan.md` as requested.

Local validation on the refreshed head:

- `lake build`
- `cargo test -p defra-agent duplicated_session -- --nocapture`
- `cargo test -p defra-agent generated_recovery_outcome_cases_fence_duplicate_tolerant_counting -- --nocapture`
- `cargo test -p defra-agent` — 1,147 unit, 312 conformance, all 48 lifecycle tests green
- `git diff --check`

Fresh CI is running now.

Issue #700, opened 2026-07-12
provision cannot recreate a tombstoned unique config row


jackzampolin:
## Summary

On v0.6.11, defra-agent provision cannot converge a partially bootstrapped home when a desired config row exists only as a tombstone. The diff reports the row as create, but upsert fails because the unique index still considers the deleted document.

## Observed case

AmygdalaBook had an initialized persistent identity and a quiet-store AgentPrincipal whose default behavior referenced default-tools. A server-first scratch boot had left default-tools deleted.

ToolSelection(showDeleted: true) returned exactly one matching row:

- selection_id: default-tools
- agent_did: did:key:z6MkeaEPsX24DZgRcSHvusJniqgyK9WVbBHcGvjJUoN9QQgV
- _deleted: true

Normal ToolSelection query did not return it.

With the server stopped:

    defra-agent provision --home /Users/johnzampolin/.defra-agent/amygdalabook-steward --root <pairing-free-root>

reported default-tools under create. Retrying failed with:

    can not index a doc's field(s) that violates unique index

The desired row cannot be restored through provision/config apply even though it is absent from the normal query and required by the desired manifest.

Evidence bundle: runs/agent-log-bundles/20260712T220701Z-amygdalabook-provision-tombstone

## Expected

Provision/config apply should restore or replace a tombstoned row when its unique key is desired, and then converge exactly. Operators should not need a live-store hand repair. The affected store is being preserved and replaced through the documented cold-copy/wipe/provision flow.

jackzampolin:
Completed by the DefraDB unique-index semantics update and tombstone-aware recreate identity work. The current provision path can recreate a logically unique config row after its predecessor was tombstoned, with live regression coverage.

Issue #559, opened 2026-06-25
Readiness barrier hangs (no timeout) when a runnable behavior repeatedly fails to build its client


jackzampolin:
## Summary

The startup readiness barrier can hang indefinitely (no timeout) whenever a behavior is classified **runnable** at snapshot time but then **fails to build its completion client** at slot start. The slot hot-restarts on each failure while `wait_ready()` blocks forever, so the process never reports `Ready`.

This is the general pattern behind the ChatGptCodex-credential case fixed in #530 (commit `937cbab7`); that PR closed the specific trigger (a credential-less ChatGptCodex behavior is now routed to `unavailable_behaviors` at snapshot time). This issue tracks hardening the **general** case.

## Mechanism

- Runnable-vs-unavailable is decided at snapshot time in `crates/defra-agent/src/agent/document_view/snapshot.rs` (backend `is_available()`, reference resolution, and now ChatGptCodex credential presence). Only runnable behaviors are added to the readiness barrier (`StartupBarrier::new`, `crates/defra-agent/src/agent/runtime/context.rs`).
- The completion client is built **later**, at slot start, in `RuntimeContext::run_behavior` (`crates/defra-agent/src/agent/runtime/context.rs:~201`). A build error there returns `Err` from the slot runner.
- The slot loop (`crates/defra-agent/src/agent/reconcile/slot.rs:~250`) logs and **restarts with backoff** on `Ok(Err(_))` — forever.
- `mark_behavior_ready` is only called once the behavior daemon's `run()` starts (`crates/defra-agent/src/agent/daemon.rs:~124`), i.e. **after** a successful build. So a behavior stuck failing-to-build is never removed from `pending_behaviors`, and `wait_ready()` (`crates/defra-agent/src/agent/runtime/context.rs`) never returns. The readiness task (`crates/defra-agent/src/agent/runtime/startup.rs:~230`) `select!`s only on cancel or `wait_ready()` — **no timeout**.

Anything that makes a *runnable-at-snapshot* behavior fail *at-build* reproduces the hang. Examples beyond the (now-fixed) missing OAuth credential:
- a transient backend build error (DNS/TLS/config) for an otherwise-available backend;
- a provider client constructor that rejects config the snapshot considered valid.

## Impact

A single un-buildable behavior wedges the whole process in non-`Ready` state and spins a hot restart loop, rather than degrading that one behavior and reporting `Ready` for the rest.

## Possible directions (to be designed)

[...]

PR #702, opened 2026-07-13
Rebind the Codex shim when a published generation supplies its behavior (#699)

### LINKED ISSUE #699: Codex shim stays disabled after config reconciliation creates bound behavior
## Summary

On v0.6.11, starting defra-agent server on a fresh identity-only/empty store disables the Codex shim because the configured bound behavior does not exist yet. Applying the manifest later through GraphQL makes the behavior runnable, but runtime reconciliation never retries shim binding, so the advertised WebSocket port remains closed until the whole server is restarted.

This affected all 19 fresh-store Amygdala fleet agents in a pairing-free bring-up.

## Reproduction

1. Preserve identity metadata but start with an empty DefraDB data directory.
2. Start defra-agent server with the Codex shim bound to behavior default.
3. Startup reports: Codex endpoint disabled: validating Codex shim bound behavior "default": no AgentBehavior document with that behavior_id exists.
4. Apply a manifest containing behavior default through config apply --graphql.
5. /healthz shows the runtime ready and default runnable.
6. The configured shim TCP port is still closed.
7. Restart the server without changing data or config.
8. The shim immediately listens and logs that it is running.

Observed on release v0.6.11. Amy's HTTP/runtime stayed healthy throughout; only the shim was absent.

## Expected

Either:

- runtime reconciliation should retry/start a shim disabled solely because its bound behavior was unavailable, or
- the server should expose an explicit shim rebind/restart operation and health signal.

The existing defra-agent provision --home ... --root ... flow avoids this when used before first server boot. A clean-room smoke test of that flow converged exactly. However, server-first followed by remote config apply is easy to reach in fleet recovery and currently leaves /healthz green while the operator UI is unreachable.

iverc:
Fixes #699.

## What was actually wrong

Every runtime subsystem that depends on which behaviors are runnable is a **subscriber** of the reconciler's published generation, so it re-derives its own enablement whenever the control plane changes. The Codex shim was not — and structurally could not be: `active_snapshot_tx`/`_rx` is created inside `run_agent` and never handed out, and `DefraAgent` only exposes boot-time views. So the shim sampled the control documents **once**, at boot.

On a fresh store that read found nothing, and the shim disabled itself for the life of the process. The `config apply` that arrived moments later made the behavior runnable and the runtime converged normally — while the shim, holding a boot-time verdict, kept its port closed. All 19 fleet agents came up that way, and because the shim appears in **no** health surface, all 19 also served a green `/healthz` behind a closed port.

So this is not a missing retry. The spec already proves the supply half (`ApplyReconcile.t_conv_runnable`: after a well-formed apply, every manifest behavior is runnable) and already carries the demand half's shape (`ActiveRuntimeSnapshot.wellFormed` requires `dispatchers = runnable`, re-established at every `publish`). The shim is a second runnable-gated subsystem that never consumed a conclusion the model already guarantees.

## Spec first

`proofs/Proofs/CodexShim/Binding.lean` — new, zero `sorry`. The shim re-derives its binding from each published generation, and the load-bearing part is that **not every reason to be unbound is one the control plane can retract**:

- `dependencyMissing` — the bound behavior is absent. A document fixes it, so this **must** converge.
- `hostResource` — the port is taken, or the bind address was refused. No document retracts it, so republishing must not resurrect the shim and the runtime must not spin on it. (Same shape as `PairingReconcile.dial_failure_is_nonconverging_fixpoint`.)

Collapsing the two makes a retry loop either useless (never retrying the fixable case — this bug) or noisy (hammering a taken port forever). Nine theorems, including `converges_when_dependency_published` (the #699 theorem — no restart transition exists in the model), `bound_never_unbinds`, `host_resource_is_fixpoint`, and `observePublish_coherent` (the shim's counterpart of `coherent_preserved`).

The listen is taken **inside** the transition, because acquiring the socket can still fail at the moment the behavior arrives. Recording the grant before the socket existed would force reality to walk `bound → unbound`, which `bound_never_unbinds` forbids — so `listen_failure_degrades_to_host_resource` covers that race explicitly.

Contract emitted as `codex_shim_binding_cases` (5 vectors) and consumed by `conformance::generated_codex_shim_binding_cases_pin_runnable_gated_binding`, which drives the real `ShimBinding` through every one. Reintroducing the defect fails it by witness name.

## The fix

- **`RuntimeSnapshotObserver`** (runtime): the host-facing signal that did not exist. Fires for the boot generation and every published generation, off the same watch channel the trigger engine already consumes.
- **`ShimBinding`** (runtime): the Rust mirror of the model. `grants_listen`/`settle_listen` exist because the real bind is async; `observe_publish` is defined in terms of them, so the async host path and the fenced path cannot drift.
- **Typed bind errors** (CLI): the failure class decides whether the shim may ever revive, so it is typed at the source rather than sniffed out of an error string.
- **Supervisor** (CLI): on a suppliable failure the shim now waits and binds on the generation that carries its behavior — `Codex endpoint pending: … no restart needed`, then `Codex endpoint bound: … (no restart was needed)`. A host-resource failure still degrades exactly as before.
- **`/healthz` `checks.codex_shim`**: `ok` / `pending` / `disabled` / `off`. Pending and disabled mark the payload `degraded` while keeping `ok: true` (the runtime really is serving) — a closed shim port can no longer read as fully healthy. `--no-codex-shim` is a deliberate choice and is not degraded.

## Tests

- **Regression** (`codex_shim_binds_when_config_apply_supplies_its_behavior`): boot with the shim bound to a behavior that does not exist, assert the port is closed, supply it through `config apply` against the **running** server, assert the port opens with no restart. Neuter the supervisor and it fails with *"the shim never bound … this is #699"*.
- The existing test that asserted permanent disable on a missing behavior asserted the bug; it now pins that the shim **waits**. The port-taken test is unchanged — that class is still terminal.
- Three `/healthz` tests for the visibility half.

## Gates

`cargo fmt --all --check`, `lake build` (zero `sorry`), `cargo test -p defra-agent` (full package incl. conformance + coverage ledger), `cargo test -p defra-agent-cli` (single-threaded, as CI runs it), `cargo check --workspace --all-targets` — all green.

Left for #700: provision cannot recreate a tombstoned unique config row.


jackzampolin:
Conflicts now — **#703 merged** (readiness-barrier demote-not-wedge, #559), which also reworks the startup/health path this PR touches. Needs a rebase onto main.

Two behavioral notes for when this lands (from review, non-blocking but operationally relevant):
1. **/healthz semantics change:** pending/disabled shims now flip `status` to `degraded` (with `ok:true`). Fleet alerting keyed on `status == "ok"` (Mission B) will fire for agents mid-bring-up until the shim binds — intended visibility, but coordinate before the next fleet redeploy.
2. **Lifecycle asymmetry:** a supervisor-bound shim's `JoinHandle` is dropped, so a post-bind accept-loop crash is swallowed → `/healthz` keeps reporting `codex_shim: ok/Listening` over a dead port for that one path. Narrow and strictly better than pre-fix, but consider feeding the handle back into the shutdown select in a follow-up.

Issue #706, opened 2026-07-13
Timestamp-less unique config rows (ToolSelection et al) cannot converge an identical-content recreate over a tombstone


iverc:
Follow-up from #700 / sourcenetwork/defradb.rs#1126.

Deletion is terminal in DefraDB and docIDs are content-addressed, so recreating a config row whose manifest content is **identical** to a tombstoned row regenerates the tombstoned docID. Upstream now rejects that create loudly ("deletion is terminal — recreate with distinct content to mint a new document identity") instead of silently zombieing, and the Task/Schedule/EventTrigger writers mint a fresh identity by stamping `updated_at`/`created_at` at apply time.

Collections without a timestamp field cannot mint a new identity this way. `ToolSelection` — the exact collection from #700's field incident — has no `created_at`/`updated_at`, and its writer is a tombstone-blind `upsert_`. An identical-content recreate over a ToolSelection tombstone now fails loudly (a strict improvement over the silent zombie / eternal apply loop), but it still cannot converge.

Options:
1. Add `created_at`/`updated_at` to ToolSelection (and audit the other config schemas: AgentBehavior, InferenceBackend, …) and extend the identity-minting recreate path to their writers.
2. A generic apply-path salt field (e.g. `applied_at`) on all config collections.

Either way the writers should share one `mint_recreate_identity` helper (currently per-module in `config_writes/{task,schedule,event_trigger}.rs`).

Done when: prune → re-add with identical content → re-apply converges for every config collection, with a regression test per collection class.

PR #707, opened 2026-07-14
Bump defradb.rs to the unique-index semantics fix and mint recreate identities over tombstones (#700)

### LINKED ISSUE #700: provision cannot recreate a tombstoned unique config row

iverc:
Fixes #700 by bumping defradb.rs to the unique-index semantics fix (sourcenetwork/defradb.rs#1126) and making the tombstone-aware config writers mint a fresh document identity when recreating over a tombstone.

## The two halves of #700

**Upstream (the wound):** the unique check was a bare index-key probe with no occupant-liveness awareness. Two merge orderings stranded entries owned by tombstones (a replicated update merged after a delete re-mints the entry while the `/del/` marker stays; an out-of-order delete cleans nothing when the doc isn't materialized yet), and with index constraints unpatchable there was no repair — recreating the value failed forever with `can not index a doc's field(s) that violates unique index`. #1126 makes stale entries **self-heal**: a unique conflict whose holder is deleted or missing is reclaimed on the spot (deletion is terminal, so the old holder can never return). Era-wounded stores repair themselves lazily on the exact operation that used to fail — no migration, no hand repair. Merges onto tombstoned docs no longer touch indexes (Go parity), closing both minting orderings, and live merge conflicts now converge to a deterministic winner instead of wedging the doc's sync forever (#1111).

**Downstream (the loop):** even with a clean index, recreating a config row whose manifest content is **identical** to the tombstoned row regenerates the same content-addressed docID — and a create onto a tombstoned docID does not produce a live row (the create reports success while the `/del/` marker keeps the row invisible). Provision would report `applied` yet the post-apply diff kept demanding the same create: an eternal, invisible loop. Both apply paths now **mint a fresh identity** on a recreate-over-tombstone by stamping `updated_at`/`created_at` with the apply time — the batched mutation builder (`custom_override_mutation_field`, which handled tombstones as plain creates and is the path that actually runs for diff-creates) and the per-document Task/Schedule/EventTrigger writers share one `mint_recreate_identity`. Distinct content → new docID → a genuinely live row → exact convergence.

`ToolSelection` — #700's literal collection — has no timestamp field to mint with; that residual is #706. Its different-content recreates (the common case, and the field incident's shape once the stale entry heals) converge; identical-content recreates still silently fail to materialize until #706 gives the writer something to mint with.

## Regression test

`cli_config_apply_local::reapply_recreates_a_pruned_unique_row`: apply a manifest with a task → prune it (tombstone) → re-add the same `task_id` → re-apply must converge exactly (`ok: true`, diff exact). Without the identity minting the test fails exactly as #700 reported: the apply loops on a create that never materializes.

## Baseline repair

defra-agent main's CI is red independently of this PR: the #1125 pin bump landed without updating the pinned live sync-status JSON fixture in `http/router.rs`, so `p2p_metrics_snapshot_decodes_pinned_live_sync_status` fails on clean main (`deny_unknown_fields` + the new `gossip_direction_filtered_total` / single-flight / receiver-pull counters and `peer_capacity_parks_total`). Repaired here with the missing fields and assertions.

## Gates

`cargo fmt --all --check`, `cargo test -p defra-agent` (full package incl. conformance), `cargo test -p defra-agent-cli` (single-threaded), `cargo test -p defra-agent-desktop-core`, `cargo check --workspace --all-targets` — all green.

## Worth knowing

Validating this work surfaced that the pairing reconciler **depends on** delete-then-identical-recreate of `PeerPairingApplied` rows. A Go-parity "deletion is terminal" guard upstream broke it immediately, so that guard was reverted and left as a flagged decision on #1126 — this PR does not change local create-onto-tombstone semantics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ dependencies = [
 [[package]]
 name = "acp"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=f750e009325a59ad4000e5dc31979c713438cfdf#f750e009325a59ad4000e5dc31979c713438cfdf"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=67ca1c6c0a17bccb51eecde9834beecf97cd888a#67ca1c6c0a17bccb51eecde9834beecf97cd888a"
 dependencies = [
  "async-lock",
  "async-trait",
@@ -2482,7 +2482,7 @@ dependencies = [
 [[package]]
 name = "blockstore"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=f750e009325a59ad4000e5dc31979c713438cfdf#f750e009325a59ad4000e5dc31979c713438cfdf"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=67ca1c6c0a17bccb51eecde9834beecf97cd888a#67ca1c6c0a17bccb51eecde9834beecf97cd888a"
 dependencies = [
  "async-trait",
  "bytes",
@@ -5861,7 +5861,7 @@ dependencies = [
 [[package]]
 name = "crdt"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=f750e009325a59ad4000e5dc31979c713438cfdf#f750e009325a59ad4000e5dc31979c713438cfdf"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=67ca1c6c0a17bccb51eecde9834beecf97cd888a#67ca1c6c0a17bccb51eecde9834beecf97cd888a"
 dependencies = [
  "async-trait",
  "defra-core",
@@ -5998,7 +5998,7 @@ checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 [[package]]
 name = "crypto"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=f750e009325a59ad4000e5dc31979c713438cfdf#f750e009325a59ad4000e5dc31979c713438cfdf"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=67ca1c6c0a17bccb51eecde9834beecf97cd888a#67ca1c6c0a17bccb51eecde9834beecf97cd888a"
 dependencies = [
  "aes-gcm",
  "blst",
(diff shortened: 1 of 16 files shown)


jackzampolin:
Heads up — main moved under this PR while landing the batch:
- **#1126 merged** (defradb.rs main is now `89a4b46f`). This PR pins to `67ca1c6c`, which is the pre-squash branch SHA and is **not on main** — please re-pin the defradb.rs deps to `89a4b46f`.
- **#712 merged** ("Recreate config docs after tombstone deletion", #706), which already added `mint_recreate_identity` calls in `config_import.rs`. This PR now shows **CONFLICTING/DIRTY** against those files — needs a rebase onto main to reconcile the config-writer half.

After rebase, this is the release vehicle for #1125/#1126/#1127 → the fleet. Note the v0.6.12 lesson: a defradb.rs pin bump can add struct fields to `SyncStatus`/`PushBacklogSnapshot` that the `deny_unknown_fields` conformance adapter rejects — run `cargo test --workspace --no-run` (not just `cargo check --workspace`) so the conformance/prometheus literal fences are compiled. (#701 already wired the v0.6.12 counters into Prometheus emission.)

jackzampolin:
Status update — this PR's two halves have diverged from the plan:

- **Pin bump: delivered separately.** Rather than untangle this branch's config-writer conflicts, the defradb.rs pin bump to `89a4b46f` (#1126 unique-index + #1127 ACK) went in as a clean PR (#721, merged). So the pin half of #700 is done.
- **Recreate-identity work: appears already on main.** `config_client` now carries tombstone-aware recreate (`event_trigger`/`schedule`/`task`.rs "live row missing _docID after recreate", `config_import.rs:632` tombstone detection) via #712 + the #695 config_client refactor — a different structure than this branch's `mint_recreate_identity`.

If there's still unique behavior here not covered by main's recreate path, please rebase what's left; otherwise this is likely safe to **close as superseded**. Your call — flagging rather than closing someone else's PR.

jackzampolin:
Final isolated-worktree review: this PR should be closed as superseded, not rebased.

The DefraDB pin here is obsolete; current `main` already carries the release-tested `e74733a` pin from #752 plus the watcher/P2P compatibility work. The recreation changes are carried by #753 with the shared runtime helper, six schema fields, integration coverage, and the Lean protected-field fence. The sync-status fixture repair is already on `main`.

No behavior remains unique to this branch. Closing it does not imply #753 is ready: blocking repeated-tombstone and imperative-writer findings are posted there separately.

PR #736, opened 2026-07-15
Lean-fence the timeline interleaving; share the ordering skeleton across shells (#608)

iverc:
First execution step of the desktop/mobile parity work (`.agents/desktop-mobile-parity-contract.md`, T0). Tackles the single biggest parity risk *before* a second shell exists.

## The problem

`build_rendered_timeline` — the function that interleaves a session's messages with their tool groups, places the pending turn, appends orphan tool groups, and the live-assistant overlay — lived **only in the desktop Tauri bridge**, unshared and unfenced. It carries subtle discipline (first-wins dedup by message key *and* presentation key, the attached-vs-orphan tool-group partition, `None`-sequence-sorts-first, overlay-not-duplicating-the-trailing-assistant). A second shell (mobile, a full peer node) that re-implements this will drift on **order** and on **which tool group is an orphan** — the exact class of divergence the parity contract exists to prevent, and impossible to catch without a shared, fenced source of truth.

## The approach — presentation-neutral skeleton in the shared layer

Rather than lift the function verbatim (which would drag the whole bridge view-type zoo down into the core), this splits along the contract's own line — **presentation-neutral projection in the shared layer; visual mapping in the shell**:

- **`defra_agent_protocol::timeline::build_timeline_order`** — a new pure function next to `derive_turn` that computes the ordered list of *neutral slots* (`Message{key,seq,role}` / `ToolGroup{seq}` / `Pending` / `Overlay`) from ordering-relevant inputs. This is the order + partition — the semantics.
- **The bridge's `build_rendered_timeline` is now a thin adapter**: it projects the rich views into neutral inputs, delegates ordering to the shared skeleton, then maps each returned slot back to its rich `RenderedTimelineItem` (attaching content, rendered tools, pending details). The overlay dedup stays in the adapter via the existing helper — identical behavior.

A mobile shell calls the same `build_timeline_order` and does its own rich mapping, so it **cannot** diverge on order.

## Lean fence

`proofs/Proofs/ClientShell/Timeline.lean` models `build_timeline_order` (zero `sorry`, 8 theorems):
- **Partition** — `group_attached_or_orphan` (every tool group is placed — attached to its owner or orphan, never dropped) + `group_not_both` (never placed twice).
- **Overlay** — `overlay_shown_iff` (emitted iff present ∧ not a duplicate of the trailing assistant) + `overlay_is_last`.
- **Pending** — `pending_shown_iff` (and structurally before orphan groups).
- **Dedup** — `kept_keys_nodup` (first-wins by key → no message rendered twice).

Model boundary (stated in the file): the message list is taken already-sorted by the shell's total sequence order; the fence is the interleave/partition/tail discipline *on top of* that order — the part a second shell re-implements.

## Tests

- **Protocol conformance** (6 tests): each drives the real `build_timeline_order` and asserts a proven invariant, citing the theorem — partition + attach-adjacency, overlay shown-iff + last, pending placement, key-dedup, the presentation-token-also-drops-the-second-group-attach nuance, and `None`-sorts-first. Breaking the attachment logic fails the partition test.
- **Behavior preservation**: the desktop `session_timeline` suite (8/8) and the full `bridge::snapshot` suite (77/0 single-threaded) pass unchanged. The 4 flakes seen under `--test-threads` many were the known node-spawn port race — all pass in isolation and single-threaded.

## Gates

`cargo fmt --all --check`, `lake build Proofs.ClientShell` (zero `sorry`), `cargo test -p defra-agent-protocol`, `cargo check --workspace --all-targets`, `cargo clippy` on the touched crates — green.

## Follow-ups (in the contract doc)

This is the **template** for the remaining T1 move-downs (session/runtime projection, cancel cascade, denial label table): extract the neutral skeleton to `protocol`, fence it, adapter in the shell. The crate rename (`desktop-core` → `client-core`) and the FFI scaffold remain T0. A latent nicety this surfaced: the neutral function guards against double-placing a group when two surviving messages share a sequence (the old bridge code did not) — a stricter-but-correct behavior that only differs on malformed input.

### DIFF
--- a/apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_timeline.rs
+++ b/apps/desktop-tauri/src-tauri/src/bridge/snapshot/tests/session_timeline.rs
@@ -103,7 +103,9 @@ fn session_snapshot_deduplicates_persisted_rows_from_multiple_sources() {
     let mut rows = make_streaming_store_with_response_content("").to_rows();
     rows.responses.clear();
 
-    let duplicate_user = rows.messages[0].clone();
+    let mut duplicate_user = rows.messages[0].clone();
+    duplicate_user.sequence = Some(9);
+    duplicate_user.content = Some(user_message_json("later duplicate"));
     rows.messages.push(duplicate_user);
     rows.message_source_agent_dids = vec![None, Some("did:defra:amy".to_string())];
 
@@ -119,7 +121,10 @@ fn session_snapshot_deduplicates_persisted_rows_from_multiple_sources() {
     };
     rows.messages.push(assistant.clone());
     rows.message_source_agent_dids.push(None);
-    rows.messages.push(assistant);
+    let mut duplicate_assistant = assistant;
+    duplicate_assistant.sequence = Some(10);
+    duplicate_assistant.content = Some(assistant_message_json("later duplicate"));
+    rows.messages.push(duplicate_assistant);
     rows.message_source_agent_dids
         .push(Some("did:defra:amy".to_string()));
 
@@ -144,6 +149,22 @@ fn session_snapshot_deduplicates_persisted_rows_from_multiple_sources() {
         })
         .collect::<Vec<_>>();
     assert_eq!(kinds, vec!["user", "assistant"]);
+    assert!(matches!(
+        &snapshot.timeline_items[0],
+        RenderedTimelineItem::UserMessage {
+            sequence: Some(1),
+            content,
+            ..
+        } if content == "hello"
+    ));
+    assert!(matches!(
+        &snapshot.timeline_items[1],
+        RenderedTimelineItem::AssistantMessage {
+            sequence: Some(2),
+            content: Some(content),
+            ..
+        } if content == "hello back"
+    ));
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


jackzampolin (changes requested):
Blocking correctness finding from an isolated-worktree review:

The shared skeleton deduplicates message keys first-wins in `crates/defra-agent-protocol/src/timeline.rs:118-132`, but the desktop adapter builds `rendered_message` with ordinary `BTreeMap::insert` at `apps/desktop-tauri/src-tauri/src/bridge/snapshot/timeline.rs:183-185`. A later duplicate therefore overwrites the rich content that is mapped back at lines 213-215. The resulting slot/order can belong to the first row while displayed content and sequence come from the last duplicate.

This matters for the real multi-source duplicate-row path that the existing desktop test covers. That test clones identical rows, so it only checks item count and cannot expose the first-wins violation. Please retain the first rendered value, for example with an entry/or-insert path or by mapping accepted slots directly, and add a regression with duplicate keys carrying different content/sequence.

Verification completed: `lake build Proofs.ClientShell`, all 105 protocol tests, formatting, and the published CI pass.

jackzampolin (approved):
Re-reviewed Ivan’s follow-up. The rich timeline map now matches the Lean-fenced skeleton’s first-wins semantics, and the strengthened regression verifies both user and assistant duplicate rows preserve the original sequence/content. The exact implementation regression, protocol suite, Lean ClientShell build, format, and diff checks pass.

Issue #739, opened 2026-07-15
Structural (JSON/YAML path) edit mode for edit_file


jackzampolin:
Deferred from #738 (decided during design review). Amy's config-editing workload (#738: batch-updating 23 inference-profile JSONs, `.max_turns` -> 250) is ultimately structural, not textual — she fell back to python json.load/json.dump. Neither Codex nor oh-my-pi has path-based structural editing (omp offers only ast-grep tree patterns as a separate tool), so this is greenfield design:

- Path syntax (jq-style? JSON Pointer?), scoped to JSON/YAML/TOML first
- Type preservation + formatting/round-trip fidelity (comments in YAML/TOML are the hard part)
- Relationship to the #738 matcher: separate `match_mode: structural`? separate tool?
- Composes with content-hash preconditions from #724/#738

Prereq: #738's edit_file rework (match ladder, dry-run, hash preconditions) should land first — this mode plugs into that surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #740, opened 2026-07-15
edit_file: drift-tolerant match ladder, dry-run, regex mode, content-hash preconditions (#738, #724)

### LINKED ISSUE #724: Add content-addressed file edits with stale-write rejection
## Context

The native read_file tool returns line-numbered content, while edit_file performs exact-text replacement against whatever bytes are present when the edit executes. Exact matching prevents some stale edits, but it does not prove that the model is editing the same file version it previously inspected:

- the selected old_text may still exist after an unrelated concurrent change
- a replace_all call may affect newly introduced matches
- a model may issue an edit based on an old partial read
- the durable tool record does not state which observed file version authorized the mutation

For a coding harness, file mutations should support an explicit optimistic-concurrency precondition.

## Oh My Pi prior art

Oh My Pi's hashline edit mode records a whole-file snapshot when the model reads a file. The read result includes a content-derived tag; later edits cite anchors associated with that snapshot, and stale anchors are rejected before mutation.

References:

- [read tool behavior](https://github.com/can1357/oh-my-pi/blob/79faf94f265100a5c05234a16ce67cd621f6e5e8/docs/tools/read.md)
- [edit tool behavior](https://github.com/can1357/oh-my-pi/blob/79faf94f265100a5c05234a16ce67cd621f6e5e8/docs/tools/edit.md)
- [hashline implementation](https://github.com/can1357/oh-my-pi/tree/79faf94f265100a5c05234a16ce67cd621f6e5e8/packages/coding-agent/src/edit/hashline)

Defra-agent should borrow the optimistic-concurrency property, not necessarily OMP's exact display grammar or short hash.

### 1. Return a byte-exact content identity from reads

read_file should return a stable cryptographic hash of the complete raw file bytes, even when the model-facing content is a line range or is truncated.

The compact text and structured JSON projections should expose the same identity, for example:

~~~json
{
  "path": "src/main.rs",
  "content_hash": "sha256:...",
  "start_line": 1,
  "end_line": 200,
  "truncated": true
}
~~~

Specify byte normalization explicitly. Prefer hashing raw bytes so line-ending or encoding changes are observable rather than silently normalized away.

### 2. Add an edit precondition

Add an optional expected_content_hash to edit_file and define a guarded mode in which it is required.

Before computing replacements, the runtime:

1. resolves the path under the existing workspace/root policy
2. reads the current bytes
3. computes the same content hash
4. rejects without writing if it differs from expected_content_hash
5. returns the current hash plus a concise re-read/retry instruction

Successful output should include both the expected/pre-edit hash and the resulting post-edit hash.

Unset configuration must preserve compatibility initially. A behavior/tool policy should be able to require guarded edits once clients and prompts support the field.

### 3. Explore hashline-style anchors

After the basic compare-before-edit contract is correct, evaluate a model-facing anchored edit format that lets the model cite compact line/range anchors derived from a particular snapshot instead of retransmitting large old_text blocks.

Requirements for any anchor format:

- collision resistance appropriate to its authorization role
- deterministic resolution against one file snapshot
- clear stale-anchor error
- no fallback that silently applies an anchor to a different version
- compatibility with partial reads and paged output
- lower token cost or higher edit success demonstrated by a benchmark

The durable authorization precondition must use the full cryptographic identity even if the UI/model sees a shorter display tag.

### 4. Define the race and write contract

A pre-write hash check alone has a residual check-to-write race with external writers. The design must state the guarantee provided:

- serialization among defra-agent writes to the same resolved path
- atomic replacement where supported
- handling of symlinks/path replacement between resolution and write
- what can and cannot be guaranteed against unrelated external processes

Do not claim filesystem-wide compare-and-swap semantics unless the implementation actually provides them.

## Persistence and observability

Persist enough structured information on the normal tool-call/result path to answer:

- which file identity the model expected
- which identity the runtime observed
- whether the call was rejected as stale
- the resulting identity after success

A stale rejection is a normal precondition failure, not a partial write. It must never leave a success-looking AgentToolCall result.

## Formal model and conformance

Start with the mutation contract:

- an edit may transition file content only when the expected identity equals the observed pre-state identity
- a rejected stale edit leaves bytes unchanged
- the successful post-state is exactly the declared replacement applied to the observed pre-state
- repeated application with the old expected hash is rejected
- path/root policy remains authoritative

Drive Rust tests from conformance cases covering matching/mismatching hashes, partial reads, replace_all, empty files, line-ending changes, symlink/path races within the supported guarantee, and concurrent runtime calls. Zero sorrys if the file-mutation contract enters Lean.

## Acceptance criteria

- read_file exposes one documented full-file content identity in compact and JSON output.
- edit_file accepts and enforces expected_content_hash.
- A behavior/tool policy can require guarded edits without silently weakening the operator ceiling.
- Stale rejection performs no write and returns the current identity.
- Success returns the post-edit identity.
- The check and mutation use one documented synchronization/atomicity contract.
- Tool-call persistence distinguishes stale rejection from execution failure and success.
- Tests cover concurrent and adversarial path cases.
- A hashline-style model-facing format is either specified with benchmark evidence or explicitly deferred after the base guarded-edit slice.
- cargo test -p defra-agent and cargo check --workspace --all-targets pass.

## Non-goals

- Automatic three-way merge.
- Git commits or worktree snapshots.
- Pretending a short display hash is collision-proof authorization.
- Replacing existing root/path/command policy.
- LSP integration.

### LINKED ISSUE #738: edit_file: add fuzzy/regex match modes and dry-run preview
## Problem

`edit_file` requires exact `old_text` match with zero tolerance — no substring matching, no regex, no fuzzy matching. When it fails, there's no preview mode to see what the tool *would* have changed before committing.

## Real-world evidence from Amy (fleet steward agent)

In a single session (2025-07-21, 19:46–21:21Z), I made 3+ `edit_file` attempts:

### Failed: `edit_file` on inference profile JSON
I needed to change `max_turns: 20` to `max_turns: 250` in a Kitchen agent profile. The `edit_file` call failed because the exact `old_text` didn't match due to trailing whitespace or slight formatting differences. I **gave up** and used `write_file` to overwrite the entire file instead — a riskier operation that could have clobbered unrelated fields.

### Successful but brittle: `edit_file` on kitchen behavior JSON
Changing `backend_id` from `workstation-1` to `workstation-2` and `model_name` from `d4f` to `glm52`. This worked only because I had the exact bytes from `cat`. A single trailing space or different indentation would have failed.

### Replaced with `write_file` entirely for 23+ inference profiles
Rather than risk 23 individual `edit_file` calls (each requiring perfect byte-match), I wrote a Python script that used `json.load`/`json.dump` to batch-update all profiles in `~/amygdala/infra/agents/*/inference-profiles/*/object.json`. This was the pragmatic call but meant bypassing the built-in tool entirely.

## Pattern observed

**80% of mutation work in this session ended up using raw Python/CLI tooling instead of `edit_file`.** The tool is only usable when you can guarantee exact byte-for-byte `old_text` — which is rare in practice for JSON, YAML, or any formatted file where whitespace, indentation, or line endings may vary.

### 1. Match modes (ordered by value)

| Mode | Behavior | Example |
|---|---|---|
| `exact` (current) | Byte-exact match, fail on first mismatch | Keep for backward compat |
| `contains` | Match substring within a line | `"max_turns": 20` -> `"max_turns": 250` anywhere |
| `regex` | Match pattern, capture groups for dynamic replacement | `stream_liveness_timeout_secs: \d+` -> `3600` |
| `structural` | Parse as JSON/YAML, match by path | `.max_turns` -> `250` (no whitespace concerns) |

### 2. Dry-run / preview mode

```
{
  "path": "config.json",
  "old_text": "...",
  "new_text": "...",
  "dry_run": true
}
```

Returns a diff/patch of what would change without modifying the file. Lets the agent inspect the match before committing.

### 3. Better error messages
When `old_text` isn't found, report:
- How many characters matched before failure
- The byte offset / line where the first mismatch occurred
- The actual bytes found at that position

### 4. Convenience modes
- `insert_after: "search string"` — append after match rather than replace
- `prepend_before: "search string"` — insert before match
- `delete: true` — remove `old_text` without replacement

## Usage context

- Agent working directory: per-request workspace under the configured root
- File types: primarily JSON, YAML, Markdown, TOML, and source code
- Common failure mode: trailing whitespace, carriage returns, tab vs space indentation, trailing newlines

## See also

- `replace_all` field exists (good) — but defaults to `false`, so multiple occurrences require multiple calls or knowing to set it
- `write_file` exists as a nuclear option but has no merge/safe-patch semantics

jackzampolin:
Fixes #738. Fixes #724.

Production shape this fixes (amy's own filing, #738): exact byte-match `edit_file` failed on trailing whitespace / CRLF / indentation drift, so **80% of her mutation work bypassed the tool entirely** — falling back to whole-file `write_file` overwrites and Python scripts, with no preview and no concurrency protection.

## Design (prior art researched in local Codex + oh-my-pi checkouts)

Codex's `apply_patch` matcher is a deterministic relaxation ladder (exact → rstrip → trim → unicode-normalize) with no fuzzy scoring; oh-my-pi adds Levenshtein-fuzzy apply behind threshold+dominance guards, rich failure diagnostics, dry-run, and content-hash staleness protection. This PR takes the ladder (deterministic, provable), oh-my-pi's diagnostics/dry-run/hash ideas, and draws the line at **fuzzy-applies-edits: similarity scoring is diagnostics-only** — it ranks closest-match error hints but never selects an edit site.

## Lean first (`Proofs/EditMatch/`, zero sorries)

- **E1** exact-hit priority (relaxation never shadows an exact match)
- **E2** occurrence soundness; **E3** ladder keys provably coarsen (a rung fires only when every stricter rung missed everywhere)
- **E4** ambiguity gate: ≥2 occurrences without `replace_all` is an error, never a guess
- **E5** one pure decision shared by dry-run and apply; **E8** no-op honesty
- **E6** stale-precondition gate (#724): rejects before any matching, document untouched
- **E7** exact windows are literal (insert_after provably preserves the matched text)

Conformance home `tests/conformance/edit_match.rs` fences E1–E8 against the real matcher (structure-fence entry added).

## What the model gets

- **`edit_file`**: drift-tolerant ladder with the replacement re-indented to the matched site; `match_strategy` reported in metadata. `dry_run: true` returns the numbered diff without writing. `match_mode: "regex"` (opt-in, bounded compile, capture groups). `operation: insert_after | insert_before | delete`. `expected_content_hash` rejects stale edits before matching, with the current hash and a re-read instruction. Success echoes the numbered hunk diff, pre/post hashes, and `first_changed_line`; writes are byte-verified after the fact.
- **Failure UX**: not-found errors carry the closest match (% similarity, line, first-differing line pair) and explicit "re-read, don't retry" guidance; ambiguity errors list the occurrence locations.
- **`read_file` / `write_file`**: raw-byte `content_hash` (sha256, whole file even on ranged/truncated reads) — the #724 optimistic-concurrency handle.
- BOM and CRLF round-trip: matching runs in LF space, the original flavor is restored on write.

## Verification

- 16 unit tests (each of amy's documented failure modes has one), 7 conformance fences, 8 tool-wiring tests (stale-gate ordering, dry-run no-write, hash echo, CRLF round-trip, regex + operations end-to-end).
- **Live e2e against real inference** (`tests/e2e_live/edit_file_live.rs`, gated on `DEFRA_AGENT_D4F_LIVE=1`): seeds a CRLF + trailing-whitespace config, asks the model to change `max_turns` via read → `edit_file` with `expected_content_hash`, bash disabled so there is no scripting escape hatch. Asserts the request completes, the file is correctly edited with CRLF preserved, a completed `edit_file` call carries `match_strategy`, and **`write_file` was never used**. Ran against the fleet d4f (DeepSeek-V4-Flash): PASSED.
- Gates: full `cargo test -p defra-agent` ✅, `cargo check --workspace --all-targets` ✅, `lake build Proofs.EditMatch` ✅, zero clippy warnings in new code.

## Follow-ups (filed)

- #739 structural (JSON/YAML path) edit mode — deferred by design decision.
- Guarded-mode policy knob (require `expected_content_hash` per behavior/tool policy) — #724's remaining slice once clients and prompts carry the field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6734,6 +6734,7 @@ dependencies = [
  "proptest",
  "query",
  "rand 0.9.2",
+ "regex",
  "reqwest 0.12.28",
  "rig-core",
  "rmcp 1.3.0",
--- a/crates/defra-agent/Cargo.toml
+++ b/crates/defra-agent/Cargo.toml
@@ -40,6 +40,7 @@ query.workspace = true
 rand.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+regex = "1"
 sha2.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
--- a/crates/defra-agent/proofs/Proofs.lean
+++ b/crates/defra-agent/proofs/Proofs.lean
@@ -48,5 +48,6 @@ import Proofs.SelfConfig
 import Proofs.ReversePairingHandlers
 import Proofs.Identity
 import Proofs.Skills
+import Proofs.EditMatch
 import Proofs.EventDelivery
 import Proofs.Conformance.EventDelivery
--- a/crates/defra-agent/proofs/Proofs/EditMatch.lean
+++ b/crates/defra-agent/proofs/Proofs/EditMatch.lean
@@ -1,0 +1,12 @@
+import Proofs.EditMatch.Model
+import Proofs.EditMatch.Properties
+
+/-!
+# EditMatch (#738, #724)
+
+The `edit_file` matcher: deterministic relaxation ladder (exact →
+trailing-whitespace → trim/indentation → unicode-normalized), ambiguity
+gating, convenience-operation desugaring, one pure decision shared by
+dry-run and apply, and the #724 optimistic-concurrency stale gate.
+Conformance home: `tests/conformance/edit_match.rs`.
+-/
--- a/crates/defra-agent/proofs/Proofs/EditMatch/Model.lean
+++ b/crates/defra-agent/proofs/Proofs/EditMatch/Model.lean
@@ -1,0 +1,175 @@
+/-!
+# EditMatch — model (#738, #724)
+
+The `edit_file` matcher: a deterministic relaxation ladder over lines
+(exact → trailing-whitespace-insensitive → trim/indentation-insensitive →
+unicode-normalized), an optimistic-concurrency precondition (expected
+content identity), uniqueness gating, and a single pure decision function
+shared by dry-run and apply.
+
+A line is abstracted as leading blanks, body, and trailing blanks; interior
+whitespace lives in `body`, so trim-level strategies never see it. The
+unicode fold is an arbitrary idempotent character map (the runtime's
+punctuation/space normalization is one instance). Content identity is
+modeled by the document itself: the runtime's cryptographic hash is the
+boundary that makes `expected = current` checkable without the bytes.
+-/
+
+namespace Proofs.EditMatch
+
+/-- A line: leading blank count, body characters, trailing blank count. -/
+structure Line where
+  lead  : Nat
+  body  : List Char
+  trail : Nat
+  deriving Repr, DecidableEq
+
+abbrev Doc := List Line
+
+/-- Ladder strategies, strictest first. -/
+inductive Strategy
+  | exact
+  | trailingWs
+  | trim
+  | unicode
+  deriving Repr, DecidableEq
+
+namespace Strategy
+
+/-- Ladder order: every strategy the matcher tries, strictest first. -/
+def ladder : List Strategy := [.exact, .trailingWs, .trim, .unicode]
+
+end Strategy
+
+/-- The comparison key a strategy sees for a line. Coarser strategies
+    project away more of the line. `fold` is the unicode normalization. -/
+def keyAt (fold : Char → Char) : Strategy → Line → Nat × List Char × Nat
+  | .exact,      l => (l.lead, l.body, l.trail)
+  | .trailingWs, l => (l.lead, l.body, 0)
+  | .trim,       l => (0, l.body, 0)
+  | .unicode,    l => (0, l.body.map fold, 0)
+
+/-- Two line windows match under a strategy iff their keys agree pointwise. -/
+def windowMatches (fold : Char → Char) (s : Strategy) (window pat : Doc) : Bool :=
+  window.length == pat.length
+    && (window.zip pat).all fun (a, b) => keyAt fold s a == keyAt fold s b
+
+/-- Does `pat` match `doc` at index `i` under strategy `s`? -/
+def matchesAt (fold : Char → Char) (s : Strategy) (doc pat : Doc) (i : Nat) : Bool :=
+  windowMatches fold s ((doc.drop i).take pat.length) pat
+
+/-- All match positions for `pat` in `doc` under `s`. -/
+def occurrences (fold : Char → Char) (s : Strategy) (doc pat : Doc) : List Nat :=
+  (List.range (doc.length + 1 - pat.length)).filter fun i =>
+    matchesAt fold s doc pat i
+
+/-- The ladder decision: the first strategy (strictest first) with at least
+    one occurrence, together with all its occurrences. -/
+def ladderMatch (fold : Char → Char) (doc pat : Doc) :
+    Option (Strategy × List Nat) :=
+  go Strategy.ladder
+where
+  go : List Strategy → Option (Strategy × List Nat)
+    | [] => none
+    | s :: rest =>
+      let occ := occurrences fold s doc pat
+      if occ.isEmpty then go rest else some (s, occ)
+
+/-- Splice `repl` over the `len`-line window at `i`. -/
+def splice (doc : Doc) (i len : Nat) (repl : Doc) : Doc :=
+  doc.take i ++ repl ++ doc.drop (i + len)
+
+/-- Re-indent the replacement to the matched site: shift every replacement
+    line's lead by the delta between the matched window's first line and the
+    pattern's first line. Only coarse strategies (which ignored `lead` when
+    matching) re-indent; exact and trailing-ws matched the lead literally. -/
+def reindent (s : Strategy) (doc pat repl : Doc) (i : Nat) : Doc :=
+  match s with
+  | .exact => repl
+  | .trailingWs => repl
+  | _ =>
+    match doc.drop i, pat with
+    | m :: _, p :: _ =>
+      repl.map fun l => { l with lead := l.lead + m.lead - p.lead }
+    | _, _ => repl
+
+/-- One edit request. `expected` is the optimistic-concurrency precondition:
+    the document version the model believes it is editing (#724). -/
+structure Request where
+  pattern     : Doc
+  replacement : Doc
+  replaceAll  : Bool
+  expected    : Option Doc
+  deriving Repr
+
+/-- Decision outcomes. `applied` carries the full result document: dry-run
+    and apply are projections of the same decision. -/
+inductive Outcome
+  | applied (result : Doc) (strategy : Strategy)
+  | rejectedStale
+  | notFound
+  | ambiguous (strategy : Strategy) (count : Nat)
+  | noop (strategy : Strategy)
+  deriving Repr
+
+/-- Replace every occurrence, right-to-left so indices stay valid. -/
+def spliceAll (s : Strategy) (doc pat repl : Doc)
+    (occ : List Nat) : Doc :=
+  occ.reverse.foldl
+    (fun acc i => splice acc i pat.length (reindent s doc pat repl i))
+    doc
+
+/-- The result document for an accepted match set. -/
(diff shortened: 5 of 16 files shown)


jackzampolin:
All six findings addressed in `02b2a630` at head, each with a red-then-green fence. Disposition:

1. **Lost updates (P1)** — per-canonical-path async lock now covers read → hash-check → match → write; a 20-round concurrent-edit fence reproduced the loss before the fix and passes deterministically after. Boundary documented at the lock: within-process serialization + the optimistic hash check; external writers race as ordinary POSIX writes (no OS lock is claimed), lock keys are canonical paths so symlinked spellings serialize together.
2. **UTF-8 corruption (P1)** — strict `String::from_utf8` with the first-invalid-byte offset in the error; fence asserts the file bytes are untouched.
3. **Overlapping splice panic (P1)** — reproduced exactly (OOB on the second reverse splice). Selection is now greedy non-overlapping, and since this was a spec hole, the Lean model was updated first-class: `selectDisjoint` with subset + pairwise-gap theorems (E9), conformance mirror `edit_match_overlapping_windows_apply_disjoint_selection`.
4. **Unicode preview panic (P1)** — char-boundary-safe truncation; the finding also exposed that the original test's match was under the 120-byte cap — test premise fixed (`needle.*` so the match itself exceeds the cap).
5. **Regex inserts (P2)** — `${0}` whole-match substitution in regex-mode desugaring; the fence now runs regex + insert_after combined, plus regex delete.
6. **Trailing-newline patterns (P2)** — codex-style retry: trailing empty pattern line (and one trailing replacement newline) dropped for the relaxed rungs; `"foo   \n"` matches `"foo\nnext\n"` at the trailing-whitespace rung.

Gates re-run at `02b2a630`: full package suite ✅, workspace all-targets ✅, `lake build Proofs.EditMatch` ✅ (zero sorries), live e2e vs d4f re-run ✅.

jackzampolin:
Round-2 findings addressed in `22043889`, each red-then-green:

1. **write_file outside the lock (P1)** — both mutators now share the per-canonical-path lock (renamed `FILE_MUTATION_LOCKS`, boundary comment covers both). Fence: 20 rounds of concurrent edit_file + write_file asserting the final state is always one of the two legal serial orders — the torn third state (edit reads before the write, lands after it) reproduced before the fix.
2. **Regex insert $-corruption (P2)** — regex operations compose inside one replacer closure per your suggestion: `Captures::expand` gives the user template replace-mode semantics, inserts splice the actual matched text around the expansion; `"${0}"` template composition removed. Fenced with your lone-`$` repro, the `$$` escape, and capture refs inside insert text — combined in single calls this time.
3. **Newline-consumption divergence (P2)** — relaxed trailing-newline matching now merges a non-newline-terminated replacement with the following line, so drifted patterns produce byte-identical results to the exact pass. Fenced by asserting equality against the exact-pass result for single, adjacent-window replace_all (`barbarnext`), and delete.

Gate at `22043889`: full suite ✅ (one unrelated pre-existing flake surfaced — `backend_health::cycle_marks_reachable_unknown_backends_promotable`, 1-in-4 under parallel load, passes isolated; filed #743), workspace all-targets ✅, conformance ✅, live e2e vs d4f re-run ✅.

jackzampolin:
Round-3 findings addressed in `af19888d`:

1. **Newline-only replacement degraded to deletion (P2)** — root cause was one layer deeper than the strip: the replacement path conflated "one empty line" with "zero lines". Replacement handling now lives in line space, where `"\n"` splits to a blank line that survives the trailing-newline pop. Fenced with your repro (single + replace_all) asserted byte-equal to the exact pass — and the refactor surfaced the mirror-image divergence (no-newline deletes dropped the emptied line where exact keeps it), which is also fixed; the round-1 E9 fences' expectations encoded that inconsistency and are corrected with rationale in place.
2. **Alias-sensitive lock keys for new files (P2)** — `canonical_lock_key` canonicalizes the deepest existing ancestor and appends the missing suffix, per your suggestion; fenced by `Arc::ptr_eq` on the locks for `alias/new.txt` vs `real/new.txt` through a symlinked directory.

Gates at head: full suite ✅ (no flake recurrence this run), workspace all-targets ✅, conformance ✅, live e2e vs d4f ✅ (third clean run).

PR #753, opened 2026-07-16
Fix config recreation over tombstones

### LINKED ISSUE #706: Timestamp-less unique config rows (ToolSelection et al) cannot converge an identical-content recreate over a tombstone

jackzampolin:
Closes #706.

## What changed

- add an indexed `updated_at` recreation identity to the six apply-managed config schemas that previously had no mutable identity field
- stamp only add/recreate mutations, leaving update mutations stable
- share the recreation helper across CLI apply and runtime Task/Schedule/EventTrigger writers
- classify `updated_at` as protected in the Lean self-config contract and prove that every managed target exposing it protects it
- cover generic overrides, custom automation mutations, embedded DefraDB delete/recreate behavior, and the full CLI create → prune → identical re-add flow

This is a clean extraction of the #706 fix from the stale/conflicting #707 branch, rebased onto current `main`. It is independent of the DefraDB pin in #752.

## Validation

- `lake build`
- `cargo test -p defra-agent`
- `cargo test -p defra-agent-cli recreate -- --nocapture`
- `cargo test -p defra-agent-cli generic_override_stamps_only_the_add_branch`
- `git diff --check`

### DIFF
--- a/crates/defra-agent-cli/src/commands/config/profile.rs
+++ b/crates/defra-agent-cli/src/commands/config/profile.rs
@@ -3,6 +3,7 @@ use defra_agent::graphql::escape_graphql_string;
 use serde_json::json;
 
 use crate::cli::*;
+use crate::config_writes::mint_recreate_identity_timestamp;
 use crate::extract_mutation_doc_id;
 use crate::optional_bool_field;
 use crate::optional_f64_field;
@@ -76,6 +77,10 @@ pub(super) async fn inference_profile_set(args: InferenceProfileUpsertArgs) -> R
         optional_i64_field("retry_max_resample", args.retry_max_resample),
         optional_bool_field("retry_allow_repair", args.retry_allow_repair),
         optional_i64_field("retry_interactive_max", args.retry_interactive_max),
+        Some(format!(
+            r#"updated_at: "{}""#,
+            escape_graphql_string(&mint_recreate_identity_timestamp())
+        )),
     ]
     .into_iter()
     .flatten()
--- a/crates/defra-agent-cli/src/config_import.rs
+++ b/crates/defra-agent-cli/src/config_import.rs
@@ -183,7 +183,8 @@ async fn apply_manifest_pairing_documents(
                         doc.unique_value
                     )
                 })?;
-            let add_literal = graphql_input_literal(&doc.add_doc)?;
+            let add_doc = crate::config_writes::mint_recreate_identity(&doc.add_doc);
+            let add_literal = graphql_input_literal(&add_doc)?;
             let update_literal =
                 graphql_input_literal(doc.update_doc.as_ref().ok_or_else(|| {
                     anyhow::anyhow!("missing update document for PeerPairingDesired")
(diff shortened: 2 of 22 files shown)


jackzampolin:
Blocking review findings — this should remain draft until both are addressed:

1. Repeated resurrection fails. `select_existing_document` rejects more than one deleted row at `crates/defra-agent/src/config_client/common.rs:77-84`, and the batched CLI selector repeats that behavior at `crates/defra-agent-cli/src/config_import.rs:648-655`. After create → prune → recreate → prune, two terminal tombstones exist for the logical key, so the next recreate errors before minting a new identity. The regression at `crates/defra-agent-cli/tests/cli_config_apply_local.rs:599-732` covers only one cycle. Please accept arbitrary tombstone history when no live row exists and test at least two cycles.

2. Imperative writers remain tombstone-blind. `config tools rm` → `tools set` still uses an add branch without `updated_at` in `config_client/tool_selection.rs:17-53`. The same gap exists for backend set in `config_client/inference_backend.rs:75-117` and profile set in `defra-agent-cli/src/commands/config/profile.rs:112-129`. Identical remove→set therefore regenerates a terminal content-addressed docID. These paths, including init callers, need the shared recreate identity behavior and direct regressions.

Residual risk: `mint_recreate_identity` promises a fresh identity using wall clock alone; consecutive equality or clock rollback is not modeled or tested.

Lean, formatting, diff checks, and all published CI pass. Schema upgrade coverage, GraphQL escaping, and empty-list handling looked correct.

jackzampolin:
Addressed the review blocker in the follow-up commit. Live rows are queried independently from tombstone history, historical lookup is bounded per logical key, and all imperative add paths mint a fresh monotonic `updated_at` identity. Added repeat prune/recreate coverage and server-backed remove→identical-set regressions for tool selections, inference backends, and inference profiles. Verification: `lake build Proofs.SelfConfig`; focused runtime/CLI resurrection tests; `cargo test -p defra-agent -- --test-threads=1` (1,152 unit tests plus all integration/conformance targets); `cargo check --workspace --all-targets`; format and diff checks—all green.

PR #754, opened 2026-07-17
fix(llm): rig pin bump — Responses tools keep optional parameters (#542)

### LINKED ISSUE #542: rig hardcodes strict:true on Responses tools + rewrites 'required' (breaks tools with optional params)
## Problem
The pinned rig fork forces strict function-calling on every OpenAI Responses tool and mangles the schema:
- \`providers/openai/responses_api/mod.rs\` \`ResponsesToolDefinition::function()\` sets \`strict: true\` **unconditionally**.
- \`providers/openai/mod.rs\` \`sanitize_schema()\` **overwrites** each tool's \`required\` with *all* property keys.

For a tool with optional params (defra-agent's \`defra_query\` has optional \`filter\`/\`limit\`), the resulting schema is rejected by OpenAI strict validation:
\`\`\`
Invalid schema for function 'defra_query': ... 'required' ... Extra required key 'filter' supplied.
(code: invalid_function_parameters, param: tools[6].parameters)
\`\`\`
Only \`defra_query\` failed because the other 6 default tools have all-required schemas (so the rewrite is idempotent).

## Ground truth
The real OpenAI Codex CLI sends \`strict: false\` for all tools (\`codex-rs/tools/src/responses_api.rs\`: \`strict: false\`), and the Codex backend does not require strict mode.

## Current state
Worked around for **ChatGptCodex only** in #530 (d03ec7cd): the Codex request-body patch forces \`strict: false\` on every tool. But the underlying rig behavior still affects the **hosted-OpenAI Responses** path (non-Codex), where any tool with optional params would break.

## Asks
- In the rig fork (sourcenetwork/rig): make \`strict\` configurable (don't hardcode true), and fix \`sanitize_schema\` to not blindly overwrite \`required\` (or only do so when strict is on AND make optionals nullable per OpenAI's structured-outputs rules). Then bump the pin.
- Alternatively, author all defra-agent tool schemas strict-compliant (all-required + nullable optionals via schemars) so they survive strict mode.
- Ties to the rig-removal epic #438 (the provider/streaming seam) and the multi-provider audit #509.

iverc:
Bumps the `sourcenetwork/rig` fork to `1ebeeae`, which stops hardcoding `strict: true` on Responses-API tools and stops rewriting `required` to all properties — tools with optional parameters were silently broken on that path (#542).

Fence: `responses_capture_preserves_optional_tool_parameters` in `rig_compat.rs` asserts the rendered tool JSON keeps `required: ["path"]` and carries no `strict: true`.

Gates: full `cargo test -p defra-agent` green (all suites), `cargo check --workspace --all-targets` clean (two pre-existing `failing_peers` warnings, untouched), `cargo fmt --all --check` clean.

Closes #542.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17790,7 +17790,7 @@ dependencies = [
 [[package]]
 name = "rig-core"
 version = "0.35.0"
-source = "git+https://github.com/sourcenetwork/rig?rev=81c34131627e1331f02d36bd12742155d1fa9865#81c34131627e1331f02d36bd12742155d1fa9865"
+source = "git+https://github.com/sourcenetwork/rig?rev=1ebeeaee3b569bb6e242e000c11df1ed2c0cca90#1ebeeaee3b569bb6e242e000c11df1ed2c0cca90"
 dependencies = [
  "as-any",
  "async-stream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ tungstenite = { git = "https://github.com/openai-oss-forks/tungstenite-rs", rev 
 # as `response.reasoning_text.delta`, which rig has no variant for (also a
 # `reasoning` vs `reasoning_content` mismatch on Chat Completions). Without the
 # fork our Responses-API reasoning capture is silently empty.
-rig-core = { git = "https://github.com/sourcenetwork/rig", rev = "81c34131627e1331f02d36bd12742155d1fa9865" }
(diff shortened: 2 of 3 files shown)


jackzampolin (approved):
Approved after isolated-worktree review.

No blocking findings. The rig pin advances one narrowly scoped commit, and the regression in `rig_compat.rs` reaches the real Responses `CompletionRequest::try_from` plus serialization path. It verifies optional parameters remain optional and `strict: true` is absent. The synthetic merge against current `main` is conflict-free and changes only `Cargo.toml`, `Cargo.lock`, and `rig_compat.rs`.

All published checks are green. Please refresh the base for a formally current CI run before the release merge, since the existing merge ref predates #698.

PR #755, opened 2026-07-17
polish(desktop): chat turn feedback — thinking indicator, load skeleton, reliable auto-follow

iverc:
First WS3 (chat) slice of #608 — the three feedback gaps that made the chat feel inert:

- **Thinking indicator**: an animated placeholder assistant turn renders whenever the turn is non-terminal and the assistant has produced no visible output yet (previously the transcript sat motionless between send and first token).
- **Conversation-switch loading skeleton**: opening an existing conversation showed the "Send the first message" empty state while the snapshot loaded — factually wrong copy. A selected-but-loading session now gets skeleton rows; the empty state is reserved for no-session.
- **Reliable auto-follow**: follow scrolling is now `behavior: "instant"` — the CSS smooth-scroll animated each programmatic scroll, and a chunk landing mid-animation left the panel short of bottom, which the position heuristic misread as the user scrolling away (silently disengaging follow mid-stream). A fresh user send also force-re-engages follow.

All animation uses the motion tokens; skeleton/dots use existing color tokens (ratchets unchanged). Fences: `chat-transcript-states.test.tsx` (6 tests incl. the instant-scroll call shape). Unit 200, e2e 120, visual 3×3 green.

Part of #608 (WS3).

### DIFF
--- a/apps/desktop-tauri/src/components/chat/ChatTranscriptPanel.tsx
+++ b/apps/desktop-tauri/src/components/chat/ChatTranscriptPanel.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from "react";
 
+import { isTerminalTurnState } from "../../lib/chat-shell";
 import type { DesktopSessionSnapshot } from "../../lib/types";
 import { MessageList } from "../Transcript";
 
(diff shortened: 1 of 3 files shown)


jackzampolin (changes requested):
Blocking auto-follow finding from the isolated-worktree review:

Re-engagement at `apps/desktop-tauri/src/components/chat/ChatTranscriptPanel.tsx:62-71` depends on observing a transient `pendingUserTurn`. The next snapshot can already contain the materialized `userMessage`; the snapshot path explicitly removes pending once materialized. If the user had scrolled up, `autoFollowTranscript` stays false and every subsequent transcript-signature update returns early at lines 73-76, so the response streams off-screen.

The new test starts with follow enabled on initial mount, so it proves only the instant `scrollIntoView` call, not disabled→enabled re-engagement. Please key re-engagement to the submitted request/user turn in a way that survives pending-row elision, and add a test that first disables follow and then receives an already-materialized send. A retry with the same content should also be distinguished by request identity.

Focused tests, build, formatting, and desktop Playwright all passed.

jackzampolin (approved):
Re-reviewed after the follow-up commits. Auto-follow is now keyed to turn/request identity, including already-materialized identical retries. Local build, focused tests (8/8), and desktop Playwright suite (21/21) pass.

PR #756, opened 2026-07-17
polish(desktop): transcript copy actions + error-card retry

iverc:
Second WS3 (chat) slice of #608:

- **Copy everywhere it matters**: hover-reveal copy on user/assistant message cards (copies the raw markdown source) and a per-code-block copy button via a `pre` renderer override. Clipboard helper prefers `navigator.clipboard` with a hidden-textarea `execCommand` fallback for WKWebView permission quirks — zero new dependencies.
- **Failed turns are recoverable**: the error card now leads with plain language, keeps the raw provider text behind a disclosure, and offers one-click **Retry** that re-submits the failed turn's content verbatim (send logic extracted to a shared `submitContent`, no duplication).

The spacing ratchet caught its own first draft (a raw `1px` padding) — fixed to tokens; ratchets unchanged. Fences: `chat-copy-retry.test.tsx` (clipboard fallback chain, message + code-block copy, retry wiring incl. no-handler case). Unit 205, e2e 120, visual regenerated for the flex message-role row.

Stacked on #755. Part of #608 (WS3). Syntax highlighting is deliberately the next slice — it needs a bundled-highlighter decision (rehype-highlight vs shiki; CSP forbids CDNs).


jackzampolin (changes requested):
Two findings from the isolated-worktree review:

1. Blocking: Retry is normally nonfunctional. `submitContent` checks the precomputed `shellProjection.sendStatus` at `apps/desktop-tauri/src/hooks/desktopShellChatActions.ts:46-54`. That projection was computed from the composer draft, not the retry content. The accepted original send clears the draft, so Retry normally reports “Type a message to send” and never calls `sendChatMessage`. The new test mocks `onRetryMessage` at the panel boundary and misses the shell/action integration. Please validate the supplied retry content or recompute the projection, with an integration test starting from an empty draft.

2. Accessibility/cleanup: the WKWebView fallback at `apps/desktop-tauri/src/lib/clipboard.ts:13-25` selects a hidden textarea, taking focus from the keyboard-triggered copy control, then removes it without restoring focus. If `execCommand` throws, cleanup is skipped and hidden textareas accumulate. Please restore prior focus and put removal in `finally`.

Build, formatting, focused tests, and visual Playwright pass. This PR is stacked on #755 and should be rebased after that blocker is fixed.

jackzampolin (approved):
Re-reviewed after the follow-up commits. Retry validation is independent of the empty composer draft, and clipboard fallback now cleans up and restores focus. Local build, focused tests (16/16), chat-shell/Lean-contract tests (17/17), and visual Playwright tests (3/3) pass.

PR #759, opened 2026-07-17
polish(desktop): message differentiation + reading measure

iverc:
Fifth WS3 (chat) slice of #608. User and assistant messages were visually identical full-width cards.

- **Reading measure**: message cards cap at 76ch — long assistant prose no longer spans the full panel width.
- **Differentiation**: the user's asks sit in distinct tinted cards (`--color-surface-muted`); assistant turns read as open page text (transparent, borderless), matching the ask-vs-answer rhythm of every mature chat app. Error/pending/thinking cards keep their special treatments.
- **Fence repair**: the box-shadow ratchet's `\s*` sat before its `(?!none…)` lookahead, so backtracking let `box-shadow: none` slip back into the count — the lookahead now anchors directly after the property. (Found because this slice's two legitimate `none` declarations tripped it.)

Unit 212, e2e 120, visual baselines regenerated (chat states). Stacked on #758. Part of #608 (WS3).

PR #760, opened 2026-07-17
polish(desktop): interrupt toast + tool-row chrome

iverc:
Sixth WS3 (chat) slice of #608:

- **Interrupt feedback is a toast**: results render as a pill overlaying the chat column (no layout reflow on appear/dismiss), errors styled as errors instead of muted small text, copy humanized ("Interrupted" / "Couldn't interrupt: …" — no raw timestamp). The two e2e assertions pinned to the old copy updated to the toast testid.
- **Tool rows tell you what they did**: collapsed summaries now carry a one-line args digest (query/path/first arg, 64-char cap, mono), so operators stop expanding every row. The redundant "Tool Calls — N tool calls" group header is gone, and the static "View" label is a chevron that rotates on open (motion token).

Unit 212, e2e 120, visual 3×3 green (unchanged baselines). Stacked on #759. Part of #608 (WS3).


jackzampolin:

PR #761, opened 2026-07-17
polish(desktop): humanize cancel-cause + cascade dialog copy

iverc:
Seventh WS3 (chat) slice of #608. The cancel/interrupt surfaces spoke in raw protocol terms.

- **CancelCauseDetails**: leads with a plain-language headline (e.g. "Interrupted · via a cascade from the parent request · 3:14:07 PM") mapping the cause/source enums to operator language; the raw derivation (cause, confidence, source, evidence rows) moves behind a "Technical details" disclosure for bug reports.
- **CascadeCancelDialog**: group headings and the confirm button are now human ("Will be interrupted", "Will keep running detached", "Already finished", "Interrupt all"); the unknown-policy warning drops `cancel_policy`/bridge-row jargon.

Test copy assertions (2 unit, 1 e2e) updated to the new strings. Unit 212, e2e 120, visual 3×3 green. Stacked on #760. Part of #608 (WS3).

PR #762, opened 2026-07-17
feat(desktop): slash-skill autocomplete in the composer

iverc:
Eighth WS3 (chat) slice of #608 — send-time skill selection, with zero backend changes.

The runtime already consumes leading `/skill-id` lines as skill selection (`skills::prompt_slash_skill_selection`, wired through every desktop submit path) — the desktop just never surfaced it. Now:

- Typing `/` on a leading line opens a skills menu (deployment's enabled skills, filtered by the typed prefix, capped at 8). Arrow keys navigate, Enter/Tab select (send is suppressed while the menu is open), Escape dismisses, click works.
- Selection inserts the `/skill-id` selector line exactly as the runtime parses it; the suggestion logic mirrors the parser's leading-block rule (only blank/selector lines above) so the menu never appears where a selector wouldn't be honored.
- The idle footer advertises the affordance ("⏎ send · ⇧⏎ new line · / skills") when the deployment has skills.

Fences: `slash-skills.test.tsx` (suggestion block rules, prefix filter, disabled-skill exclusion, body-preserving insertion, Enter-selects-not-sends, footer hint). Unit 218, e2e 120, visual regenerated (narrow footer).

Stacked on #761. Part of #608 (WS3). This closes the WS3 chat list except per-message timestamps (needs bridge plumbing).


jackzampolin:

PR #764, opened 2026-07-17
polish(desktop): readable read-tool blocks

iverc:
WS3 Code-mode slice of #608. `read_file`/`grep`/`glob`/`list_files` results dumped the raw `defra_fs` envelope JSON at operators through the generic disclosure.

They now project into collapsed read blocks: a summary line with the tool verb, target path/pattern, and honest counts ("120 of 400 lines · truncated", "12 matches"), and a copyable mono body with the JSON head stripped. Collapsed by default — reads are bulky and secondary — with the same rotating-chevron affordance as other tool rows. The projection follows the established honesty rule: no parseable metadata → generic disclosure, never a fabricated result.

Fences: `read-tools.test.tsx` (projection per tool, args-pattern fallback, no-metadata honesty, collapsed default, count label, copy). Unit 226, e2e 120, visual unchanged.

Stacked on #763. Part of #608 (WS3 code-mode).

PR #765, opened 2026-07-17
polish(desktop): honest write_file overwrite rendering

iverc:
WS3 Code-mode slice of #608. A `write_file` onto an existing file rendered as an all-additions diff labeled "edited" — implying pure addition when the previous contents were destroyed and are unknowable client-side.

Overwrites (envelope `created: false`) now label as **overwrote** and render the new contents as a plain copyable block with an explicit note ("Replaced the file's previous contents — no diff basis exists"), instead of a fabricated `+` gutter. Created files keep the honest all-add diff; `edit_file` keeps its real del/add diff. Legacy envelopes without the `created` field keep today's rendering.

Fences: projection flags + render assertions in `code-tools.test.tsx`. Unit 227, e2e 120, visual unchanged.

Stacked on #764. Part of #608 (WS3 code-mode).


jackzampolin:

PR #766, opened 2026-07-17
feat(desktop): stuck-work attention badge + diagnostics strip

iverc:
WS6 slice of #608. The bridge computes stuck-work diagnostics on every operations snapshot, but nothing rendered them — the drawer could not alert an operator to the very conditions it was built to surface.

- **Attention badge**: a workspace-level snapshot poll (the panels only poll while mounted in the open drawer) badges the collapsed rail handle and the Background tab with the count of stuck/at-risk items; the handle's aria-label says "N items need attention".
- **Diagnostics strip**: the Background panel leads with the bridge's diagnoses in operator language ("bash on req_0123456789… has stopped making progress", "Request … ran past its deadline", remote cancel-ack waits), warning/critical toned, with the pulsing-dot motion token.
- The live-backend contract mock gained the operations-snapshot route the new poll exercises, and the operations-rich spec's `getByText("cargo test")` locator became ambiguous **because the fixture's diagnostic now actually renders** — tightened to the exact gridcell.

Fences: `ops-attention.test.tsx` (operator sentences, badge + aria at count>0, quiet handle at 0). Unit 230, e2e 120, visual unchanged.

Stacked on #765. Part of #608 (WS6).

PR #767, opened 2026-07-17
polish(desktop): honest ops panel chrome

iverc:
WS6 slice of #608 — three honesty fixes in the Background panel:

- **Filter chips are data-driven**: state/await chips derive from the rows actually present (plus any engaged filter so clearing stays possible) — an offered filter always matches something, and the hardcoded chip lists (including an "Await: bridge" option no view row carries) are gone.
- **A failed poll no longer nukes the table**: the full error screen is reserved for having nothing to show; with a last-good snapshot the panel keeps rendering it under a "Live updates interrupted — showing the last snapshot" note.
- **Honest counts**: "N live / 8 max" claimed a hardcoded capacity the runtime never confirmed and counted *filtered* rows against it. Now: "N backgrounded · M shown" (the shown suffix only when filters hide rows).

The operations-rich spec's "2 live" assertion updated to the new count testid. Unit 230, e2e 120, visual baseline regenerated (ops drawer state).

Stacked on #766. Part of #608 (WS6).


jackzampolin:

PR #768, opened 2026-07-17
polish(desktop): visible fleet health + DID copy

iverc:
WS4 slice of #608. Agent health was a 10px hover-only dot with a raw pipe-joined tooltip, and failing rows showed no visible error.

- **Visible status labels**: every row shows Online / Syncing / Not connected / Error next to the dot (tones mapped; detailed states stay in the tooltip). Error takes precedence over connectivity.
- **Failing rows explain themselves**: a visible error line renders under the agent identity, run through the existing peer-connection formatter so operators get remediation-aware copy instead of raw transport errors.
- **DID copy affordance**: hover copy on the agent identity (the raw DID was prominent but uncopyable).

Fences: `fleet-health-visibility.test.tsx` (label mapping, error-line render + precedence, copy affordance). Unit 233, e2e 120, fleet visual baselines regenerated.

Stacked on #767. Part of #608 (WS4).

PR #770, opened 2026-07-17
feat(desktop): config dirty-state tracking (slice 1)

iverc:
WS5 slice of #608. Config editors gave no indication of unsaved changes.

**Design — timing-free by construction**: each editor expresses its view→form hydration as a pure function, used both by the id-keyed reset effect and to compare current form values against the persisted document (`isDirty` = deep compare). No baseline capture racing hydration effects; after a save the snapshot refresh updates the view and the comparison heals itself — and a backend that normalizes values keeps the editor honestly marked dirty until the form matches what's actually persisted.

- `ConfigEditorHeader` gains an amber **Unsaved changes** chip (takes precedence over Saved; token-only styling, no new raw literals).
- Wired for **Skill** and **Backend** in this slice; the remaining eight editors follow the identical mechanical pattern in slice 2.

Fences: `config-dirty-state.test.tsx` (clean → dirty on edit → clean when the view catches up, both editors). Unit 235, e2e 120, visual unchanged.

Stacked on #768. Part of #608 (WS5).

### DIFF
--- a/apps/desktop-tauri/src/components/config/BackendConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/BackendConfigPanel.tsx
@@ -6,6 +6,7 @@ import type {
   DeploymentView,
   InferenceBackendView,
 } from "../../lib/types";
+import { isDirty } from "./configDirty";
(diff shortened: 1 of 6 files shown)


jackzampolin:

PR #771, opened 2026-07-17
feat(desktop): config dirty-state tracking (slice 2)

iverc:
WS5 continuation of #770 — the same timing-free pattern (pure view→form hydration function shared by the id-keyed reset effect and the `isDirty` comparison, amber Unsaved-changes chip) applied to five more editors:

- **InferenceProfile**, **ToolService**, **Task**, **Schedule**, **EventTrigger** — including the secondary hydration inputs (fallback task id for Schedule/EventTrigger, fallback behavior id for Task) threaded through the pure functions so the baselines are exact.
- Non-form state (test results, run status) stays outside the dirty comparison; the Agent panel is deferred to the one-editing-paradigm item since it has no ConfigEditorHeader.

Fence extended with the schedule editor case. Unit 235 (all suites), e2e 120, visual unchanged.

Remaining for slice 3: Behavior and ToolSelection (their hydration has derived inputs — profile-set fallback and known-service migration — needing careful extraction).

Stacked on #770. Part of #608 (WS5).

### DIFF
--- a/apps/desktop-tauri/src/components/config/EventTriggerConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/EventTriggerConfigPanel.tsx
@@ -7,6 +7,7 @@ import type {
   EventTriggerView,
   TaskView,
 } from "../../lib/types";
+import { isDirty } from "./configDirty";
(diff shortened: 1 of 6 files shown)


jackzampolin:

PR #772, opened 2026-07-17
feat(desktop): config dirty-state tracking (slice 3 — ToolSelection)

iverc:
WS5 continuation of #770/#771 — the timing-free dirty pattern applied to the largest editor, **ToolSelection** (22 fields), with the derived hydration inputs handled exactly: the pure `toolSelectionFormValues(toolSelection, toolServiceIdKey)` reproduces the legacy service-delegate migration (delegates matching known service ids fold into the MCP allowlist) so the baseline matches the reset effect bit-for-bit.

This completes dirty tracking for all eight header-based editors. Behavior and Agent use the inline-title editing paradigm with no `ConfigEditorHeader` — they get the chip as part of the WS5 "one editing paradigm" item rather than bolting a header onto the divergent layout.

Unit 236, e2e 120, visual unchanged. Stacked on #771. Part of #608 (WS5).

### DIFF
--- a/apps/desktop-tauri/src/components/config/ToolSelectionConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/ToolSelectionConfigPanel.tsx
@@ -7,6 +7,7 @@ import type {
   ToolSelectionView,
   ToolServiceRegistryView,
 } from "../../lib/types";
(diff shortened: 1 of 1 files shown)


jackzampolin:

PR #773, opened 2026-07-17
fix(desktop): keyboard-accessible config tabs

iverc:
WS5 slice of #608. The config tab strip used a roving tabindex with no arrow-key handling — the ARIA tabs contract without its keyboard half, leaving the ten tabs mouse-only.

The tablist now implements the WAI-ARIA tabs pattern: Left/Right cycle with wrap-around, Home/End jump to the edges, selection and focus move together, and the saved-status reset matches click behavior.

Fence: `config-tabs-a11y.test.tsx` (cycle, wrap, Home/End, roving tabindex flip). Unit 237, e2e 120, visual unchanged.

Stacked on #772. Part of #608 (WS5).

### DIFF
--- a/apps/desktop-tauri/src/components/ConfigWorkspace.tsx
+++ b/apps/desktop-tauri/src/components/ConfigWorkspace.tsx
@@ -165,7 +165,32 @@ export function ConfigWorkspace({
         </div>
       </header>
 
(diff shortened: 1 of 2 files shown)


jackzampolin:

PR #774, opened 2026-07-17
feat(desktop): visible validation reasons in config editors

iverc:
WS5 slice of #608. Invalid input silently disabled Save with zero explanation.

Every validity flag that gates a Save button now renders an inline reason under its field (`FieldHint`, `role="alert"`, error tone): 14 fields across seven editors — Backend (max concurrent/queue depth), InferenceProfile (context window, output tokens, turns, temperature, stream batch/liveness, deadline), Schedule (interval), ToolService (port range), ToolSelection (spawn timeout), Task (run-args JSON), Behavior (compaction threshold 0–1). Messages state the constraint ("Whole number of 1 or more", "Port between 1 and 65535"), appear on invalid input, and disappear when fixed.

Fence: `config-validation-hints.test.tsx` (absent → present with constraint text → absent again). Unit 238, e2e 120, visual unchanged.

Stacked on #773. Part of #608 (WS5).

PR #775, opened 2026-07-17
fix(desktop): ops-panel liveness consistency

iverc:
WS6 slice of #608. The four ops tabs had wildly inconsistent liveness: Background polled at 2s, MCP at 10s, while **Backend health** was a one-shot whose relative ages froze at mount and **Lineage** was a one-shot snapshot — static during the very live turn it exists to observe.

- **Backend health** now polls at 10s (matching MCP); background refreshes skip the loading flip so the table doesn't flicker, and relative ages re-render live.
- **Lineage** polls at 5s while mounted. Background refreshes preserve the operator's expand/collapse choices and auto-expand only genuinely new nodes (a new subagent spawning appears expanded; nothing the operator collapsed reopens). Transient background poll failures keep the last-good tree instead of blanking the panel.

Fences: `ops-liveness.test.tsx` (both panels keep calling under fake timers; lineage tree persists across refreshes). Unit 240, e2e 120, visual unchanged.

Stacked on #774. Part of #608 (WS6).

### DIFF
--- a/apps/desktop-tauri/src/components/backendHealth/useBackendHealth.ts
+++ b/apps/desktop-tauri/src/components/backendHealth/useBackendHealth.ts
@@ -15,8 +15,10 @@ export function useBackendHealth(): BackendHealthState {
   const [loading, setLoading] = useState<boolean>(true);
   const [error, setError] = useState<string | null>(null);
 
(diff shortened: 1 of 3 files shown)


jackzampolin:

PR #776, opened 2026-07-17
polish(desktop): short request ids in the ops drawer

iverc:
WS6 slice of #608. Full-length request ids made the parent filter chips, the Parent column, and the lineage tree unreadable at drawer width.

A shared `shortId` helper (14-char prefix + ellipsis) now renders everywhere ids appear in the drawer — parent chips, Parent cells, lineage node labels — with the full id preserved in the `title` tooltip. The diagnostics strip already used the same truncation; it now shares the one helper.

The operations-rich spec's full-id text locators moved to title-attribute locators (the ids are deliberately no longer full text). Unit 240, e2e 120, visual unchanged.

Stacked on #775. Part of #608 (WS6).


jackzampolin:

PR #777, opened 2026-07-17
feat(desktop): fleet-to-Code path

iverc:
WS4 slice of #608. Code mode existed but had no path from the fleet — operators had to open chat first, then find the buried entry.

Fleet rows gain a Code action (brackets glyph, between Chat and Configure) that selects the agent and opens Code mode directly, matching the existing action-button pattern with a proper aria-label and testid. The dashboard/App threading mirrors `onOpenChat`.

Fence: click-through test in `fleet-health-visibility.test.tsx`. Unit 241, e2e 120, fleet baselines regenerated (new action column width).

Stacked on #776. Part of #608 (WS4).

PR #778, opened 2026-07-17
feat(desktop): save errors near the form

iverc:
WS5 slice of #608 — the last of the config-editor UX items. Save failures surfaced only as a raw `String(err)` in the global banner at the top of the app, disconnected from the form the operator was editing.

All eight header-based editors now catch the (already-rethrown) save failure locally and render "Save failed: {message}" directly under the editor header — cleared on the next successful save and on document switch. The global banner behavior is unchanged (shell handlers still set it), so nothing regresses for surfaces that rely on it; the form just stops being silent. The now-unused `ignoreHandledActionError` imports came out of the five editors whose only catch was the save path.

Fence: skill-editor save-failure case in `config-dirty-state.test.tsx`. Unit 241, e2e 120, visual unchanged.

Stacked on #777. Part of #608 (WS5).


jackzampolin:

PR #779, opened 2026-07-17
feat(desktop): one editing paradigm for Agent/Behavior editors

iverc:
WS5 slice of #608. The Agent and Behavior editors were the last two outside the shared editor paradigm: no dirty tracking, and save failures vanished into the global banner.

- Shared `EditorStatusChip` extracted from `ConfigEditorHeader` — one save-state vocabulary (amber "Unsaved changes" outranks green "Saved") across all ten editors, including these two custom-header panels.
- Both editors get the pure view→form baseline dirty comparator and local "Save failed: …" routing; Cancel on an agent rename clears the abandoned edit's error.
- The Behavior profile fallback is now derived instead of stored. The old fallback effect baked its pick into state, which adversarial review showed both defeats dirty tracking after profile-list churn and — worse — silently persists the fallback over the document's own `inferenceProfileId` on Save. Deriving the effective profile (state → document → first) removes the divergence at the root and restores the document's selection when its profile registers late.

Fences: skill toggle round-trip is not an edit, document profile beats positional first, drop-out/reappear churn stays clean, cancelled rename clears the failure. Unit 250, e2e 120, visual unchanged.

Stacked on #778. Part of #608 (WS5).

### DIFF
--- a/apps/desktop-tauri/src/components/config/AgentConfigPanel.tsx
+++ b/apps/desktop-tauri/src/components/config/AgentConfigPanel.tsx
@@ -8,8 +8,7 @@ import type {
   BootstrapSummary,
   DeploymentView,
 } from "../../lib/types";
-import { PencilIcon } from "./ConfigChrome";
-import { ignoreHandledActionError } from "./formUtils";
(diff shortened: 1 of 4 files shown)


jackzampolin:

PR #780, opened 2026-07-17
feat(desktop): dismissible error banner

iverc:
WS2 slice of #608. The global error surface was a bare div dumping `String(err)` with no way to clear it short of a successful retry.

The banner is now a proper alert: error icon, wrapping message (long GraphQL errors no longer distort the layout), a Copy-error affordance for bug reports, and a dismiss button (`useDesktopShell.onDismissError`). Same `error-banner` testid and message text, so every existing sad-path assertion holds; e2e additionally fences dismiss, and the bridge-error visual baseline is regenerated for the new layout.

Stacked on #779. Part of #608 (WS2).

PR #781, opened 2026-07-17
refactor(desktop): token sweep — zero raw rgb() literals

iverc:
Groundwork slice of #608 for the upcoming light theme: every color the desktop client paints must come from `tokens.css` before a second theme can override it.

All 89 literal `rgb()` sites across 18 stylesheets now reference token triplets (`rgb(var(--x-rgb) / α)` — alphas stay at use sites so a theme only overrides the triplets). New tokens cover the recurring families the sweep exposed: white/black overlay bases, the near-black green inks, and rgb forms of the existing error/muted/accent/success colors; two one-off error tints became hex tokens.

Zero visual change, proven two ways: the visual suite passes against **unchanged** baselines, and a whole-file check (forward-map HEAD's content, compare whitespace-normalized) shows every substitution is exactly value-preserving. The conformance ratchet drops from ≤90 to =0 so literals cannot return.

Stacked on #780. Part of #608 (WS2, theme groundwork).

### DIFF
--- a/apps/desktop-tauri/src/styles/backend-health/backend-health.css
+++ b/apps/desktop-tauri/src/styles/backend-health/backend-health.css
@@ -60,7 +60,7 @@
     border-color: rgb(var(--source-orange-rgb) / 0.36);
   }
   .backend-health__summary-chip[data-tone="error"] {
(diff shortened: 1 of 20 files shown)


jackzampolin:

PR #782, opened 2026-07-17
feat(desktop): light theme

iverc:
WS2 slice of #608, on top of the #781 token sweep.

- **Light theme**: `:root[data-theme="light"]` token set in tokens.css (green-tinted paper surfaces, darkened text/error/success ramps, brand green kept). The overlay/scrim triplets got semantic names (`--overlay-rgb`/`--scrim-rgb`, ex `--source-white-rgb`/`--source-black-rgb`) so inverting them in light is meaningful, and the brand logo tile is pinned to a theme-stable dark backdrop (the white glyph was invisible on light). A conformance fence rejects light overrides that don't redefine a dark-root token.
- **Toggle + persistence**: sun/moon `ThemeToggle` in the fleet header; choice persists via localStorage and is stamped by `AppShell` on boot (App's outer function stays hook-free for the boundary test). Dark remains the default.
- **Fleet status overlap fix**: the agent cell's grid column was still sized for the old 14px dot, so the WS4 "Online" label painted over the DID. Column is `auto` now.
- **Visual fence hardening**: that overlap shipped invisibly because `maxDiffPixelRatio: 0.01` tolerates ~13k differing pixels — the fleet baselines on main are still #550-era. Tolerance is now `maxDiffPixels: 64` and every baseline is regenerated from current code (21 files, including 3 new light captures; the light capture runs last in the spec because the toggle persists into localStorage and would repaint subsequent captures).

Unit 254 (theme round-trip + persistence fences), e2e 120, visual 3/3 at the tight tolerance.

Stacked on #781. Part of #608 (WS2).


jackzampolin:

PR #783, opened 2026-07-17
feat(desktop): per-message timestamps

iverc:
WS3 bridge slice of #608. Message rows persisted a timestamp all along (`MessageView.timestamp`), but the rendered timeline dropped it, so operators had no idea when anything happened without opening the trace CLI.

The bridge now threads `timestamp` through `RenderedTimelineItem::{UserMessage,AssistantMessage}`, and `MessageList` renders it as a right-aligned `<time>` in the role row — time-of-day for today, dated (`Jun 3 09:05 PM`) otherwise, raw ISO on hover, omitted entirely when missing or unparsable. The harness seeds a fixed past timestamp so the label is pinned by the chat visual baselines (regenerated ×3); dynamic send flows stamp real times for e2e.

Bridge tests 110 (correct crate is `defra-agent-desktop-tauri` — `-p defra-agent-desktop` silently runs zero), unit 256, e2e 120, visual 3/3, workspace check + fmt clean.

Stacked on #782. Part of #608 (WS3).


jackzampolin:

PR #784, opened 2026-07-17
feat(desktop): remove and rename saved deployments

iverc:
WS4 bridge slice of #608. Saved peers were append-only: a dead or mislabeled deployment sat in the fleet forever.

- **Core**: `ClientCore::rename_peer` (new) and a local-runtime guard on the existing `remove_peer` — removing the deployment that *is* this machine now fails loudly instead of stranding the UI. Remove also keeps its existing status-cleanup and "transport stays up until restart" warning.
- **Bridge**: `desktop_peer_remove` / `desktop_peer_rename` tauri commands returning the refreshed snapshot, mirrored through `desktop-api` and shell peer actions (errors land in the banner via `formatPeerConnectionError`).
- **Fleet UI**: hover pencil on the row name for inline rename (Enter commits, Escape/blur cancels, no-op commits skipped — renames the *saved label*), and a trash action with a ConfirmDialog for saved peers only; local runtime rows get no remove affordance, and the dialog states the remote agent itself is untouched.

Fences: confirm-gated remove, local-runtime hiding, rename commit/cancel; harness mocks both commands. Core tests 4, bridge tests 110, unit 259, e2e 120, visual 3/3 (fleet baselines regenerated for the new affordances), workspace check + fmt clean.

Stacked on #783. Part of #608 (WS4).


jackzampolin:

PR #786, opened 2026-07-17
feat(desktop): keyboard shortcuts

iverc:
WS2 slice of #608 — the audit's "everything requires the mouse" item.

Cmd/Ctrl+1–4 switch views (fleet/chat/code/config), Cmd/Ctrl+N starts a conversation on the selected behavior, Cmd/Ctrl+K jumps to the composer (switching to chat first), and Cmd/Ctrl+/ opens a small shortcut-reference dialog (Escape/backdrop closes). One window listener installed once, handlers on a ref; chords with Shift/Alt pass through untouched so text editing is unaffected.

Fences: unit hook mapping + dialog behavior; a Playwright journey (via the shared desktop fixture — the fixture guard rejects bare `@playwright/test` imports) drives the real chords. Unit 264, e2e 123, visual 3/3 unchanged, prettier clean.

Stacked on #785. Part of #608 (WS2).

PR #787, opened 2026-07-17
feat(desktop): sidebar agent switcher

iverc:
WS3 slice of #608 — two audit items: no way to switch agents from the chat sidebar, and `SavedPeersSection` shipping as dead code with unreachable testids.

The Connected Peer card's title becomes a styled select when more than one deployment exists (static heading otherwise), switching agents in place; App clears the session selection on switch so the conversation list never shows another agent's session as selected. `SavedPeersSection` is deleted rather than mounted — one card that switches beats two lists that overlap.

Visual baselines untouched (the harness models a single deployment, so the switcher stays out of captures). Unit 266, e2e 123, visual 3/3.

Stacked on #786. Part of #608 (WS3).


jackzampolin:

PR #788, opened 2026-07-17
feat(desktop): automation document deletion

iverc:
First slice of the audit's "eight of nine document types can be created but never removed" gap, covering the automation documents: Task, Schedule, EventTrigger.

- **Core**: `delete_task` / `delete_schedule` / `delete_event_trigger` following the `delete_skill` blueprint — existence pre-check, GraphQL delete, then a shared tail (refresh, local prune so the UI reflects the delete immediately, mutation-error recording). These collections carry no `agent_did` (they're node-scoped by `behavior_id`/`task_id`), so deletes filter by document id — the initial agent-scoped design was wrong and is not what shipped.
- **Task guard**: deletion is blocked with a counted, actionable error while any schedule or event trigger still references the task — silently orphaning automation that keeps firing would be worse than asking the operator to detach it first.
- **Bridge/UI**: three tauri commands (snapshot-refreshing, like skill delete), shell handlers, and Delete buttons with `ConfirmDialog` in all three editors (hidden for unsaved documents); the editor stays mounted on a rejected delete so the error is actionable.

Fences: confirm-gated delete, rejection keeps the editor, no delete for new documents; harness mocks enforce the task-reference guard. Unit 269, e2e 123, visual 3/3, core 101 + bridge 110 Rust tests, workspace check + fmt clean.

Stacked on #787. Part of #608 (WS5).


jackzampolin:

PR #789, opened 2026-07-17
feat(desktop): inference document deletion

iverc:
Second slice of the document-deletion sweep (#788 covered automation docs): InferenceBackend and InferenceProfile.

Same blueprint — id-filtered GraphQL deletes (both collections are id-scoped), existence pre-checks, and the shared refresh/prune/error tail. Both types are guarded by named behavior references: deleting a backend or profile that a behavior still points at fails with the referencing behavior ids spelled out, since a dangling `inference_profile_id` would silently re-route that behavior through the first-profile fallback.

Editors get the confirm-gated Delete button (hidden for unsaved documents); harness mocks enforce the same guards. Unit 270, e2e 123, visual 3/3, Rust core 101 + bridge 110, workspace check + fmt clean.

Stacked on #788. Part of #608 (WS5). Remaining types for a final slice: ToolSelection, ToolService, Behavior (needs a default-behavior guard).


jackzampolin:

PR #790, opened 2026-07-17
feat(desktop): tool + behavior document deletion

iverc:
Final slice of the document-deletion sweep (#788 automation, #789 inference): ToolSelection, ToolServiceRegistry, and AgentBehavior. Every config document type the desktop can create can now also be deleted.

Guards, all with named referents in the error:
- a **tool selection** referenced by any behavior's `tool_selection_id` blocks;
- a **tool service** still allowed by any selection's `allowed_mcp_service_ids` blocks;
- a **behavior** blocks when it is the agent's default (deleting the request fallback would strand behavior-less requests) or while tasks reference it. The editor additionally hides Delete on the default behavior.

Same core blueprint (id-filtered deletes, existence pre-check, shared refresh/prune/error tail), tauri commands, shell handlers, confirm-gated editor buttons, harness mocks enforcing the same guards. Unit 271, e2e 123, visual 3/3, Rust core 101 + bridge 110, workspace check + fmt clean.

Stacked on #789. Part of #608 (WS5) — closes the "created but never removed" audit item.


jackzampolin:

PR #791, opened 2026-07-17
feat(desktop): real slot capacity in ops

iverc:
WS6 slice of #608 — the audit's "slot capacity is invisible / was once a hardcoded 8" item, done with real data.

The AgentRuntime document already carries `behavior_executor_capacity`, `behavior_executor_queue_depth`, and runnable/unavailable behavior counts; the bridge's `RuntimeView` simply dropped them. They're threaded through now, and the Background tools panel shows a "capacity N (· M queued)" line sourced from the selected deployment's runtime document — rendered only when the runtime actually reports capacity, so nothing is invented for older runtimes.

Unit 273, e2e 123, visual 3/3 (ops-drawer baselines regenerated for the new line), bridge tests 110, fmt clean.

Stacked on #790. Part of #608 (WS6).


jackzampolin:

PR #792, opened 2026-07-17
feat(desktop): macOS shell treatment

iverc:
WS2 slice of #608 — the native-feel item. The app shipped with the stock opaque macOS titlebar over a dark shell.

- `titleBarStyle: Overlay` + `hiddenTitle` in tauri.conf.json: the shell surface extends to the window edge with inset traffic lights, Claude-desktop style.
- A `data-tauri-drag-region` strip (token-sized: `--titlebar-height`) keeps the window draggable, and the shell reserves `--titlebar-clearance` of top padding — both gated on a `data-shell="mac"` stamp applied only when running inside Tauri on macOS (`__TAURI_INTERNALS__` + platform check), so the browser harness, e2e, and visual captures are untouched by design.
- The spacing ratchet caught the first draft's raw px paddings; the geometry now lives in tokens.

**Verification honesty**: unit-fenced classifier (no stamp in the harness, stamp under a stubbed Tauri global) and all suites green, but the actual overlay rendering needs a real `tauri dev` launch on macOS — flagging that for your visual check when reviewing, since it can't be exercised headless.

Unit 274, e2e 123, visual 3/3 (unchanged — by design), prettier clean.

Stacked on #791. Part of #608 (WS2).

### DIFF
--- a/apps/desktop-tauri/src-tauri/tauri.conf.json
+++ b/apps/desktop-tauri/src-tauri/tauri.conf.json
@@ -16,7 +16,9 @@
         "width": 1480,
         "height": 920,
         "minWidth": 1180,
-        "minHeight": 720
+        "minHeight": 720,
+        "titleBarStyle": "Overlay",
+        "hiddenTitle": true
       }
     ],
     "security": {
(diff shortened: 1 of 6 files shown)


jackzampolin:

Issue #803, opened 2026-07-19
flaky: backend_health cycle_reproduces_fleet_evidence_dead_backend_demotes_and_recovers failing on main


iverc:
Main's post-#794 run fails this single test in "Run full library and integration tests" (1151 passed / 1 failed). Desktop branches on the same base pass it, suggesting timing sensitivity rather than a v0.7.0 regression — but it is currently the only thing keeping main's checkmark red, so it deserves a look before the next release cut.

Re-enqueued main's failed job to confirm flake vs. regression; if the rerun stays red this is a real DefraDB v0.16.0 fallout instead.

iverc:
Rerun evidence: the backend_health test passed on rerun of the same commit — flake confirmed, not v0.16.0 fallout. The rerun then failed a different singleton, `server_keeps_running_when_codex_shim_port_is_taken` (port collision — shared-runner environment noise; 29 passed / 1 failed). Three consecutive main runs have each failed a different timing/environment-sensitive singleton, which points at runner-state pollution after the release merge rather than any code regression. Re-enqueued again.

iverc:
Same family: `completion_retry_tape::backend_restart_cluster_recovers` is load-sensitive — it passes its binary solo (64/64, ~10s) but intermittently times out (~16-25s, hitting the harness's wait deadline) when the full workspace test run puts multiple binaries in contention. Reproduced on origin/main's own baseline (pre-#805 features) and post-#812. The tape's request deadline appears too tight for a contended machine.

jackzampolin:
Closing as a duplicate of #743. That issue already records this exact `cycle_reproduces_fleet_evidence_dead_backend_demotes_and_recovers` failure and the broader shared `ModelsListener`/parallel-suite flake family, including subsequent recurrences.

PR #805, opened 2026-07-19
feat: live tool-output streaming

iverc:
One batched PR closing the audit's "no live streaming of command output despite the runtime capturing live output" item (#608). First runtime-crate feature of the epic; framed and verified as additive telemetry — no lifecycle transition changes, zero proof/conformance-table edits, full `cargo test -p defra-agent` green (598 tests including the Lean fences).

**Runtime**
- `partial_output_tail` (capped 4KiB) + `partial_output_seq` (monotonic total-bytes-seen) on AgentToolCall — schema + R5-style additive `patch_collection` migration (nullable adds need no Lens).
- A **demand-driven** flusher on the persistence hook: spawned only when a live-output writer is handed out, ticks every 2s, exits when the live-buffer registry empties. This shape is load-bearing — an always-on interval breaks paused-clock tape tests (tokio auto-advance leaps through the ticks and expires request deadlines; `backend_restart_cluster_recovers` failed deterministically until the flusher became demand-driven).
- Flush discipline, hardened by adversarial review: iterates the live registry's own ids (the in-flight lifecycle map's internal call ids are a disjoint key space — the first draft silently never flushed), CAS-guards on `lifecycle_state == "running"` so a straggler tick can never stamp a terminal row, and resupplies the row's DateTime columns (DefraDB re-validates the whole document on any update).

**Desktop**
- Fields threaded through the store queries (remote list kept in parity with local — the query guard test enforces it), bridge views, and the transcript: a running tool's disclosure auto-opens with a pulsing "live output" tail block.
- Harness's active-turn scenario seeds a running `defra_exec` with a tail; fences cover shown-while-running, never-on-finished, and silent-when-empty.

**Scope note**: only background-executed tools produce live output today — foreground tool execution has no output writer plumbed. That gap is visible in the code (one `writer_for` call site) and left for a follow-up.

Includes the standard `manage_view` cherry-pick (green standalone; merges cleanly with the other open PRs).

Gates: runtime 598 (full suite, twice), workspace check zero errors, desktop-core + bridge suites, fmt; desktop unit 301, e2e 123, visual 3/3 unchanged.


iverc:
Extended in place (same feature, no new PR): **foreground tools now stream too**, closing #807 — the dispatch path re-enters the tool runtime scope with a per-call writer keyed by the internal call id (which is exactly the persisted row's `tool_call_id`, so the flusher mutation needed no change); buffers are released in `on_tool_result` so the demand-driven flusher idles between calls. Rebased onto #812's main. Gates: lib 1152, full runtime+CLI suites (1623 passing; the one intermittent is the pre-existing load-sensitive tape test documented on #803 — green solo), e2e_runtime binary 64/64 under its own parallelism.

PR #864, opened 2026-07-22
feat(cli): interactive first-run backend picker for `gents init`

nasdf:
When `gents init` runs in a terminal with no backend selection, fall through to an interactive picker instead of silently defaulting to a local llama-server that may not be running. Non-interactive runs (pipes, CI, --identity-only) skip it and keep the historical default, so every existing script and test is unaffected.

The picker (crates/gents-cli/src/interactive_backend.rs) probes for a local server on :8080/:11434 and offers OpenAI key / local server / custom URL / ChatGPT-Codex subscription. Prompts go to stderr so init's stdout JSON stays clean.

Selecting the Codex subscription seeds the keyless chatgpt-codex preset and, interactively after config is persisted, offers an inline ChatGPT login. The login core is extracted from the `codex-login` command (run_codex_login) and reused against init's live node, so no second DB is opened; the standalone command is unchanged and remains the re-auth path. The inline login is best-effort: a declined/failed login or a non-terminal session leaves the `gents codex-login` next-step in place and init still exits 0. On success the redacted credential folds into init's single output JSON and the next-step drops.

### DIFF
--- a/crates/gents-cli/src/commands/codex_login.rs
+++ b/crates/gents-cli/src/commands/codex_login.rs
@@ -3,41 +3,86 @@ use codex_login::{
     run_device_code_login, run_login_server, AuthCredentialsStoreMode, AuthManager, ServerOptions,
     CLIENT_ID,
 };
-use serde_json::json;
+use serde_json::{json, Value};
 use uuid::Uuid;
 
 use crate::cli::args::CodexLoginArgs;
+use crate::config_writes::ConfigAccess;
 use crate::{print_json, resolve_agent_did, resolve_config_access};
+
+/// Inputs to a single ChatGPT/Codex OAuth login, independent of how the caller
+/// obtained the target node. Shared by the `codex-login` command and `init`'s
+/// inline first-login.
+pub(crate) struct CodexLoginOptions {
+    pub(crate) provider: String,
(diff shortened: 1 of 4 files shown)


jackzampolin (changes requested):
Two small correctness blockers before landing:

1. `interactive_backend::has_backend_arg` does not treat `provider_kind` or `openai_wire_api` as explicit backend signals. In a TTY, an invocation using either flag can still enter the picker; the picker then applies its preset/URL while leaving those explicit fields in place, producing a mixed configuration. Please suppress the picker for both fields and cover them in `each_backend_signal_is_detected`.

2. `maybe_inline_codex_login` returns `None` when an existing OAuth credential is found, but the caller derives `codex_logged_in` from `codex_login.is_some()`. A rerun therefore skips the login prompt yet still emits `gents codex-login` in `next_steps`. Please carry the existing-authenticated state separately (or return a richer outcome) and add a regression test.

The picker flow and login extraction otherwise look solid.

jackzampolin (approved):
Review blockers are resolved: explicit provider/wire flags suppress the picker, and existing Codex credentials now suppress the redundant login next step. Targeted regressions pass.

Issue #630, opened 2026-07-06
P2P sync: hub-side PushLog backpressure — 16 subagent-host pushers saturate the pending-DAG queue into a restart-resistant drop/re-push loop


jackzampolin:
## Summary

When N (~16) subagent-host peer-pairing legs push to a single coordinator (the Amy-hub topology), the hub's sync manager saturates: the pending-DAG queue hits its 1000-item cap and enters a non-converging drop/re-push loop that burns ~10 cores, kills the hub's QUIC/gossip connections to all peers within minutes, and **survives restarts** (the pushers immediately re-saturate the fresh process). Observed fleet-wide 2026-07-01→07-06 as recurring CPU waves; full incident data available.

This is the reason the amygdala fleet's steward→Amy pairing legs are currently checked in `enabled:false` (16/32 topology) — which also disables spawn **result-readback**, the multi-agent orchestration path.

## Why a dedicated tracker

The evidence currently lives only in comments on the now-closed #604 ("coordinator push leg dead" — fixed by #623). The manifests' re-enable gate reads "Re-enable with #604", which is *technically satisfied* — inviting someone to re-enable all 16 legs on the #604 fix alone and reproduce the storm. #604 fixed the coordinator→host direction; this issue is the host→coordinator fan-in saturation, a separate defect.

## Observed signatures (v0.6.2 hub, studio-1)

- `p2p::sync::coordinator::broadcast: PushLog to replicator timed out` — even to same-host peers (~360 WARNs/3h at idle with legs disabled; storm-level when enabled)
- pending-DAG `missing_count` observed at 210 with cap-1000 queue churn
- Hub QUIC path deaths (`noq_proto: PTO expired while unset`) within minutes of restart under fan-in load; downstream, every spoke's `iroh_gossip` send loop terminates and retries without backoff (separate spoke-side issue)
- ~130% steady hub CPU from the loop at 16 pushers, ~10 cores at peak saturation

## Suggested direction

Backpressure/fairness on the sync-manager intake: per-peer quotas or windowed admission rather than a global 1000-cap with drop/re-push; drop should quench the pusher (explicit nack/backoff signal) instead of triggering immediate re-push. A hub with N pushers needs admission control that converges.

## Acceptance sketch

Re-enable 16 subagent-host legs against one coordinator on real hardware; hub CPU and pending-DAG depth stay bounded at idle and under a 15-target spawn-and-readback sweep (the fleet topology spec's acceptance test).

[...]

PR #866, opened 2026-07-22
perf(p2p): narrow subagent return replication

iverc:
## Summary

- reduce the `subagent-host` return leg from the full conversation collection set to the projection consumed by coordinators
- retain `AgentRequest`, `AgentResponse`, `AgentMessage`, and `AgentToolCall` while keeping host-local session, conversation, standalone tool-result, and compaction rows local
- update the Lean model, conformance coverage, reconciliation tests, and embedded two-node replication test

## Why

Each host previously pushed eight conversation-related collections toward a coordinator even though coordinator-side completion and transcript reads consume only four. The narrower projection removes unnecessary replicated state while preserving durable child lifecycle, final response, readable transcript, and tool-call identity.

This is a targeted mitigation for #630, not a claim that the fleet saturation issue is closed. The 16-to-1 hardware acceptance run remains required.

## Validation

- `lake build Proofs.ScopeTemplates`
- `cargo test -p gents`
- `git diff --check`


jackzampolin (approved):
Lean model, conformance fence, Rust template, and embedded replication coverage all agree on the narrowed return projection. CI is green; landing.

Issue #705, opened 2026-07-13
First-class mobile clients: Codex shim remote access + fleet ops (not pairing)


jackzampolin:
## Summary

Support **remote first-class clients** (starting with DefraMobile) of surfaces we already run for CLI/desktop:

1. **Codex app-server shim** as the conversation control plane (model switch, thread list/resume/fork, structured turn/item streaming).
2. **Fleet management / operations** pieces currently concentrated in the desktop Tauri app (multi-deployment dashboard, health, open work, config entry points, operations rail).

**Out of scope for this issue:** the P2P QR pairing / bearer-claim / conversation-replication bootstrap already shipped in DefraMobile. Pairing stays as-is. This issue is about *what else* the runtime should expose (or document as already sufficient) so a phone can be a real operator surface after it is already paired.

Related mobile product issue: [sourcenetwork/defra-mobile#3](https://github.com/sourcenetwork/defra-mobile/issues/3).

## Why this belongs in defra-agent

The hard protocol work is mostly done:

- Codex shim projects Defra documents into Codex Thread/Turn/Item semantics (`codex_shim/`).
- Lean proofs + conformance + `cli_codex_shim` protocol tests treat the shim as a product surface.
- Ops HTTP already has `GET /status`, `/fleet`, `/fleet/slots`, `/sessions`, `/subagents/*`, etc.
- Desktop builds a rich `DeploymentView` over replicated docs + status/peer dial state and renders Fleet Dashboard + Operations rail.

Mobile does **not** need a second agent loop. It needs the host runtime to be **safe and discoverable** for non-loopback clients, and the **fleet/ops projection** to be usable from a phone without the Tauri shell.

### Already enough for a protocol spike

A Swift (or any) client can already:

```
initialize → model/list → config model write
thread/list | resume | start | fork
turn/start (+ steer / interrupt)
← item/* / turn/* notifications
```

against a running `defra-agent server` with the default shim on `ws://127.0.0.1:9292/`, same as `defra-agent codex`.

No pairing changes required for that spike on LAN loopback / simulator → host.

### Gaps for real mobile use (beyond pairing)

| Gap | Today | Why mobile cares |
| --- | --- | --- |
| **Transport auth** | Shim has **no** transport authentication; server warns on non-loopback bind | Phone will not stay on loopback; trusted private bind alone is weak for Tailscale/LAN phones |
[...]

PR #867, opened 2026-07-22
Authenticate remote app-server clients

iverc:
## Summary

- add environment-sourced bearer authentication to the app-server WebSocket listener and embedded terminal client
- refuse unspecified listeners and unauthenticated non-loopback binds
- advertise the dialable WSS URL, authentication requirement, and bound agent/behavior through startup readiness and `/healthz`
- document TLS reverse-proxy operation and cover missing, incorrect, and valid credentials with a real process-level handshake test

## Security model

- tokens are read from named environment variables and are never placed in command arguments, readiness output, health payloads, or logs
- advertised public endpoints must be root `wss://` URLs with an explicit port
- loopback use remains backward compatible without authentication

## Validation

- `cargo test -p gents-cli`
- `git diff --check`

Related to #705. The wider thin-client and desktop transport migration remains separate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8665,6 +8665,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "subtle",
  "tempfile",
  "tokio",
  "tokio-tungstenite 0.26.2",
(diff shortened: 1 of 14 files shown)


jackzampolin (approved):
Reviewed the listener policy, environment-only token handling, constant-time bearer check, WSS/loopback client guard, readiness/health redaction, documentation, and process-level handshake coverage. All checks are green; landing.

Issue #714, opened 2026-07-14
Machine index service: host-level agent discovery, creation, and batch pairing


noahsaso:
## Summary

Introduce a **machine index service**: a Rust service that runs on every machine hosting defra agents (alongside the runtime, in the spirit of a background "remote control" daemon) and acts as the host-level control plane for **discovery, creation, and pairing** of agents.

Today a remote client (DefraMobile) must pair 1:1 with every individual agent via QR scan. That is a great demo of the DefraDB P2P bootstrap, but it is not a production UX: N agents on a machine means N pairing ceremonies, and every *new* agent means another one. The machine index replaces per-agent pairing with per-machine attachment.

This is the "discovery service layer" agreed in the 2026-07-14 sync. Its feature set will inform what mobile and web clients can eventually do.

## What the service does

1. **Agent index (discovery).** Enumerate the agents (principals + behaviors + deployments) running on this machine, with enough metadata for a client to render a picker: display name, DID, behaviors, dial/health state. This is largely a projection over documents and ops surfaces that already exist (`/status`, `/fleet`, replicated config docs) — the index makes it *dialable and discoverable* from one endpoint per machine instead of one per agent.

2. **Agent creation (the "create layer").** Create new agents on the machine through the index: mint the principal, write the behavior/config documents (role, tools, model), and stand up the deployment. The desktop app can already create local agents; this moves that capability behind a host-level API so remote clients (PWA, possibly mobile) can drive it. Frontend mockups for the agent role/tool/config flow will target this API.

3. **Batch pairing / ACP grants.** Pair a remote control (e.g. the mobile app) with the *machine* once, and have the index grant it ACP to all of the machine's agents — including agents created after the pairing. One QR scan (or login-based attach) → standing access to the whole machine's fleet, instead of a scan per agent.

4. **Identity attach (with the unreleased login system).** Ivan noted we already have a working login/auth system (OpenPubKey-based, OAuth via Google/GitHub) that is unreleased/unused. Ideally the index attaches to an *account*: the account has a DID, and pairing/ACP grants flow from account ↔ machine-index attachment rather than device ↔ agent QR ceremonies. Users should never see blockchain concepts or raw DIDs; cross-device sync falls out of the account linkage.

## How it fits the client architecture (agreed direction)

[...]

PR #873, opened 2026-07-24
feat(p2p): machine index v1 — fleet discovery via machine pairing template (#714)

noahsaso:
## Summary

V1 slice of #714: **fleet discovery for machine-paired clients**, built on the bearer/QR/readiness/iPhone pairing infrastructure now on `main`.

This is not a second pairing system. A bearer pairing already gives the requester a home-wide conversation plane, so the unique piece here is discovery: the `machine` template adds a replicated `AgentDirectoryEntry` projection while reusing the existing signed bearer claim, reciprocal intent, readiness acknowledgement, and client reconnect paths.

## What changed

- Added branchable `AgentDirectoryEntry` rows derived from enabled `AgentPrincipal`, `AgentBehavior`, and `AgentRuntime` documents.
- Made the directory source-owned: every row carries `source_did`, machine replication selects only the home issuer partition, and each projector preserves foreign replicated rows.
- Defined `machine` as the complete conversation projection (including `BearerPairingReady`) plus `AgentDirectoryEntry`:
  - conversation artifacts filter by immutable `requester_did`;
  - readiness filters by `claimant_did`;
  - directory rows filter by issuer-owned `source_did`.
- Replaced unordered “last claim wins” behavior with deterministic authority-signed claim preference by `(issued_at, nonce)`, preventing template oscillation across DefraDB query orders.
- Bound issuer-signed readiness to the actual accepted template (`conversation` or `machine`).
- Extended the desktop/iPhone client path to accept machine bearer invites, install the correct filtered replicator, persist the selected template, restore it after relaunch, revalidate template-specific readiness, and remove legacy/incompatible replicators.
- Added Lean models and proofs for source-partition preservation, deterministic claim preference, and the exact machine-template scope; conformance tests fence the Rust implementation.

The central safety property is that discovery grants nothing: directory rows are read-only derived metadata, while membership, reciprocal intent, and requester-scoped conversation filters remain the authorization/data-plane boundaries.

## Usage

```bash
gents p2p pairings invite --bearer --qr --template machine
```

Existing `conversation` pairings remain unchanged. Re-pair with a machine QR to opt into directory replication.

## Verification

Green locally on the merged current-main state:

- `lake build Proofs.ScopeTemplates Proofs.PeerRegistryDiscovery.DirectoryProjection Proofs.PeerRegistryDiscovery.BearerClaim`
- `cargo test -p gents`
- focused desktop bearer-pairing, CLI template, directory projection, and conformance suites

No `sorry` was introduced. The full `gents` gate passed all 1,187 library tests plus every integration/conformance binary; live credential-dependent cases remained ignored as intended.

## Follow-ups outside this v1 slice

- A picker UI consuming the replicated directory
- Agent creation API extraction
- Account attachment/login integration
- Multi-home/index federation
- A narrower single-agent scope template


noahsaso:
I resolved the merge-from-main conflict and pushed `84c4d994`; GitHub now reports the PR cleanly mergeable. The resolution preserves main's shared GraphQL helpers/test-module extraction and ports the machine-template production changes and regression tests. `cargo fmt --all -- --check` passes. Focused Rust tests could not start because the private `sourcenetwork/defradb.rs` SSH dependency was unavailable in this environment, so CI still needs to run on the current head.

I found one blocking spec/safety issue:

**Critical — source-owned directory partitions are not preserved when two homes contain the same agent DID.** `crates/gents-schemas/schemas/agent/agent_directory_entry.graphql:15` makes `agent_did` globally unique, while `crates/gents/src/agent/directory_projection.rs:458-470` upserts by `agent_did` alone and overwrites `source_did`. A local projector can therefore take over a replicated foreign row with the same agent DID (or encounter a uniqueness conflict), rather than maintaining independent `(source_did, agent_did)` partitions. The conformance fixture also keys entries only by `agent_did` (`crates/gents/tests/conformance/directory_projection.rs:53-59`), and its foreign-row test uses a different DID (`:193-221`), so it cannot represent or fence this collision. This contradicts the PR's explicit partition-preservation property and is weaker than the Lean model.

Please make the persisted identity and all upsert/list/delete/conformance paths genuinely source-partitioned (for example, with a deterministic composite source+agent key supported by the schema), then add an embedded-DefraDB regression where foreign and local rows share the same `agent_did`. The test should prove the local sweep neither overwrites nor deletes the foreign row and converges its own row independently.

I cannot submit a formal “Request changes” review because GitHub identifies this account as the PR author, so this comment records the blocking verdict.

noahsaso:
I re-reviewed `11f4305e` against the updated corrective scope and current merged main.

The blocking source-partition issue is fixed:
- persisted identity is now a unique immutable key derived from `(source_did, agent_did)`;
- projector read/upsert/delete paths use that partitioned identity;
- the conformance fixture can represent identical agent DIDs from different sources;
- both fixture and embedded-DefraDB regressions prove local convergence and retraction preserve the foreign same-DID row;
- the Lean vocabulary now models source-plus-agent identity.

The correction is limited to the four expected schema/projector/conformance/proof files and preserves merge commit `84c4d994` and the machine-template scope. `cargo fmt --all -- --check` and `git diff --check origin/main..HEAD` pass. GitHub reports `MERGEABLE` / `CLEAN` at head `11f4305e`.

I found no remaining blocking spec-compliance or code-quality issues. My verdict is **approve**. GitHub does not permit this account to submit a formal approval review on its own PR, so this comment records the approval. Rust/Lean execution remains for CI because the private DefraDB SSH dependency and Lean toolchain are unavailable here.

PR #881, opened 2026-07-27
fix(cli): rebind DIDs embedded in subagent_targets on config apply

### LINKED ISSUE #876: rebind_manifest_agent_did does not rebind DIDs embedded in subagent_targets JSON strings
## Summary

`rebind_manifest_agent_did()` in `crates/gents-cli/src/commands/config/binding.rs` (L371-379) rebinds the `agent_did` field on the principal, behaviors, and tool selections — but it does **not** rebind the `agent_did` values embedded inside the JSON-encoded `subagent_targets` strings on each `ToolSelection`.

When `config apply --bind-agent-did home --force-rebind-concrete-did` is used (manifest has a placeholder DID like `did:defra-agent:amy`, runtime resolves to `did:key:z6Mki...`), the rebind updates the selection's own `agent_did` to the runtime DID but leaves the `subagent_targets` entries still referencing the placeholder DID.

## Impact

The validator in `validate_subagent_targets()` (`crates/gents-cli/src/desired_state/validate.rs` L1255-1266) compares each target's `agent_did` against the selection's own `agent_did`. After rebind, these mismatch:

```
cross-deployment subagent delegation is deferred; remote target session-classifier requires subagent_allow_cross_deployment=true (trusted-fleet only).
cross-deployment subagent delegation is deferred; remote target glm52 requires subagent_allow_cross_deployment=true (trusted-fleet only).
```

This blocks `gents config apply` for any manifest that uses a placeholder DID and has same-deployment subagent targets. The only workaround is either:
1. Manually editing the `subagent_targets` JSON in the manifest to use the concrete runtime DID before apply, or
2. Setting `subagent_allow_cross_deployment=true` (which defeats the safety guard entirely).

## Reproduction

1. Create a manifest with `agent_did: "did:defra-agent:amy"` (placeholder) and `subagent_targets` referencing the same placeholder DID.
2. Run: `gents config apply --root <manifest-dir> --home <home-dir> --bind-agent-did home --force-rebind-concrete-did`
3. Observe validation errors about "cross-deployment subagent delegation is deferred."

## Proposed Fix

In `rebind_manifest_agent_did()`, after setting each `selection.agent_did = target_did`, iterate through `selection.subagent_targets`, parse each entry as a `SubagentTarget`, replace any `agent_did` matching the old placeholder DID with `target_did`, and re-serialize back to the JSON string form.

```rust
fn rebind_manifest_agent_did(manifest: &mut DesiredStateManifest, target_did: &str) {
    let old_did = manifest.agent_principal.agent_did.clone();
    manifest.agent_principal.agent_did = target_did.to_string();
    for behavior in &mut manifest.agent_behaviors {
        behavior.agent_did = target_did.to_string();
    }
    for selection in &mut manifest.tool_selections {
        selection.agent_did = target_did.to_string();
        // Rebind DIDs embedded in subagent_targets JSON strings
        for entry in &mut selection.subagent_targets {
            if let Ok(mut target) = SubagentTarget::parse(entry) {
                if target.agent_did.trim() == old_did.trim() {
                    target.agent_did = target_did.to_string();
                    *entry = target.to_json_string(); // or equivalent serialization
                }
            }
        }
    }
}
```

The `SubagentTarget` struct already has `parse()` and serialization; the fix just needs to round-trip each entry through parse → patch → serialize when the embedded DID matches the old DID.

## Context

Discovered while enabling the `sessions` tool (PR #398, #417) for the amy agent on studio-1. The manifest at `~/amygdala/infra/agents/studio-1/amy/` uses `did:defra-agent:amy` as a placeholder. `config apply` failed; had to fall back to a direct `update_ToolSelection` GraphQL mutation to flip `enable_session_history_tool` to `true`.

jackzampolin:
## Summary
- Extend `rebind_manifest_agent_did` so same-deployment DIDs inside each `ToolSelection.subagent_targets` JSON entry are rebound with the resolved runtime DID (not only the selection's top-level `agent_did`).
- Preserve genuine cross-deployment targets and malformed JSON so `subagent_allow_cross_deployment` and desired-state validation stay intact.
- Add unit coverage for the rewrite helper and load-path regression tests through `config validate --bind-agent-did home --force-rebind-concrete-did`.

Fixes #876.

## Test plan
- [x] `cargo test -p gents-cli --bin gents -- commands::config::binding::tests`
- [x] `cargo test -p gents-cli --test cli_config_validate -- config_validate_bind_home`
- [x] `cargo check --workspace --all-targets`
- [x] `git diff --check`
- [ ] CI green on this PR

### DIFF
--- a/crates/gents-cli/src/commands/config/binding.rs
+++ b/crates/gents-cli/src/commands/config/binding.rs
@@ -369,12 +369,43 @@ fn insert_nonempty(values: &mut BTreeSet<String>, value: &str) {
 }
 
 fn rebind_manifest_agent_did(manifest: &mut DesiredStateManifest, target_did: &str) {
+    // Capture source/local DIDs before overwriting so same-deployment
+    // `subagent_targets` entries can be recognized after the top-level rewrite.
+    // Only DIDs this function already treats as local (principal, behaviors,
+    // selections) are candidates — genuine cross-deployment target DIDs are
+    // left unchanged.
+    let source_local_dids = manifest_agent_dids(manifest);
+
     manifest.agent_principal.agent_did = target_did.to_string();
     for behavior in &mut manifest.agent_behaviors {
         behavior.agent_did = target_did.to_string();
     }
     for selection in &mut manifest.tool_selections {
         selection.agent_did = target_did.to_string();
+        rebind_subagent_target_dids(
+            &mut selection.subagent_targets,
+            &source_local_dids,
+            target_did,
+        );
+    }
+}
+
+/// Rewrite DIDs embedded in `subagent_targets` JSON entries that match a
+/// source/local DID being rebound. Malformed entries are preserved so desired-
+/// state validation can still report the precise parse error.
+fn rebind_subagent_target_dids(
+    entries: &mut [String],
+    source_local_dids: &BTreeSet<String>,
+    target_did: &str,
+) {
+    for entry in entries {
+        let Ok(mut target) = gents::SubagentTarget::parse(entry) else {
+            continue;
+        };
+        if source_local_dids.contains(target.agent_did.trim()) {
+            target.agent_did = target_did.to_string();
+            *entry = target.to_entry();
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


jackzampolin:
Thanks for the review — addressed the nits and filed the follow-up.

1. **Skills / projection ACP rebind gap** → tracked in #882 (out of scope for this PR as agreed).
2. **Lossy SubagentTarget round-trip** → doc comment on `rebind_subagent_target_dids` notes normalization and unknown-field drop.
3. **Dead `allow_cross_deployment` param** → removed from the test helper; call sites always used `false`.

Pushed in 863d9778.

jackzampolin:
Addressed the request-changes findings in af185324:

1. **Blocking — rename-guard fixtures**: replaced `did:defra-agent:amy` with `did:test:amy` in unit + CLI tests. `scripts/rename-to-gents.sh guard all` now PASS locally.

2. **P2 — unknown field loss**: `rebind_subagent_target_dids` now mutates only `agent_did` on the raw JSON object (no `SubagentTarget` reserialize). Unknown fields / key order survive; already-bound DIDs are left byte-identical. Added `rebind_preserves_unknown_subagent_target_fields`.

Verification: 7 binding unit tests pass; 4 `config_validate_bind_home*` tests pass; rename guard pass; `git diff --check` clean.

PR #936, opened 2026-07-29
test(r5): give Crash a real process restart boundary

jackzampolin:
## Summary

R5 crash scenarios (`r5_a_crash_mid_wait`, `r5_b_crash_mid_execution`) previously went green without exercising any crash/restart behavior. The harness arm was:

```rust
Action::Crash { node: _ } => {}
```

This PR repairs the R5 document harness so `Crash` is an honest process boundary for cross-deployment background **subagent** work (`await_mode = "background"` on an `AgentToolCall` bridge), without inventing new lifecycle semantics.

### False-green evidence (pre-fix)

- `Action::Crash` was a no-op in `runner.rs`.
- Deleting or no-op'ing Crash did not fail `r5_a_crash_mid_wait` / `r5_b_crash_mid_execution`.
- After this PR, restoring the no-op fails with:
  `Crash(A) left a_process_generation at 0 (no-op crash)`
  (demonstrated during development against the new invariants).

### Crash boundary now exercised

`TestDb::simulate_process_crash`:

1. `EmbeddedNode::shutdown()`
2. Exclusive `Arc` drop (fails if outstanding clones would leave process state live)
3. Reopen the **same durable data path**
4. Advance `process_generation`

This matches the defra-node restart test pattern and the TLA lead (`SubagentCompletion` CrashA/CrashB, `SubagentCancelPropagation` Crash): volatile process identity is cleared; durable intent/rows remain.

### Durable vs volatile

| Durable (survives Crash) | Volatile (does not survive) |
|---|---|
| `AgentRequest`, `AgentToolCall`, `AgentResponse`, `AgentMessage`, pairing docs | Process-local workers, registries, subscriptions, live observer tasks |
| Bridge lifecycle fields, cancel intent, child terminals | `process_generation` / node handle identity |

Harness does **not** own a live `AgentRuntime`; observers/recovery are invoked as explicit actions against the reopened store (same as production startup recovery re-drives from durable rows).

### Formal / conformance binding

Harness-only repair; **no Lean change** (runtime contract already correct).

Binds to existing formal surface:

- TLA+ `SubagentCompletion` CrashA/CrashB (clear process-local inbound; durable child/bridge/notification/queue retained)
- TLA+ `SubagentCancelPropagation` Crash (clear `inFlight` / `pendingInbound`; retain durable cancel intent/state)
- Lean `Proofs.Background.Properties.Projection` (child terminal projects onto parent bridge)
- Production recovery: `ToolCallLifecycle::recover_all` leaves running background **subagent** bridges running while a live parent waits; `project_background_subagent_completion` / recovery side effects project terminal children after restart

R6 `r6_background_recovery` covers native background **tools** (interrupt-on-restart). These R5 scenarios remain the cross-deployment background **subagent** slice.

### What changed

| Area | Change |
|---|---|
| `TestDb` | `process_generation`, `simulate_process_crash` |
| `Action::Crash` | Real reopen boundary |
| Observation | `a/b_process_generation`, `crashed_node` |
| Parent tool writes | Stamp immutable `agent_did` so recovery can scope bridges |
| Invariants | `assert_crash_boundary` (generation advances; durable rows survive) |
| Crash scenario tests | End-state projection / notification / generation assertions |

### Tests and commands

```text
cargo test -p gents --test conformance 'r5_scenarios::' -- --nocapture
# 5 passed

# Falsifiability: no-op Crash → panic
# Crash(A) left a_process_generation at 0 (no-op crash)

cargo test -p gents --test e2e_subagent r6_background_recovery
# 1 passed

cargo test -p gents
# all targets green (0 failed)

cargo check --workspace --all-targets
# ok

cargo fmt --all -- --check
# ok

git diff --check
# ok
```

No Lean/TLA artifacts changed; `lake build` / TLC not re-run.

### Known limitations / non-goals

- Does not spin a full multi-process agent daemon or P2P mesh restart.
- Does not claim coverage of R6 native background tool registries.
- Does not restore hollow #900 invariant stubs or fake convergence timeouts.
- Replication remains an explicit `ReplicateDoc` harness action (not live gossip during crash).

### Review notes

- Semantic parity for non-crash scenarios: happy path, multi-completion, partition-during-cancel only gain observation fields defaulting to generation 0 / no crash marker.
- Independent review still welcome on exclusive-drop safety of `simulate_process_crash` and whether stamping `agent_did` on fixture bridges is the right recovery coupling.

## Test plan

- [x] `cargo test -p gents --test conformance 'r5_scenarios::'`
- [x] Falsify no-op Crash against new invariants
- [x] `cargo test -p gents`
- [x] `cargo check --workspace --all-targets`
- [x] `cargo fmt --all -- --check` and `git diff --check`
- [ ] CI on draft PR

### DIFF
--- a/crates/gents/tests/conformance/r5_scenarios.rs
+++ b/crates/gents/tests/conformance/r5_scenarios.rs
@@ -22,6 +22,15 @@ async fn run_scenario(filename: &str) -> Vec<Observation> {
     history
 }
 
+async fn run_crash_scenario(filename: &str) -> Vec<Observation> {
+    let history = run_scenario(filename).await;
+    // Crash fixtures must cross a real process boundary; these checks fail if
+    // Crash is deleted or restored as a no-op while the fixture still claims
+    // a crash (false-green regression from the pre-fix harness).
+    invariants::assert_crash_boundary(&history);
+    history
+}
+
 #[tokio::test]
 async fn r5_happy_path() {
     run_scenario("happy_path.json").await;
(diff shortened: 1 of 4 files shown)


jackzampolin:
## Review follow-up (`9508843b`)

Addressed the hold-merge findings on `simulate_process_crash`:

### Main concern: unsafe soundness
- **Removed all `unsafe`** (`ptr::read` / `ptr::write` are gone).
- Exclusive drop is now `mem::replace` with a short-lived ephemeral stand-in so `self.node` is **always initialized** under cancel, panic, or reopen failure — no double-free path.
- Kept `pub node: Arc<EmbeddedNode>` rather than `Option` + accessor: that field is the harness surface used by ~1700 call sites; the stand-in replace is the same soundness property without the rename churn. Documented in the method.
- **Deleted the pointer-equality check** (allocator reuse flake; `process_generation` is the identity signal).
- **Single-shot reopen** (no identical retry-then-bail).
- On `strong_count > 1` after replace: restore the shut-down handle and error that it is **unusable**.
- Pre-check `strong_count == 1` *before* shutdown so a shared-handle bail still leaves a live node.

### Smaller points
- `import_tool_call` missing `agent_did` falls back to **destination node** DID (A/B), not always `NODE_A_DID`.
- Doc: `crashed_node` says **`None` when…**
- Collapsed A/B arms in `assert_durable_rows_survive_crash`.
- Dropped `checked_add` ceremony.

### Validation
```
cargo test -p gents --test conformance r5_scenarios::
# 5 passed
```

Issue #938, opened 2026-07-29
Bearer invite QR exceeds practical camera-decode density (~1.1-1.3k chars, QR v24-25)


noahsaso:
v0.9.0 bearer tokens measure 1,136 chars (machine template) to 1,263 chars (conversation) — the optional signed routing hint added to `BearerInviteToken` pushed them past the practical threshold for terminal-rendered QR scanning. At EC level L (already the minimum, `invite.rs:154`) this is QR version ~24–25 (117×117 modules); iOS VisionKit detects the finder patterns but consistently fails to decode (observed live 2026-07-29: camera "sees" the QR, nothing fires; pasting the same token works).

Options, roughly in order of leverage:
1. `--qr-png <path>` (or auto temp-file + `open`) — a crisp raster at controlled module size scans far better than Dense1x2 unicode in a terminal.
2. Shorter wire encoding: bs58 over CBOR costs ~37% expansion; base64url or CBOR-in-base45 with QR alphanumeric mode would cut module count substantially.
3. Trim token fields (the routing hint is the marginal addition; nonce/ticket dominate).

Workaround: paste the token into the client manually (works today).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Arv3YzUJH1oSvYL21kekca

jackzampolin:
Obsolete: the bearer invite/QR/claim model was removed; authenticated enrollment is the only route (#1313, tests/misc/legacy_authority_absence.rs).

PR #939, opened 2026-07-29
fix(p2p): machine pairing fails to install its replicator — source_did must be @immutable

noahsaso:
## Summary

Every `machine`-template pairing on current main fails to install its data-plane replicator:

```
add_replicator: replication filter field 'source_did' in collection '<AgentDirectoryEntry cid>' must be marked @immutable
```

defradb validates that replication-filter fields are `@immutable` at install time (the same rule migration.rs already documents for `requester_did`). The #873 merge amendments added the `source_did` partition filter to the machine template but left the field mutable in the SDL. Net effect observed live: bearer claim succeeds, phone pushes fine, but the server's reverse push never installs — chat clients hang on "thinking…" while the server visibly processes and streams.

## Fix

- `agent_directory_entry.graphql`: `source_did` gains `@immutable` (semantically correct regardless — it is half of the `directory_key` identity hash, which is already immutable), with a comment pointing at the defradb rule.
- **New catalog-wide fence** (`embedded_all_builtin_template_filters_pass_replicator_validation`): every builtin template's filter set is installed through real `add_replicator` against embedded nodes with the full runtime schemas. This is the validation defradb only performs at install time, now enforced at test time — a future template/filter/schema mismatch cannot ship again. RED reproduced the exact live error before the fix.
- Drive-by: `gents-schemas` `ALL.len()` count literal was stale (32 → 33 after BearerPairingReady landed) — `cargo test -p gents-schemas` fails on clean main; suggests CI does not run that crate's suite.

## Upgrade note

Homes bootstrapped between #873's merge and this fix carry the mutable-field collection and will keep failing after upgrading the binary; a fresh home (or manual collection recreation) is needed. Pre-#873 homes are unaffected (collection is created from the fixed SDL).

## Related

- #938 (bearer QR density — the other live-demo finding from the same session)

### DIFF
--- a/crates/gents-schemas/schemas/agent/agent_directory_entry.graphql
+++ b/crates/gents-schemas/schemas/agent/agent_directory_entry.graphql
@@ -18,7 +18,11 @@ type AgentDirectoryEntry @branchable {
     agent_did: String @index
     # DID of the runtime that derived this row. Machine replication filters
     # on this field, and each projector reconciles only its own partition.
-    source_did: String @index
+    # @immutable is REQUIRED: defradb rejects any replication-filter field
+    # that is not immutable at add_replicator time (see migration.rs's
+    # requester_did precedent) - and it is semantically immutable anyway,
+    # being half of the directory_key identity hash.
+    source_did: String @index @immutable
     display_name: String
     # Behavior display names for the picker. Emitted as null (never []) when
     # the principal has no behaviors - empty list literals corrupt nillable
--- a/crates/gents-schemas/src/lib.rs
+++ b/crates/gents-schemas/src/lib.rs
@@ -182,8 +182,17 @@ mod tests {
     use super::*;
 
     #[test]
-    fn all_contains_every_agent_schema() {
-        assert_eq!(ALL.len(), 33);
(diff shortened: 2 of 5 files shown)


noahsaso (commented):
Review outcome: the machine-pairing fix and regression fences are ready from a code-review standpoint (GitHub prevents this account from formally approving its own PR).

- `source_did` is now immutable in the schema and remains present only in the add payload, avoiding illegal immutable-field updates.
- The embedded catalog test exercises real `add_replicator` validation, verifies complete per-collection filter coverage, and waits for peer readiness.
- The schema catalog test no longer relies on a stale numeric count.
- The unavoidable upgrade boundary is documented clearly.

I verified the branch is based on current `main`, is mergeable, and has no whitespace errors in the PR diff. Lean proofs, desktop UI, and rename guard pass. The remaining `rust-and-cli` failure is at `cargo fmt --all --check` exclusively in files unchanged by this PR (`crates/gents/src/schema.rs` and `crates/gents-migration/**`), demonstrating baseline rustfmt drift rather than a defect in this change. The repository baseline formatting still needs correction or its toolchain pinned before the required check can become green.

PR #944, opened 2026-07-29
style: format merged migration code

jackzampolin:
## Summary

Restore the repository-wide rustfmt gate after migration PR #931 landed unformatted source. This is the exact output of `cargo fmt --all` across the five files reported by the failing gate.

## Evidence

- Current main parent `c873acbe` fails `cargo fmt --all -- --check` in these five files.
- Formatting an archived copy of the original #931 merge parent and comparing the entire tracked tree against this commit produced a byte-for-byte match.
- Token-aware comparison found zero added or removed tokens; identifiers, operators, literals, and comment wording are unchanged.
- Stable patch ID remained `2a2b0061429da07785d2a6e8f4e28bd69001fcb5` through the #942 and #943 rebases.
- No tests, assertions, runtime logic, APIs, errors, logging, or generated outputs are changed.

## Validation

- `git diff --check HEAD^ HEAD`
- `cargo check --workspace --all-targets` on exact head/base `c873acbe`
- Independent semantic-parity review: APPROVE, no findings

## Diff

Five files, 53 insertions / 31 deletions, formatting only.

### DIFF
--- a/crates/gents-migration/src/engine.rs
+++ b/crates/gents-migration/src/engine.rs
@@ -187,7 +187,10 @@ async fn apply_add_collection(
 
     // If a pin is present, locate and verify the active version by scanning.
     if let Some(pin) = expected_version {
-        let versions = node.get_all_collection_versions().await.map_err(Error::Node)?;
+        let versions = node
+            .get_all_collection_versions()
+            .await
+            .map_err(Error::Node)?;
         let Some(active) = versions
             .iter()
             .find(|v| v.version_id == pin && !v.is_placeholder)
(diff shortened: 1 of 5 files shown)


jackzampolin:
Superseded by #946. The formatting-only patch and CI-wiring patch form a bootstrap cycle on current main: each PR alone deterministically fails the other inherited gate. #946 retains this exact formatting patch as a separate commit with the same stable patch ID, plus the independently reviewed CI repair, so the combined repair can satisfy the merge-green rule.

PR #946, opened 2026-07-29
ci: restore post-migration gates

jackzampolin:
## Summary

This two-commit bootstrap repairs the two mutually blocking CI regressions introduced by #931:

1. retarget the deleted migration-lens WASM package and explicitly run both migration package suites
2. apply only rustfmt output to the merged migration code

These were prepared and reviewed as independent slices, but neither can produce a green PR alone: the CI-wiring slice inherits the formatting failure from main, while the formatting slice reaches the deleted-package command. Keeping the two reviewed commits together is the narrow path that preserves the merge-green rule.

## CI wiring evidence

The previous WASM command deterministically exits 101 because agent-tool-call-lifecycle-v1-to-v2-lens no longer exists. Cargo metadata identifies gents-lens-fixture-add-label as the canonical current lens crate consumed by the migration build.

The new unfiltered package commands close a #898-class false-green gap:

- gents-migration: 13 non-doc tests; one ignored doctest
- gents-lens-fixture-add-label: 3 tests

A missing package, vanished suite, compile failure, or test failure now fails CI.

## Formatting parity evidence

The second commit has stable patch ID 2a2b0061429da07785d2a6e8f4e28bd69001fcb5, identical to the independently reviewed formatting PR. It is exactly cargo fmt output across five Rust files. Whole-tree byte comparison against an archived rustfmt result matched, and token-aware comparison found zero added or deleted Rust tokens.

The two commits touch disjoint file sets and retain their independently reviewed patch IDs.

## Validation

- cargo fmt --all -- --check
- cargo check --workspace --all-targets
- cargo build -p gents-lens-fixture-add-label --target wasm32-unknown-unknown
- cargo test -p gents-migration
- cargo test -p gents-lens-fixture-add-label
- bash -n scripts/rename-to-gents.sh
- scripts/rename-to-gents.sh self-test
- scripts/rename-to-gents.sh guard all
- cargo metadata --no-deps resolves both package IDs
- git diff --check

Independent reviews of both constituent patches: APPROVE with no findings. Final composition review: APPROVE with no findings; patch identities, disjoint file sets, gate failures, and absence of order-dependent interaction were independently verified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,7 +167,12 @@ jobs:
         run: cargo check --workspace --all-targets
 
       - name: Build the WASM lens target
-        run: cargo build -p agent-tool-call-lifecycle-v1-to-v2-lens --target wasm32-unknown-unknown
+        run: cargo build -p gents-lens-fixture-add-label --target wasm32-unknown-unknown
+
+      - name: Run migration package tests
+        run: |
+          cargo test -p gents-migration
+          cargo test -p gents-lens-fixture-add-label
 
       # Guards the noq 1.0.1 active_connections underflow fix (gents#634 /
       # n0-computer/noq#743). Vendored under third_party/; not a workspace member.
--- a/crates/gents-migration/src/engine.rs
+++ b/crates/gents-migration/src/engine.rs
@@ -187,7 +187,10 @@ async fn apply_add_collection(
 
     // If a pin is present, locate and verify the active version by scanning.
     if let Some(pin) = expected_version {
-        let versions = node.get_all_collection_versions().await.map_err(Error::Node)?;
+        let versions = node
+            .get_all_collection_versions()
+            .await
+            .map_err(Error::Node)?;
         let Some(active) = versions
             .iter()
             .find(|v| v.version_id == pin && !v.is_placeholder)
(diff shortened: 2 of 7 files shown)


jackzampolin:
Exact combined-head validation update: cargo check --workspace --all-targets passed at c274b8601a116811cda493b4e7c01a47a759d454 in 13m31s after a cold full dependency/workspace/desktop build. cargo fmt --all -- --check and git diff --check remain clean.

jackzampolin:
Hosted CI update: desktop-ui passed. lean-proofs compiled and ran 319 conformance tests; 314 passed and five runtime-boot cases timed out together at the pre-existing shared 10-second readiness helper. The same cases passed in PR #945 on the current main merge, so this is being reproduced and fixed as a separate test-integrity slice rather than folded into this CI repair. I canceled the still-queued rust-and-cli job because current main has a separately reproduced AgentDirectoryEntry migration-baseline omission that makes the full gents suite fail outside this diff; the broader baseline correction already exists in PR #930. This draft remains intentionally blocked on those two independent prerequisites.

jackzampolin:
Superseded by #947, which merges this branch (CI retarget + rustfmt) together with the materialize hard-fail follow-up. Closing to avoid double-landing engine.rs conflicts.

PR #949, opened 2026-07-30
fix(migration): advance DefraDB and restore green baseline

jackzampolin:
## Summary

Restores a single green path from current `main` after the migration and directory-schema changes:

- Advances all nine DefraDB workspace dependencies from `e201d77c` to the immutable stabilization revision `966cf42bd337ada1c6f64584bea9e52f82001393` (including the upstream early-idle-exit P2P sync fix).
- Registers `AgentDirectoryEntry` in the canonical production baseline before desktop P2P subscription.
- Pins all 39 production collection roots and rejects a foreign single-root lineage as `UnknownLineage`, not only foreign multi-version DAGs.
- Regenerates the two roots that differ on current `main` (`AgentRequest` and `AgentToolCall`) instead of copying #930's branch-specific CIDs.
- Makes the chain-replay authoring test print and compare every computed root even when frozen pins are stale.
- Keeps the historical v0.6.12 store fail-closed and adapts the duplicate-tolerant recovery fixture to use an explicit test-only catalog.
- Asserts P2P subscription truth from the transport rather than treating the per-peer applied ownership ledger as global subscription state.
- Extends the two shared runtime-ready test waits from 10s to 30s, removing the self-hosted scheduling flake that made the migration and readiness PRs mutually block each other.

No release files or release workflow are changed.

## Why the Defra pin and migration roots move together

Defra collection version IDs are content-addressed. The all-root table generated on #930 could not be copied blindly onto current `main`: the built-in authoring test found that #930's request/tool-call schemas produce different CIDs. Conversely, keeping the old Defra pin while using the new root table causes a fresh install to reject its own `AgentRequest` root.

This PR therefore moves the dependency revision, exact current-main schema catalog, and matching root pins atomically.

## Main CI failures addressed

1. Current `main` timed out after 10 seconds waiting for runtime readiness in Lean-backed conformance.
2. The dedicated readiness PR then reached desktop smoke and failed 10/125 `gents-desktop-core` tests because `AgentDirectoryEntry` was subscribed before it had been registered.
3. A registration-only fix exposed the existing unpinned single-root lineage gap in the historical migration fixture.

The combined commits close all three without weakening migration admission or adding startup cleanup writes.

## Formal boundary

`Proofs/Migration.lean` explicitly assigns full foreign-DAG pin soundness to Rust conformance. This change strengthens that existing fail-closed boundary and does not alter the modeled activation or materialization transitions, so no Lean model change is required. The full conformance suite still runs and contains zero `sorry`s.

## Validation

Passed on exact head `50e7e6a6`:

- `cargo test -p gents-migration` — all migration tests passed (one doctest intentionally ignored).
- `cargo test -p gents-desktop-core` — 125 unit + 18 integration tests passed; all ten hosted startup failures are covered.
- `cargo test -p gents` — 1150 unit, 319 conformance, 39 lifecycle, 47 runtime, 87 subagent, 8 trigger, and all remaining package tests passed (live/external tests intentionally ignored).
- `cargo check --workspace --all-targets`.
- `cargo fmt --all -- --check`.
- `git diff --check`.

The root authoring test independently registered the ordered canonical SDL catalog and confirmed all 39 computed VersionIDs match the frozen table.

## PR ownership

This consolidates the two-line readiness fix from #948 so hosted CI has one non-cyclic recovery path. It extracts the Defra stabilization pin from #930, but does not include or supersede #930's desktop/client correctness work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ dependencies = [
 [[package]]
 name = "acp"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4#e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=966cf42bd337ada1c6f64584bea9e52f82001393#966cf42bd337ada1c6f64584bea9e52f82001393"
 dependencies = [
  "async-lock",
  "async-trait",
@@ -2464,7 +2464,7 @@ dependencies = [
 [[package]]
 name = "blockstore"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4#e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=966cf42bd337ada1c6f64584bea9e52f82001393#966cf42bd337ada1c6f64584bea9e52f82001393"
 dependencies = [
  "async-trait",
  "bytes",
@@ -5843,7 +5843,7 @@ dependencies = [
 [[package]]
 name = "crdt"
 version = "0.5.0"
-source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4#e201d77c5b20eb76ae6f20de1db3e61cbc2e1cb4"
+source = "git+ssh://git@github.com/sourcenetwork/defradb.rs.git?rev=966cf42bd337ada1c6f64584bea9e52f82001393#966cf42bd337ada1c6f64584bea9e52f82001393"
 dependencies = [
  "async-trait",
  "defra-core",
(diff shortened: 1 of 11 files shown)


jackzampolin:
CI classification update:

- `lean-proofs` compiled and checked the Lean model successfully.
- The job then ran the Rust conformance suite: 318/319 passed.
- The sole failure was `cancel_propagation_cases_drive_production_interrupt`, which timed out in `tests/support/interrupt.rs` with `agent did not reach ready state`.
- That is the independent readiness flake isolated in #948; this PR does not touch the runtime-ready helper or cancellation behavior.

I am leaving this draft unchanged and will rebase/rerun after #948 lands. The migration catalog/pinning validation remains the substantive gate for this slice.

jackzampolin:
Independent review completed on exact head `50e7e6a601bb46f8b45aaddd43f4207cae4a31a7`: **APPROVE, no code blockers**. Grok and Claude Opus independently reached the same conclusion.

Static and authoring evidence reviewed:

- all six commits and 11 changed files were inspected against exact `main` `5d8d8628`;
- the production baseline is protocol-ordered (1 runtime + 38 canonical collections), with `AgentDirectoryEntry` immediately after `AgentRuntime`;
- all 39 frozen roots are unique 59-character CIDs;
- the existing compiled root-authoring test, built against Defra `966cf42`, recomputed and matched all 39 roots;
- all nine direct Defra dependencies and lockfile transitive sources advance together;
- a missing single-root pin remains fail-closed and is now classified as `UnknownLineage`;
- duplicate-tolerant recovery remains isolated to an explicit test-only catalog;
- the historical v0.6.12 fixture remains fail-closed;
- P2P readiness now inspects actual transport subscriptions while retaining the peer-address ownership evidence;
- the readiness commit has the same stable patch ID as #948, which is fully superseded.

Exact-head hosted CI remains the merge gate; rename-guard is green while desktop and Lean jobs are still queued.

Important #930 integration requirement: after this PR lands, rebase #930 and drop its overlapping Defra pin/catalog/baseline-root work. Preserve these cutover roots as immutable. The new `retry_key` and completion-notification schema fields must be expressed as post-baseline `PatchVersioned` migration steps (with destination pins), not by replacing the frozen `AgentRequest`/`AgentToolCall` baseline roots. Preserve #930's later P2P harness serialization/shutdown work during that rebase.

Issue #798, opened 2026-07-19
19-node fresh-store control mesh saturates pending DAGs and loses AgentNetwork applied scope


jackzampolin:
## Summary

The ignored 19-process release acceptance test consistently failed during the fresh-store control-plane phase when exercised for the v0.7.0 release against DefraDB v0.16.0 (`4d7a2f42f2e7a675acdbae0d82a339e64b63dbc1`). The failure occurs before model inference or the delegation sweep.

We are accepting this as a tracked release exception for v0.7.0 because the normal Rust, CLI, desktop compile, and hermetic two-process E2E gates are green. The ignored fleet test remains runnable and must stay linked to this issue.

## Reproduction

```bash
DEFRA_AGENT_RELEASE_ACCEPTANCE=1 \
DEFRA_AGENT_LIVE_OPENAI_ENDPOINT=http://<openai-compatible-host>/v1 \
DEFRA_AGENT_LIVE_OPENAI_MODEL=<model> \
cargo test -p defra-agent-cli --test cli_fleet_delegation_live \
  nineteen_process_release_acceptance_live \
  -- --ignored --nocapture --test-threads=1
```

Observed runtime: about 133 seconds. The run failed in Phase G while converging the 18-spoke genesis/control-plane mesh.

## Expected

- Every coordinator-to-spoke and spoke-to-coordinator control edge materializes the `AgentNetwork` subscription scope.
- Pending DAGs and other P2P queues remain bounded and reach quiescence.
- No node emits capacity-shedding diagnostics.

## Actual

One reverse applied-pairing row acquired a replicator address but never materialized its collection scope:

```text
coordinator -> did:key:z6Mkqv2VEQs9Hn1Hxm9uZxKDbBMW4AHMrcB7uwMPN4LwhgL9
replicator containing AgentNetwork

timed out waiting for replicator install containing AgentNetwork on edge
peer=259da6d98015179ea46fbf55bb5093aee18b8936e989d766a1f3d8dcf831e1d8
last PeerPairingApplied rows:
[{"peer_id":"259da6d98015179ea46fbf55bb5093aee18b8936e989d766a1f3d8dcf831e1d8",
  "collections":null,
  "replicator_addresses":["endpointaasz3jwzqakrphven67vlo2qsoxodc4jg3uytv3guhz5rxhyghq5qaibab7qaaab3g7ag"]}]
```

At the same time, all 19 daemons repeatedly emitted:

```text
[...]

PR #950, opened 2026-07-30
test(cli): require fleet live prerequisites

jackzampolin:
## Summary

Makes the six ignored fleet qualifications fail explicitly when their documented opt-in gate is missing, instead of logging a skip and returning `Ok(())`.

- Five tests require `GENTS_LIVE_OPENAI=1`.
- The 19-process release qualification requires `GENTS_RELEASE_ACCEPTANCE=1`.
- All six remain `#[ignore]`, so ordinary package runs are unchanged.

The repeated gate is consolidated into one private test helper. Endpoint/model resolution, default fallbacks, process setup, network behavior, assertions, diagnostics, timeouts, and cleanup remain unchanged.

## False-green evidence

Previously, explicitly selecting the ignored target with `--ignored` but omitting the advertised environment gate produced a green test without exercising any fleet behavior.

After this change, running all ignored tests with both gates unset produces six failures in 0.04 seconds:

- five name `GENTS_LIVE_OPENAI`;
- one names `GENTS_RELEASE_ACCEPTANCE`;
- each says `set <GATE>=1 and pass --ignored to run this fleet live qualification`.

The intentional `GENTS_FLEET_SUBSTRATE_ONLY=1` early success remains untouched: that path first brings up the fleet and verifies substrate convergence.

## Coverage preservation

No test is deleted or disabled. Every live body and assertion remains after the prerequisite fence. The six live qualifications were not executed because no live OpenAI-compatible endpoint/model credentials were supplied; the 19-process test also remains explicitly known-failing under #798.

## Validation

- `cargo test -p gents-cli --test cli_fleet_delegation_live --no-run` — pass
- normal focused target run — 1 passed, 6 ignored
- gates-unset ignored run — expected six prerequisite failures
- `git diff --check origin/main...HEAD` — pass

Stable patch ID was preserved across the current-main rebase: `6bd3162493bf87a53ad2bfa20970de9019df0ce9`.

An independent Codex review verified all six bodies and the substrate-only exception.

### DIFF
--- a/crates/gents-cli/tests/cli_fleet_delegation_live.rs
+++ b/crates/gents-cli/tests/cli_fleet_delegation_live.rs
@@ -76,6 +76,14 @@ use uuid::Uuid;
 
 type FleetShimWebSocket = WebSocketStream<MaybeTlsStream<TcpStream>>;
 
+fn require_live_gate(gate: &str) -> Result<()> {
+    anyhow::ensure!(
+        std::env::var(gate).as_deref() == Ok("1"),
+        "set {gate}=1 and pass --ignored to run this fleet live qualification"
+    );
+    Ok(())
+}
+
 const P2P_LOOPBACK_ARGS: &[&str] = &[
     "--p2p-bind-addr",
     "127.0.0.1",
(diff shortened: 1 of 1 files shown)


jackzampolin:
Hosted `rust-and-cli` failure is the known current-`main` migration catalog baseline, not this live-gate patch.

The job passed formatting, workspace all-target compilation, WASM, migration-package, noq, and the earlier PR smoke targets. `gents-desktop-core` then completed with 115 passed / 10 failed; all 10 failures report:

> collection 'AgentDirectoryEntry' not found - add schema before subscribing to P2P

The changed file is only `crates/gents-cli/tests/cli_fleet_delegation_live.rs`, so none of the failing desktop-core paths overlap this diff. #949 isolates the missing catalog registration, while #930 currently carries the coherent registration plus all-root lineage pins; this PR should be rerun/rebased after that migration foundation lands rather than copying the fix here.

Failing job: https://github.com/source-inc/gents/actions/runs/30513467619/job/90784609524

jackzampolin:

PR #951, opened 2026-07-30
refactor(cli): remove dead schema access flag

jackzampolin:
## Summary

Removes the inert `ensure_local_schemas` boolean from the private CLI `resolve_config_access` helper and its 42 call sites, then updates the migration design wording to describe the behavior that already exists.

This is deletion-only API cleanup inside the CLI crate: 83 lines removed, 50 retained/reflowed across 28 files.

## Dead-code evidence

Before this change, the helper discarded the argument with `let _ = ensure_local_schemas` and then unconditionally called `ensure_all_runtime_migrations` whenever it opened a local embedded node. Explicit GraphQL access returned before local node creation and never observed the flag.

Therefore `true` and `false` were behaviorally identical. No `ensure_local_schemas` occurrence remains after the patch.

## Parity evidence

- Local access still creates the data directory, opens the same RocksDB node, and unconditionally runs the same migration entry point.
- Explicit GraphQL and live-runtime endpoint resolution retain the same early-return behavior.
- Every caller keeps the same home and GraphQL inputs, return values, error propagation, and command flow.
- Public CLI flags, command names, output, error strings, logging, feature gates, and module paths are unchanged.

The two commits were rebased onto current `main` with exact `git range-diff` equality. Stable patch IDs remain `85751c...` for code and `e4b33e...` for docs.

## Validation

- `cargo check -p gents-cli --all-targets` — pass
- `cargo test -p gents-cli --no-run` — pass, including all 42 integration binaries
- focused local config apply — pass
- focused GraphQL config apply — pass
- focused schema apply/idempotence — pass
- focused config read/list/show — pass
- `git diff --check origin/main..HEAD` — pass

An independent Codex semantic review found no observable branch controlled by the removed argument.

### DIFF
--- a/crates/gents-cli/src/commands/background.rs
+++ b/crates/gents-cli/src/commands/background.rs
@@ -25,7 +25,7 @@ async fn background_list(args: BackgroundListArgs) -> Result<()> {
     let now = Utc::now();
     let age_cutoff = age_cutoff(args.age_gt.as_deref(), now)?;
     let (access, _home_dir) =
-        resolve_config_access(args.home.as_deref(), args.graphql.as_deref(), false).await?;
+        resolve_config_access(args.home.as_deref(), args.graphql.as_deref()).await?;
     let tool_calls = load_background_tool_calls(&access, &args, age_cutoff).await?;
     let liveness = match &access {
         ConfigAccess::Graphql(graphql) => {
--- a/crates/gents-cli/src/commands/codex_auth_probe.rs
+++ b/crates/gents-cli/src/commands/codex_auth_probe.rs
@@ -27,7 +27,7 @@ struct OpenAiModelSummary {
 
 pub(crate) async fn codex_auth_probe(args: CodexAuthProbeArgs) -> Result<()> {
     let (access, home_dir) =
-        resolve_config_access(args.home.as_deref(), args.graphql.as_deref(), true).await?;
+        resolve_config_access(args.home.as_deref(), args.graphql.as_deref()).await?;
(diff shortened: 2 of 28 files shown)


jackzampolin:
External semantic-parity review completed after publication:

- **Claude Opus: APPROVE.** Confirmed the boolean had only a `let _ = ensure_local_schemas` discard, all 42 call sites remain present with the two-argument signature, local migration execution and GraphQL early returns are byte-equivalent, and no public CLI/API/feature behavior changed.
- **Grok: APPROVE.** Independently confirmed zero stale references or three-argument calls, unchanged local-vs-GraphQL selection and error/logging behavior, and accurate updated migration wording.

No review findings required code changes.

jackzampolin:
Hosted `lean-proofs` failure is the reproduced current-`main` runtime-readiness flake, not this CLI-only schema-flag removal.

The conformance target completed 317/319. Both failures stopped at the same shared helper and message:

- `generated_streaming_response_interrupt_flow_cases_drive_daemon_contract`
- `generated_r5_cross_deployment_cases_drive_production_dispatch`
- `crates/gents/tests/support/interrupt.rs:69`
- `agent did not reach ready state`

This PR changes only CLI argument/config plumbing and does not touch either runtime or conformance path. #949 now contains the reviewed readiness patch that extends this exact 10-second helper to 30 seconds as part of the coherent migration/Defra/catalog foundation; its readiness commit has the same stable patch ID as superseded #948.

No branch-local code fix is justified. Rebase and rerun after #949 lands.

Failing job: https://github.com/source-inc/gents/actions/runs/30513484468/job/90778242588

jackzampolin:

PR #959, opened 2026-07-30
chore(gents): simplify redundant content borrows

jackzampolin:
## Summary

Apply a narrow clippy cleanup to redundant content access in Gents tests and one equivalent macOS keychain coercion:

- 20 `first_content(&content)` calls become `first_content(content)`
- 4 `content.iter().nth(n)` calls become `content.get(n)`
- 1 macOS keychain `&bytes` argument becomes `bytes`

## Semantic-parity evidence

At every `first_content` site, match ergonomics bind `content` as `&Vec<T>` and the helper accepts `&[T]`; the removed `&` only created an immediately dereferenced extra reference. Each `get(n)` returns the same `Option<&T>` at the same zero-based index as `iter().nth(n)`, with ordering and surrounding assertions unchanged.

`private_key.raw()` returns `&[u8]` in the pinned DefraDB crypto crate and `SecKeychain::set_generic_password` accepts `&[u8]`; the old `&&[u8]` was deref-coerced. No key bytes, keychain service/account, context, or error string changes.

Independent review found exactly these 25 one-line substitutions and no attribute, function, assertion, literal, order, error, or control-flow changes.

## Validation

- `cargo check -p gents --lib` (macOS): pass
- `cargo test -p gents --lib`: 1,146 passed; 4 failed solely because current `main` omits `AgentDirectoryEntry` from the baseline catalog (#949); all affected test modules passed
- focused clippy with `-D clippy::needless_borrow -D clippy::iter_nth`: no `iter_nth` finding; only remaining `needless_borrow` is `tests/conformance/r5_scenarios.rs:19`, intentionally left to the separate user-owned R5 workstream
- `git diff --check`
- independent semantic review: APPROVE

## Scope

No API, persisted data, runtime behavior, logging, errors, retries, visibility, feature gates, or module paths change. This is a reviewable clippy family slice, not a repository-wide warning sweep.

### DIFF
--- a/crates/gents/src/agent/loop_stream/tests.rs
+++ b/crates/gents/src/agent/loop_stream/tests.rs
@@ -1061,7 +1061,7 @@ async fn context_message_is_sent_before_prompt() {
     assert!(matches!(
         &histories[0][0],
         Message::User { content }
-            if matches!(first_content(&content), UserContent::Text(text) if text.text.starts_with("<context>"))
+            if matches!(first_content(content), UserContent::Text(text) if text.text.starts_with("<context>"))
     ));
 }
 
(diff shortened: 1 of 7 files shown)


jackzampolin:

PR #964, opened 2026-07-30
test(gents): scope mutex guards before awaits

jackzampolin:
## Summary

Replace two explicit mutex-guard `drop` calls in tests with lexical scopes:

- the P2P reconcile filter-recording seam test
- the directory projection same-agent convergence test

This removes the two targeted `await_holding_lock` diagnostics without changing production code or test coverage.

## Semantic-parity evidence

An exact word-level comparison shows that only the scope boundaries replace the explicit `drop(guard)` calls. Every assertion, map lookup, literal, comment, test name/attribute, statement, and order remains unchanged.

In both tests, the closing brace occupies the old explicit-drop position immediately before the same async operation:

- `add_replicator(...).await`
- the second `reconcile_directory_tick(...).await`

The mutex guards are therefore released at the same point and still never cross an await.

## Validation

- exact P2P seam test: 1 passed, 1,149 filtered
- exact directory projection test: 1 passed, 318 filtered
- `cargo clippy -p gents --tests -- -D clippy::await_holding_lock`: pass; zero denied findings
- `git diff --check`: pass

The targeted Clippy run still reports unrelated pre-existing warnings (87 in the lib-test target and 8 in conformance), which remain outside this reviewable family.

## Scope

Test-only mechanical lifetime cleanup. No assertion, coverage, public API, runtime behavior, formal invariant, error string, logging, retry behavior, visibility, feature gate, or module path changes.

### DIFF
--- a/crates/gents/src/agent/p2p_reconcile/engine/tests.rs
+++ b/crates/gents/src/agent/p2p_reconcile/engine/tests.rs
@@ -1988,13 +1988,14 @@ async fn add_replicator_records_filters_at_seam() {
         .await
         .expect("add_replicator empty filters");
 
-    let calls = admin.recorded_filters.lock().unwrap();
-    assert_eq!(calls.len(), 1);
-    assert!(
-        calls[0].1.is_empty(),
-        "empty filters should record as empty"
-    );
-    drop(calls);
+    {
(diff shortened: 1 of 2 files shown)


jackzampolin:

PR #968, opened 2026-07-30
refactor(desktop): simplify optional agent scoping

jackzampolin:
## Summary

Simplify seven optional-agent scope predicates across the desktop bridge snapshot builder and desktop core store.

This is one cohesive Clippy family:

- five snapshot/session selection predicates
- two core store source-agent match predicates

No naming, ownership-model, filtering, or adjacent structural change is included.

## Semantic-parity evidence

Every substitution is:

```rust
option.map_or(true, |value| predicate(value))
```

to:

```rust
option.is_none_or(|value| predicate(value))
```

The complete truth table is identical:

- `None`: true
- `Some(value)`: the same predicate result

Both APIs consume the receiver once and invoke the closure exactly once only for `Some`. Outer `&&` short-circuiting, iterator `find`/`filter`/`min` ordering, source-chain evaluation, borrows, false flags, equality operands, absent-source permissiveness, and all literals remain unchanged. Three closures became short enough for rustfmt to collapse; their formatted bodies are otherwise identical.

## Validation

- strict all-target `clippy::unnecessary_map_or` for `gents-desktop-core` and `gents-desktop-bridge`: pass; unrelated baseline warnings remain
- exact bridge same-session agent-scoping test: 1 passed
- bridge snapshot filter: 38 passed
- core store unit filter: 4 passed
- desktop package-boundary check: pass
- `git diff --check`: pass

Broad baseline classification:

- bridge: 97 passed, 18 failed, 2 ignored
- desktop-core lib: 115 passed, 10 failed
- desktop-core `client_store`: 5 passed, 4 failed
- every failure stops at startup with `collection 'AgentDirectoryEntry' not found - add schema before subscribing to P2P`, before any changed predicate executes
- #949 now owns the coherent catalog, root-pin, Defra revision, and readiness foundation

## Scope

No public API, runtime result, ownership or visibility rule, error string, logging, retry behavior, feature gate, or module path changes.

### DIFF
--- a/crates/gents-desktop-bridge/src/snapshot/session.rs
+++ b/crates/gents-desktop-bridge/src/snapshot/session.rs
@@ -116,9 +116,7 @@ pub fn build_session_snapshot_from_store_for_agent(
 ) -> Option<DesktopSessionSnapshot> {
     let conversation = store.conversations.iter().find(|row| {
         row.session_id == session_id
-            && agent_did.map_or(true, |agent_did| {
-                row.agent_did.as_deref() == Some(agent_did)
-            })
+            && agent_did.is_none_or(|agent_did| row.agent_did.as_deref() == Some(agent_did))
     });
     let session_row = store
         .sessions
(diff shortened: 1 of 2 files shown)


jackzampolin:

Issue #994, opened 2026-08-02
Invert the request-lifecycle illegal-transition fence: assert against production writers, not a hand-written mirror table


jackzampolin:
## Problem

`tests/conformance/request_lifecycle.rs` verifies the `illegal` half of the Request contract against a **hand-written writer inventory in the test file itself** (`rust_request_transition_action` / `rust_request_recovery_sweep_writer`), never against production. The assertion for an illegal pair reduces to "the table I wrote has no entry for this pair" — it is tautological, and a production writer that the inventory does not list is invisible to it.

This is not hypothetical: three edges (`claimed -> completed`, `claimed -> dead`, `processing -> dead`) had real production writers while the contract published them as `illegal`. Fixed in #<PR> by classifying them `recoveryReachable` with a boundary citation — but the fence that should have caught the drift structurally could not.

The `legal` half is fine: those cases are driven end-to-end through the real `RequestLifecycle` against an embedded DefraDB and asserted on the persisted `lifecycle_state`.

## Proposed work

Invert the fence. Instead of asking "does my table list a writer for this illegal pair?", enumerate production writers and derive the reachable edge set:

1. Enumerate the production writer entry points (the `RequestLifecycle` transition methods plus the registered recovery sweeps).
2. Drive each writer from every persisted start state against an embedded node, recording the observed `(from, to)` persisted edge — including the no-op case where the CAS filter correctly rejects the write.
3. Assert `observed ⊆ legal ∪ recoveryReachable`, and that every `legal` edge is observed by at least one writer.
4. Keep the generated Lean contract as the source of truth for the partition; delete the hand-maintained mirror table.

This makes the illegal claim a statement about the code, and makes an unlisted new writer fail the build rather than pass unnoticed.

## Also worth folding in

- Drive the two `-> dead` recovery-reachable edges for real. `drive_generated_request_recovery_reachable_case` currently only drives `claimed -> completed`; the `-> dead` edges need a running-bridge + expired-child fixture to exercise `ToolCallLifecycle::reconcile_subagent_liveness`.

## Acceptance

- Adding a production writer for an illegal edge fails `generated_request_transition_cases_cover_lifecycle_policy`.
- The hand-written action mirror table is gone.

Found during a multi-agent review of the Lean↔Rust correspondence.

jackzampolin:
The fence inversion landed in #995 (`abcef1f7`, `fc5127df`, `51d6bac6`). Status of what this issue asked for:

[...]

Issue #993, opened 2026-08-02
Compaction proofs quantify over identity functions; the production summarize reducer is unmodelled


jackzampolin:
## Problem

`Proofs/Compaction/Transition.lean` proves its properties about **identity functions**. The production reducer — the summarize path in `crates/gents/src/compaction/` that actually drops and rewrites transcript rows — is not the function the theorems quantify over. The model therefore proves that *doing nothing* preserves meaning, which is true and not useful.

Consequences observed in review:

1. **Pair-closure is violated in production.** The summarize split can strip a tool-call/tool-result pair asymmetrically, so orphaned tool-result content is dropped from the provider view — a property the model asserts is preserved.
2. **`safeToReduce` has no runtime counterpart.** The gate exists in Lean; the conformance case in `tests/conformance/streaming_compaction.rs` implements the gate *inside the test* rather than calling production, so the test cannot detect its absence.
3. **Compacted-prefix accounting mismatch.** `drop_compacted_prefix` (`agent/daemon/request.rs`) counts in the **unsanitized** message space, while `messages_compacted` was measured in the **sanitized** space. The two can disagree, so the prefix dropped is not necessarily the prefix summarized.

## Why it matters

Compaction is where the durable transcript and the provider view diverge on purpose. It is the one place where "everything persisted can be projected back out" is at risk, and it is currently the weakest link between the model and the code — graded `adequate` where every other reviewed dimension graded `good`.

## Proposed work

1. Model the real reducer: give `Proofs/Compaction` a `summarize` transition parameterised over the actual drop/replace policy, rather than proving over `id`.
2. Restate and prove pair-closure over that reducer — a compacted transcript must not contain a tool result whose call was dropped, or vice versa. Then fix `compaction/history.rs` to satisfy it.
3. Give `safeToReduce` a production counterpart and have the conformance test call it instead of reimplementing it.
4. Unify the counting space for `messages_compacted` / `drop_compacted_prefix` (measure and drop in the same space) and add a regression test that a sanitize-shrinking history cannot cause an over-drop.
5. Verify `run_timeline` reconstruction still round-trips after compaction.

## Acceptance

- Compaction theorems quantify over the production policy, not identity.
- A pair-splitting change to the reducer fails a proof or a conformance test.
- `cargo test -p gents` green; zero `sorry`s maintained.

Found during a multi-agent review of the Lean↔Rust correspondence.

jackzampolin:
[...]

Issue #992, opened 2026-08-02
PromptAssembly has no Lean contract extraction — its conformance fence is social, not mechanical


jackzampolin:
## Problem

`PromptAssembly` is billed in `crates/gents/proofs/README.md` as one of the central proven areas — `sanitize` soundness/fixpoint/idempotence/split-stability over the permissive transcript, plus the `run_loop_stream` entry chokepoint. But unlike `Request`, `Process`, `InferenceCall`, `ManagedExec`, `RuntimeReconcile`, and `SessionRecovery`, it emits **no contract through `Proofs.Conformance.Contracts`**.

Its Rust fence (`crates/gents/tests/conformance/prompt_assembly.rs`) is therefore *social*: a human read the spec and wrote assertions against a test-local oracle. Nothing mechanically ties the Rust sanitizer to the function the Lean theorems quantify over. A change to either side that breaks the correspondence does not fail a test.

This is the exact failure mode already recorded in `agent/loop_stream.rs:689-709`, where repair rewrote only `new_messages` while the poisoned tool-call arguments lived in loaded history: *"The fence described a transform that did not exist."*

## Why it matters

Provider-input correctness is supposed to be a property, not a discipline — the claim is that every completion request is born in the owned loop and sanitized at entry. That claim currently rests on code review rather than on the contract extraction that fences the other machines.

## Proposed work

1. Extend `Proofs/Conformance/Contracts.lean` to emit a `PromptAssembly` contract: the sanitize input/output witness rows (including the tool-call pairing cases), the fixed layer order from `assembleWithContext`, and the repair payload-only/idempotence cases.
2. Add a `consumerCoverage` ledger row in `Proofs/Conformance/CoverageLedger.lean` naming the Rust consumer, and register it in `crates/gents/tests/support/conformance_consumers.rs`.
3. Rewrite `tests/conformance/prompt_assembly.rs` to drive the **production** `sanitize_history_for_provider` / `assemble_new_messages` / `repair_provider_input` against the generated rows, deleting the test-local oracle.
4. Confirm the entry-chokepoint claim mechanically: assert no completion-request construction site bypasses the owned loop (`completion_factory`, `background_completion`, `oneshot`, `inference_http`, compaction summarization, subagent paths).

## Acceptance

- `lake env lean --run Proofs/Conformance/Contracts.lean` emits `PromptAssembly` rows.
- Changing the Rust sanitizer's behaviour fails a conformance test rather than passing silently.
[...]

PR #995, opened 2026-08-02
Fix silent tool-failure persistence and reconcile the Request contract with production writers

jackzampolin:
Fixes found by a multi-agent review of the Lean↔Rust correspondence and the runtime's engineering quality. Three commits, each independently verified.

## 1. Failed tool calls were persisted as successes

`tool_outcome_to_result` rendered a failed tool call as the `ToolError` Display text (`"tool call error: …"`), but `classify_runtime_failure` matched on the `ToolCallError:` / `JsonError:` prefixes that rig's toolset dispatcher used to stamp. When the owned loop took over dispatch (#400, D6) the two ends silently disagreed: classification returned `None`, `on_tool_result` fell through to `lc.complete(...)`, and **every tool call returning `Err` was durably recorded as a successful call** — losing its failed state, its `FailureClass`, and any structured command-policy denial. The run timeline and adapter projections then replayed those failures as successes.

The prefixes are now shared constants next to `ToolError`, referenced by both ends so a rename cannot desync them. The inner error is rendered rather than the wrapper: a policy denial arrives as a `ToolCallError` carrying denial JSON and `parse_command_policy_denial` strips the prefix and parses the remainder, so the wrapper's own text in front of the payload would downgrade every denial to an unstructured failure.

The existing test hand-wrote the string the classifier expected instead of obtaining it from the dispatcher — which is exactly why this passed while production was broken. The three new tests drive both production functions end to end.

## 2. The Request contract claimed edges were illegal that production performs

`claimed -> completed`, `claimed -> dead`, and `processing -> dead` were published as `illegal`, and the conformance test asserted "Rust has no writer path" for each. All three have real writers: `RequestLifecycle::complete` accepts a persisted from-set of `{claimed, processing}`, and the subagent-liveness sweep terminalizes an expired claimed/processing child as `dead`.

They are taken by no single `RequestContext.Action` — recovery sweeps perform them on persisted rows — so they are now classified `recoveryReachable` citing a new `boundary.request.recovery-sweep-reachable`, rather than being folded into the legal transition relation. Counts move from 11/53/17 to 11/50/17 + 3. `claimed -> completed` is driven for real against an embedded node.

Also reconciles the prose: `proofs/README.md` and the dead-preclaim-only boundary both stated `dead` was pre-claim only, contradicting the subagent-liveness sweep shipped alongside them.

## 3. Terminal irreversibility now asserted against production (#994)

The `illegal` branch of the generated-case test consults the hand-written writer inventory at the top of the file, so it reduces to "the table I wrote has no entry for this pair" — which is how the three edges above stayed misclassified.

This drives the real writers for the 40 `terminal -> *` edges covered by S1. For each terminal state it takes genuine ownership of a request (so the local state machine permits the call and the persisted CAS is what decides), has another actor terminalize the row underneath the live lifecycle, then calls every terminal writer and asserts the document did not move — the race the CAS filters exist for. Verified with a negative control.

## Verification

- `cargo test -p gents` — green
- `cargo check --workspace --all-targets` — green
- Lean builds clean; zero `sorry`s maintained

The coverage-ledger and partition-validator fences both failed until the new boundary was registered, which is them working as designed.

## Follow-ups filed

- #992 — PromptAssembly has no Lean contract extraction; its fence is social, with a test-local oracle
- #993 — Compaction proofs quantify over identity functions; the production summarize reducer is unmodelled. Includes two concrete defects: the summarize split can orphan a tool-call/result pair, and `messages_compacted` is measured in `strip(sanitize(H))` space while `drop_compacted_prefix` applies it in `strip(H)` space
- #994 — Invert the illegal-transition fence: enumerate production writers and derive the reachable edge set instead of consulting a mirror table

### DIFF
--- a/crates/gents/proofs/Proofs/Conformance/Boundaries.lean
+++ b/crates/gents/proofs/Proofs/Conformance/Boundaries.lean
@@ -17,6 +17,9 @@ def boundaryRequestInputRequiredReservedId : String :=
 def boundaryRequestDeadPreclaimOnlyId : String :=
   "boundary.request.dead-preclaim-only"
 
+def boundaryRequestRecoverySweepReachableId : String :=
+  "boundary.request.recovery-sweep-reachable"
+
 def boundaryToolCallPermanentWithoutRetryEvidenceId : String :=
   "boundary.tool-call.permanent-without-retry-evidence"
 
@@ -87,7 +90,15 @@ def boundaries : List Boundary :=
     , domain := "RequestLifecycle"
     , subject := "dead terminal state"
     , statement :=
-        "dead is terminal only for stale pre-claim work; post-claim provider, retry, tool, and deadline failures remain failed."
+        "In the request machine dead is terminal only for stale pre-claim work; post-claim provider, retry, tool, and deadline failures remain failed. The subagent-liveness recovery sweep is the one exception: it terminalizes an expired claimed or processing child as dead, published as recoveryReachable under boundary.request.recovery-sweep-reachable."
+    }
+  , { id := boundaryRequestRecoverySweepReachableId
+    , domain := "RequestLifecycle"
+    , subject := "recovery-sweep reachable request edges"
+    , statement :=
+        "claimed->completed, claimed->dead, and processing->dead are taken by no single RequestContext.Action, but registered recovery sweeps perform them on persisted rows: terminal repair completes a claimed request whose response already landed, and the subagent-liveness sweep terminalizes an expired claimed or processing child as dead. They are published as recoveryReachable rather than illegal so the emitted contract does not assert Rust has no writer for an edge the product performs."
+    , acceptedFollowUp :=
+        some "Compose the Request machine with Proofs/Recovery so these edges are proven in one model instead of cited across two."
     }
   , { id := boundaryToolCallPermanentWithoutRetryEvidenceId
     , domain := "ToolExecution"
--- a/crates/gents/proofs/Proofs/Conformance/ContractCases/LifecycleTransitions.lean
+++ b/crates/gents/proofs/Proofs/Conformance/ContractCases/LifecycleTransitions.lean
@@ -9,6 +9,7 @@ inductive LifecycleTransitionClassification where
   | legal
   | illegal
   | productUnreachable
+  | recoveryReachable
   deriving DecidableEq, Repr
 
 namespace LifecycleTransitionClassification
@@ -17,6 +18,7 @@ def toContract : LifecycleTransitionClassification → String
   | .legal => "legal"
   | .illegal => "illegal"
   | .productUnreachable => "productUnreachable"
+  | .recoveryReachable => "recoveryReachable"
 
 end LifecycleTransitionClassification
 
@@ -109,6 +111,21 @@ def requestTransitionAction? (source target : String) : Option String :=
     source
     target
 
+/-- Request edges that no single `RequestContext.Action` takes, but that
+registered recovery sweeps legitimately perform on persisted rows.
+
+`claimed -> completed` is terminal repair finishing a claimed request whose
+response document already landed; `claimed -> dead` and `processing -> dead` are
+the subagent-liveness sweep terminalizing an expired child. The licensing models
+live in `Proofs/Recovery/` — the request machine alone does not model them, so
+publishing these as `illegal` made the emitted contract assert that Rust has no
+writer for edges the product actually performs. -/
+def requestRecoverySweepReachable : RequestState → RequestState → Bool
+  | .claimed, .completed => true
+  | .claimed, .dead => true
+  | .processing, .dead => true
+  | _, _ => false
+
 def requestTransitionClassification
     (source target : RequestState)
     (action : Option String) : LifecycleTransitionClassification :=
@@ -117,6 +134,8 @@ def requestTransitionClassification
   | none =>
       if source = .inputRequired ∨ target = .inputRequired then
         .productUnreachable
+      else if requestRecoverySweepReachable source target then
+        .recoveryReachable
       else
         .illegal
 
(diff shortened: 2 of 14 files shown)


jackzampolin:
All three findings fixed, plus the formatting failure. Two commits: `8cede902` (review fixes) and `abcef1f7` (the fence inversion). `cargo test -p gents` and `cargo check --workspace --all-targets` both green locally.

## [P1] Forgeable failure channel — fixed, and it exposed a second defect

Moved to the collision-free `__gents_tool_lifecycle__:` namespace, as you suggested. Worth noting how right the pointer was: `UNPARSEABLE_ARGS_MARKER`'s doc comment already warns about this exact hazard and names `JsonError:` as its example of a human-readable token ordinary output could carry. The prefixes reintroduced a problem this codebase had already solved and documented one constant above.

Fixing it surfaced a second defect I'd have shipped: the hook strips the unparseable marker before persisting (`prompt_hook.rs:414`), but the new markers weren't stripped, so `__gents_tool_lifecycle__:toolCallError:…` would have landed in the durable transcript and been replayed to the provider on the next request. The hook now classifies while the marker is attached and strips immediately after, with a test asserting no marker reaches model-facing text.

Two honest limits:

- The marker is collision-**resistant**, not unforgeable. Output containing the literal sentinel still classifies, exactly as for the pre-existing managed-terminal and unparseable-args markers. Closing that needs the typed outcome carried through `on_tool_result` — filed as #997 with the full argument, deliberately out of scope here since it changes the hook surface and its thirteen dispatch blocks.
- The `Ok(...)` impersonation predates this PR: `classify_runtime_failure` always ran on the raw string regardless of arm. Not a defence — the change made that channel load-bearing, so fixing it here is right — just so the history is accurate.

## [P2] Wrong writer for claimed → completed — fixed

Registered `repair_terminal_requests` and the probe now drives that sweep with the terminal response it requires, asserting `repaired == 1`.

Your second point stands on its own and is now pinned by `ordinary_complete_also_takes_the_claimed_to_completed_edge`. On whether `complete()` should keep accepting `Claimed`: keeping it, deliberately. A crash between `begin_execution`'s write and the completion would otherwise strand the row, and defensive behaviour wins around anything that can deadlock a request. The rationale is in the test's doc comment so it isn't re-litigated or tidied away. That the edge is reachable from the ordinary path is precisely why it's published as `recoveryReachable` rather than folded into the legal transition relation.

## [P2] Overclaimed terminal test — fixed, then inverted

You were right on both counts, including that `advance()` only issues `update_AgentResponse`, so that arm was tautological. Dropped it, renamed the test to say only what it drives, and added `claim`, `begin_execution`, and the repair sweep. Both new arms negative-controlled: with terminalization disabled, `claim` reports `Claimed` and the sweep reports `repaired: 1`.

But the deeper point — that a broken CAS in an omitted writer would still pass — needed the inversion, so `abcef1f7` does it. `production_request_writers_only_reach_contracted_edges` drives every `RequestLifecycle` writer that mutates `AgentRequest` from all nine persisted start states, reads the edge off the document, and requires the observed set to fall inside legal + recoveryReachable. It observes ten distinct real transitions and nothing outside the contract. Those ten are pinned as a coverage floor so a fixture that stops driving its writer fails loudly rather than satisfying the subset assertion vacuously.

Negative control: un-registering the recovery edge produces `production writers reached claimed -> completed, which the contract classifies illegal` — verbatim the drift that started this.

Three writers remain outside the fence and are named in the test rather than left implicit: `advance`, `queue::reconcile_coalesced_pending_request` (needs a second coalescing duplicate), and `reconcile_subagent_liveness` (needs a running-bridge plus expired-child fixture). #994 stays open for those.

## CI

`cargo fmt --all` failure fixed.

The `desktop-ui` failure is unrelated, and I checked rather than assuming: this branch changes nine files, all Rust and Lean, zero frontend. The Bombadil fuzzer is **unseeded** — `run-bombadil.mjs` passes only `--time-limit` — so each run explores a different random path, and the same job has both passed and failed on `main` and on `agent/add-atif-trace-projection`. Filed as #996 with the evidence and a suggested fix (emit and accept a `--seed` so failures are reproducible at all, then fix the underlying reload/rehydrate race). Re-running the job here.

jackzampolin:
Both fixed in `fc5127df`. `cargo test -p gents` (1699 tests) and `cargo check --workspace --all-targets` green locally.

## [1] Unknown tools persisted as successful calls

Correct, and a real hole — the unknown-tool branch was the one dispatch path I never re-read after adding the marker, and it fires whenever a model invents a tool name, so it is routine rather than exotic. It now returns a marked dispatch failure. The detail text is unchanged, so the model sees exactly what it saw before.

The end-to-end test you asked for takes the string from the **real dispatcher** against an empty tool surface and feeds it through the **real hook**, then asserts the persisted `AgentToolCall` row: `lifecycle_state == "failed"`, the result explains the failure, and no marker leaked into it. Negative control — reverting the production fix — persists:

```
lifecycle_state: "completed", tool_failure_class: null,
result: "error: unknown tool 'ghost_tool'"
```

which is the bug reproduced on demand.

The full suite then caught something my targeted runs had not: `dispatch_tool_calls_known_tool_and_reports_unknown` asserted the exact old string. I updated it to assert the new contract rather than merely loosening it — marker present, detail preserved, and `model_facing_tool_result` still exactly `error: unknown tool 'missing'`, so the model-facing view is provably unchanged.

## [2] The fence short-circuited before several CAS paths

Also correct, and the diagnosis was exact. Local state was derived from the start state, so a pending start left the lifecycle pending and `begin_execution`/`complete`/`fail` failed `ensure_state` before issuing anything, while non-pending starts left it claimed/streaming so `claim` and TTL expiry short-circuited. Each writer is now placed in its own locally-permitted state, with the persisted row forced independently to every start state.

The observed edge set is unchanged at ten. That is the right outcome — what changed is not which edges exist but whether their absence is evidence, since previously most combinations issued no mutation at all.

Worth reporting how the verification went, because my first attempt to confirm your concern **passed** and looked like the fence was still blind. It isn't. The terminal CAS conjoins `status _eq "processing"` with `lifecycle_state _in [...]`, and fixtures force canonical `(status, lifecycle_state)` pairs, so widening only a writer's `lifecycle_state` from-set leaves the edge genuinely unreachable — the fence reporting nothing was correct behaviour, not a miss. A control that actually weakens the CAS (dropping the `status` predicate) makes the test fail with:

```
production writers reached interrupted -> completed, which the contract classifies illegal
```

That sensitivity limit is now documented in the test rather than left as an implied guarantee, so nobody reads it as covering more than it does.

Issue #743, opened 2026-07-16
Flaky: backend_health ModelsListener probes fail under parallel lib suite, pass isolated


jackzampolin:
Observed 2026-07-16 on `feat/edit-file-match-modes-738` (diff touches only edit_file/toolset — nothing near backend_health) while gating PR #740.

`cargo test -p defra-agent --lib` (full parallel suite): `backend_health::tests::cycle_marks_reachable_unknown_backends_promotable` FAILED in 1 of 4 runs on the same tree; immediately passes isolated (0.05s) and in the other 3 full-suite runs.

Probe-cycle test from #640/#644 — likely timing/shared-state sensitivity under parallel load (same class as #730/#441). Filing per the no-tolerated-flakes rule; needs a root-cause pass on its reachability/promotable window rather than a rerun shrug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jackzampolin:
Widening the framing before anyone fixes this in isolation: this is the **third** parallel-suite-only / passes-isolated flake in this class —

- #441 — `trigger_conformance::subscription_reconciles_on_generation_bump` (timed out under parallel suite; closed)
- #730 — `e2e_triggers::app_collection_pairing_fires_event_trigger_via_reconcile` (open)
- #743 — `backend_health::cycle_marks_reachable_unknown_backends_promotable` (this issue)

Three different subsystems (trigger conformance, P2P pairing e2e, backend prober), one failure shape: a test that waits on an asynchronous condition with timing assumptions that hold on an idle machine and break under full-suite CPU contention. That pattern suggests the economical fix is **one systemic pass over how async-condition tests wait** — audit the wait/poll/deadline helpers these tests share (fixed sleeps, too-tight timeouts, poll intervals that starve under load) and move them onto a common condition-based waiting helper with load-tolerant deadlines — rather than three independent root-cause hunts that each rediscover the same lesson.

Suggested scope for whoever picks this up: grep the test tree for fixed-duration sleeps and sub-30s bespoke deadline loops, fix the three named tests via the shared helper as the proof of the approach, and fence the helper itself. If #441's fix already established such a pattern, the pass is 'migrate the stragglers to it.'

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jackzampolin:
Recurrence on the v0.7.0 merge commit `417f1d8707b49d9a2e233b20392c00ca4d4ce5e4` in post-merge main CI:

[...]

PR #999, opened 2026-08-02
PromptAssembly: mechanical Lean→Rust contract, and the four defects it found (#992)

### LINKED ISSUE #992: PromptAssembly has no Lean contract extraction — its conformance fence is social, not mechanical

### LINKED ISSUE #1008: PromptAssembly budget contract pins the threshold conversion only at dyadic percentages
Follow-up from review of #999.

## The gap

`PromptAssembly.Budget` computes the configured threshold budget in `Nat`:

```lean
let configured := witness.contextWindow * witness.thresholdPercent / 100
```

Production computes it in floating point
(`crates/gents/src/agent/daemon/request.rs`):

```rust
let configured_threshold_budget = (context_window as f64 * threshold) as usize;
```

The contract test asserts these agree:

```rust
assert_eq!(configured, case.configured_threshold_budget,
           "{}: configured threshold budget drifted from Lean", case.name);
```

**They do not agree in general.** Truncating `f64` differs from integer
division for percentages that are not exactly representable in binary:

| pct | context window | Lean (`Nat`) | Rust (`f64`) |
| --- | --- | --- | --- |
| 57 | 10,000 | 5,700 | 5,699 |
| 69 | 10,000 | 6,900 | 6,899 |
| 29 | 200,000 | 58,000 | 57,999 |
| 57 | 200,000 | 114,000 | 113,999 |
| 58 | 200,000 | 116,000 | 115,999 |

Every emitted witness uses `thresholdPercent := 75`, and `0.75` is exactly
representable, so the assertion holds on the cases chosen and the contract is
green. The correspondence is pinned on the witness set, not on the parameter
space it appears to cover — the same shape of gap #992 exists to eliminate.

`Budget.lean` is already explicit that "Production owns the floating-point
percentage conversion; the contract cases emit exact percentage-derived budgets
and fence that conversion in Rust." The conversion is fenced only at dyadic
percentages.

## Severity: low, and not a safety issue

`effectiveInputBudget = min(configuredThresholdBudget, providerInputBudget)`,
and `providerInputBudget = contextWindow - maxOutputTokens` is integer
arithmetic on both sides (`Nat` subtraction / `saturating_sub`). The
provider-safety theorem `within_effective_is_provider_safe` rests on the
`providerInputBudget` leg, so the D4F fix holds regardless of this discrepancy.
The blast radius is a ≤1-token shift in where the *configured-threshold* leg
trips, and only when that leg is the smaller of the two.

Not blocking, which is why #999 landed with it.

## Options

1. Move production to integer arithmetic (e.g. carry the threshold as basis
   points and compute `context_window * bp / 10_000`). Makes the Lean model
   exactly right and removes the float from the trigger entirely.
2. Model the truncation in Lean so the contract is true for all percentages.
3. Keep the current behaviour and narrow the claim: state the dyadic-only
   boundary in `Budget.lean`, and add a non-dyadic witness marked as a known
   deviation.

(1) is cleanest — the threshold is a configuration percentage, and nothing
needs sub-integer precision.

## Repro

```
python3 -c "
for pct in range(1,100):
    for win in [10000,200000]:
        if win*pct//100 != int(win*(pct/100.0)): print(pct,win,win*pct//100,int(win*(pct/100.0)))
"
```

jackzampolin:
Closes #992. Closes #1008.

`proofs/README.md` billed PromptAssembly as a proven area, but its fence was
**social**: a human read the spec and asserted against a 60-line test-local
pairing oracle. Nothing tied the Rust sanitizer to the function the theorems
quantify over. `loop_stream.rs::repair_provider_input` records a bug of exactly
that shape — *"The fence described a transform that did not exist."*

This makes the link mechanical. Doing so, and reviewing it, found **four
production defects and four model divergences.** Every one was reproduced before
being fixed.

## Production defects

Three of these put invalid input in front of the provider; the fourth is a
model/production disagreement in the compaction trigger.

| defect | measured | cause | fix |
| --- | --- | --- | --- |
| duplicate call keys in one turn | 2 calls, 1 result | pairing ran through a `HashSet`, so the turn closed with one result while every duplicate call survived | dedup within the turn |
| id reused by a later turn | 2 calls, 1 result | `resolved` was global, so an earlier turn's result resolved a later turn's reuse | scope resolution per turn (`resolved_keys_per_turn`) |
| no output reservation | provider rejected `118,785 + 393,216 = 512,001 > 512,000` | trigger consulted only `context_window × threshold`, ignoring space promised to output | `min(configured, context_window − max_output_tokens)` |
| float threshold budget | 57% of 10,000 → Lean 5,700 vs Rust 5,699 | `(w as f64 * t) as usize` truncates where integer division does not | basis points, exact integer arithmetic |

The first three each left a dangling tool call or an over-budget request — the
precise things `sanitize_history_for_provider` and the compaction trigger exist
to prevent.

## Model divergences

**Three transforms, two modeled.** `normalize_assistant_content_order` had no
Lean counterpart, so "the Rust sanitizer implements the proven function" was
false at the composition level.

**Assistant prose vs. an unresolved call.** Rust keeps the message *with its
text*; `filterCallsBy` dropped the row whole.
`AssistantTurnAccumulator::build_message` writes prose-plus-calls on every turn
that has both, so the spec was unfaithful to production in its **commonest**
shape — the pure-row model could not express the case.

**Empty messages**, dropped by Rust asymmetrically across the two stages, kept
by the model. **Empty message mid-turn**, which does not end the turn in Rust
(`has_plain_content` is false) but did in the model, stranding a valid pair.

Rust is right in every case, so the model moved to Rust and soundness was
re-proven against it.

## The compaction-model alignment

Per-turn resolution also changed `provider_view`, which #998's compacted-prefix
accounting is built on — leaving production contradicting a proven area that had
landed hours earlier. Shipping that drift would have undercut the point of this
PR.

`sanitizeTurn_eq_sanitize` resolves it: per-turn and global resolution are the
**same function** under `UniqueCallIds`, the hypothesis every theorem over
`sanitize` already carries, so Prefix.lean's results transfer rather than needing
re-proof. `reused_call_id_breaks_prefix_stability` is *kept* — it is still true
of the coarser `providerView`, and deleting a true theorem trades one inaccuracy
for another. What was wrong was attributing it to production, so
`providerViewTurn` now names what the runtime computes and
`reused_call_id_is_prefix_stable_per_turn` shows the identical witness is stable
there.

## What landed

**Lean** (zero `sorry`s) — `PromptAssembly/Content.lean` (canonical content
order; `callsOf_normalize` is load-bearing), `PromptAssembly/Provider.lean`
(`ProviderRow`, `Coherent`, `NonDegenerate`, the real three-stage
`sanitizeForProvider`, with soundness / split-stability / fixpoint / idempotence
re-proven), `PromptAssembly/Budget.lean` (`within_effective_is_provider_safe`),
and the alignment lemmas above.

**Contract** — 13 sanitize, 5 layer-order, 7 repair, 12 budget witnesses, every
expected value computed by running the model. `witnessesAreCoherent` and
`witnessesHaveUniqueCallIds` discharge **both** premises of
`sanitizeForProvider_sound` by `decide`, so `witnessOutputsAreProviderValid`
makes the emitted rows' validity a theorem rather than a comment — which is what
lets the Rust fence delete its pairing oracle entirely.

**Rust** — the fence drives production against the rows with the
row↔`Message` projection asserted to round-trip. Two source scans pin the
owned-loop chokepoint: which files may invoke a provider, and which may enter the
loop.

## Documented boundaries

Shapes the row model provably cannot express, fenced in Rust and written down
rather than left implicit: **call multiplicity** (`MessageKind` carries a
`Finset`; closing it means changing the shared `Transcript` type every pairing
theorem is stated over) and the **conditional refinement** to the row-only
`sanitize`, which is false unconditionally.

## Proving the fence fails

Reverting `drop_unpaired_tool_calls` to the old behaviour fails exactly the
intended witness, with the other five tests staying green:

```
generated_sanitize_cases_drive_the_production_sanitizer ... FAILED
sanitize disagrees with the Lean model on case "assistant-prose-survives-its-unpaired-call"
  left:  [user "text-0", user "text-1"]
  right: [user "text-0", assistant "text-7", user "text-1"]
```

## Gates

- `cargo test -p gents` — **1765 passed, 0 failed**
- `cargo test -p gents-cli` — **685 passed, 0 failed**
- `cargo fmt --all --check`, `cargo check --workspace --all-targets` — clean
- `lake build` — clean, zero `sorry`s

Rebased onto current main (past #995, #998, #1006).

## Review notes

- The conditional refinement and the multiplicity boundary are judgement calls;
  push back if you read either differently.
- #1000 tracks two unrelated load-sensitivity flakes (`backend_health`,
  `r6_background_tools`), untouched by this diff. Main's `d021bcb7` may already
  fix the second.
- Pre-existing unused-import warning in `tests/conformance.rs` left alone.

Design doc: `docs/superpowers/specs/2026-08-01-promptassembly-lean-contract-design.md`


jackzampolin:
All five findings addressed. Three were in the model; **two were live defects in
the sanitizer**, both of which left a dangling tool call in provider input —
which is the one thing `sanitize_history_for_provider` exists to prevent. Each
was reproduced before fixing.

## Round 1

**Empty messages.** Confirmed: Rust drops both empty assistant and empty user
messages, asymmetrically across the two stages. Modeled via `emptyUserRow` /
`emptyAssistantRow`, with `NonDegenerate` as the invariant the prunes establish;
the fixpoint theorems take it as a hypothesis. This also made
`project_dropOrphanedFromP` conditional — it is now false unconditionally — so
soundness uses two set-level facts proven directly instead, which hold without
the hypothesis because an empty row carries neither calls nor results.

**Duplicate call ids — production defect.** Measured 2 calls against 1 result:
pairing runs through a `HashSet`, so a turn announcing one id twice was closed
by a single result while both calls survived. Fixed by deduplicating within the
turn.

On modeling multiplicity itself I'm pushing back on **scope, not substance**:
`MessageKind.assistantToolCalls` carries a `Finset` in the shared `Transcript`
type that every pairing theorem in `Transcript` and `PairingReconcile` is stated
over. Replacing it is a much larger change than this PR. The fix makes the set
abstraction *true* of production output rather than merely assumed, the
occurrence-level behaviour is fenced in Rust, and the boundary is now written
down in `Provider.lean`. Happy to file the `Finset` → multiset work separately.

**`UniqueCallIds` premise.** Correct — matching emitted output did not
mechanically inherit validity, because the witnesses discharged only one of the
two premises. `witnessesHaveUniqueCallIds` now discharges the other by `decide`
(needed new `Decidable` instances for `Disjoint` and `UniqueCallIds`), and
`witnessOutputsAreProviderValid` derives validity from both.

## Round 2

**`pending` cleared on an empty user message.** Confirmed by probe: for
`[assistant c1, empty user, result c1]` Rust keeps the pair, the model dropped
both. Root cause exactly as described — `has_plain_content` is false, so
`pending_calls.clear()` never runs. Now recurses with `pending`, which
*simplified* soundness: that branch is literally the induction hypothesis at the
same `pending`. Witness `empty-message-does-not-break-an-open-turn` emits 3 → 2.

**Global `resolved` — production defect, and it predates the dedup.** Probe on
`[assistant c1, result c1, assistant c1]` gave 2 calls / 1 result. Worth being
precise: the round-1 dedup neither caused nor fixed this. And you were right
about the test — `call_key_reuse_across_turns_survives` passed only because both
turns were answered, so it never exercised the broken path. That was a weak
test.

Fixed by scoping resolution per turn (`resolved_keys_per_turn`), mirroring the
`pending_calls` reset `drop_orphaned_tool_results` already does, including that
plain content ends the turn *after* results riding in the same message.

The Lean side keeps a global `resolvedIn`, and I've documented why rather than
papering over it: global and per-turn coincide **exactly under `UniqueCallIds`**
— the hypothesis of `sanitizeForProvider_sound`, which forbids cross-turn id
reuse and which every witness now provably satisfies. They diverge only on
reused ids, which the theorems exclude but loaded history does not, so Rust must
be correct without the precondition.

## Regressions added

- `duplicate_call_keys_in_one_turn_do_not_leave_a_dangling_call`
- `call_key_reuse_across_turns_survives` (both turns answered)
- `incomplete_second_turn_reusing_a_key_is_not_resolved_by_the_first`
- `empty_message_between_a_call_and_its_result_does_not_break_the_pair`
- `plain_message_between_a_call_and_its_result_ends_the_turn`

Plus three new contract witnesses for the empty-message cases.

## Gates

- `cargo test -p gents` — **1690 passed, 0 failed**
- `cargo check --workspace --all-targets` — clean
- `lake build` — clean, zero `sorry`s

One caveat worth flagging: an earlier full-suite run hit an unrelated
intermittent failure in
`backend_health::cycle_reproduces_fleet_evidence_dead_backend_demotes_and_recovers`
(passes 5/5 in isolation; the next full run was 1690/1690). It is untouched by
this diff. Filed with evidence as #1000 rather than guessed at — it is the same
class as the already-fixed #743, and a load-dependent networking flake deserves
a real reproduction before a fix.

PR #1005, opened 2026-08-02
Desktop fuzz: exact trace replay, honest surface invariants, teardown fix (#996)

### LINKED ISSUE #996: Flaky: desktop-ui Bombadil fuzz trips desktop_primary_surfaces_are_not_empty (transcript panel) — unseeded, unreproducible
## Symptom

`desktop-ui` / "Run desktop Bombadil smoke" intermittently fails the
`desktop_primary_surfaces_are_not_empty` invariant with:

```
emptyPrimarySurfaceProblems =
  - "[data-testid=\"transcript-panel\"]"
```

Seen on https://github.com/source-inc/gents/actions/runs/30732894860/job/91456244874
(PR #995), whose diff touches **zero** frontend files — 9 files, all Rust under
`crates/gents` plus Lean under `crates/gents/proofs`.

## Evidence it is a flake, not a regression

- The fuzzer is **unseeded**. `apps/gents-desktop/tests/bombadil/run-bombadil.mjs`
  passes only `--time-limit` (`30s` in CI, `BOMBADIL_TIME_LIMIT ?? "20s"` otherwise);
  there is no `--seed`. Each run explores a different random click path, so
  coverage — and therefore failure — varies run to run.
- The same job has failed on `main` itself, and has both passed and failed on the
  same branch (`agent/add-atif-trace-projection`) across different runs.
- The failing path reloads the page twice, switches agent, then sends a message —
  consistent with a race between a reload and the transcript subscription
  rehydrating, rather than a content bug.

## Why it matters

An unseeded fuzzer in a required check makes every PR's CI a coin flip and trains
reviewers to ignore red. `CLAUDE.md` is explicit: *"Flaky tests are defects. The
formal-verification investment exists to eliminate that class; capture, file, and
fix — never shrug."*

## Suggested fix

1. **Make failures reproducible.** Have `run-bombadil.mjs` generate a seed, print
   it, and accept `--seed` so a CI failure can be replayed exactly. Currently the
   only reproduction handle is the `--reproduce` artifact directory on the runner,
   which is lost when the runner is recycled.
2. **Fix the underlying race.** Determine whether an empty transcript panel is
   legitimate mid-rehydrate after a reload. If so, the invariant needs to wait for
   a settled state rather than asserting continuously; if not, the panel is
   genuinely dropping its subscription on reload and that is a real desktop bug.
3. Until it is deterministic, consider whether `desktop-ui` should block merge.

## Repro

```
cd apps/gents-desktop
npm run test:ui:fuzz      # re-run until it trips; unseeded, so expect variance
```

jackzampolin:
Fixes #996 (`desktop-ui` Bombadil smoke intermittently tripping `desktop_primary_surfaces_are_not_empty` on the transcript panel, unreproducibly).

## 1. Reproducibility: the trace is the seed

Bombadil 0.6.1 (latest) has **no `--seed`** — its exploration RNG lives in the host binary (verified against the CLI surface and binary; the issue's suggested fix assumed a knob that doesn't exist). The exact-replay handle it does provide is the recorded trace, so this PR makes that handle durable and prominent:

- Every run (pass **and** fail) prints a `BOMBADIL REPRODUCTION HANDLE` block with the trace path and the exact `npm run test:ui:fuzz -- --reproduce <dir>` command.
- CI uploads `test-results/bombadil` as the `desktop-ui-bombadil-trace` artifact on **every** run (30-day retention), so a failure can be bisected against a known-good trace after the runner is recycled.
- `--seed` is rejected with an explanation pointing at `--reproduce`.
- Fixed `--reproduce` through the wrapper, which had **never worked**: the wrapper unconditionally appended `--exit-on-violation`, which Bombadil rejects alongside `--reproduce`. Replay is now verified end-to-end.

## 2. The invariant was wrong — with evidence, not by fiat

The transcript panel renders a **designed, text-free loading skeleton** (`role="status"`, "Loading conversation") between selecting a session and its snapshot arriving. Instrumenting the harness with mutation records + a task-boundary sampler shows:

- At full speed the skeleton never even spans an observable task boundary (0 empty samples across ~6.7M task-boundary samples over reloads, conversation switches, and sends).
- Under 20× CPU throttling (≈ the loaded shared CI runner) the skeleton **commits** on the send-into-new-session and conversation-switch paths — exactly the preamble of the failing CI run (reload ×2 → switch → send) — and settles immediately after.
- There is **no dropped subscription**: session loading is selection-driven refetch (`refreshSession` effect), the panel settled in every observed case, and 25/25 unseeded baseline runs were green locally.

So the old invariant asserted "non-empty text, continuously" against a state the app never guaranteed. It is replaced by two properties that state the actual guarantees, **without losing detection**:

- `desktop_primary_surfaces_are_never_blank` — a mounted primary surface with no text and no declared loading state (`[role="status"]` / `[aria-busy="true"]`) fails **immediately**. Stronger than before for genuine blank-render bugs.
- `desktop_primary_surfaces_settle` — `eventually(...).within(5, "seconds")`: an empty surface must produce content within the bound, so a skeleton stuck past 5s (the dropped-subscription failure class) still fails.

Fence-tested: a synthetic permanently-blank panel violates the first, a synthetic permanently-stuck skeleton violates the second (both exit 2). Also verified empirically that Bombadil forgives a *pending* `eventually`/`within` obligation at test end (exit 0) and only violates an *expired* bound — the settle property cannot itself flake at the time limit.

## 3. Teardown flake found by the verification sweep, and fixed

Running the new spec 25×, one **green** run exited 1: `stopProcess` threw `process group did not drain within 2000ms after SIGKILL` while cleaning up after Bombadil had already exited 0. Two defects fixed in `runner-control.mjs`:

- With the leader already exited, the SIGTERM grace raced an already-settled promise → surviving Chrome children got **zero** grace before SIGKILL. Now the group is polled for the full grace window, and SIGKILL is skipped if it drains.
- A drain timeout threw unconditionally, letting a reap-pending straggler overturn the run's verdict. When the leader exited on its own, we now warn loudly and leave cleanup to the OS; the watchdog path (leader still alive) still throws, so a hung browser can never leak into a retry.

Both behaviors pinned with new unit tests (10/10 passing).

## Verification

- `npm run test:ui:fuzz` sweep on the new spec: 24/25 green, the single failure being the teardown defect above (reproduced, fixed, covered by tests); follow-up runs on the fixed teardown green. Baseline old-spec sweep: 25/25 green locally (the original flake needs CI-runner latency to become observable, consistent with the timing analysis).
- Desktop unit suite: 333 passing; Prettier clean.

## Recommendation on merge-blocking (issue item 3)

Keep `desktop-ui` blocking. The nondeterminism was never in what the job *tests* — it was (a) an invariant asserting a guarantee the app doesn't make and (b) a teardown race, both now fixed, and any future failure ships a downloadable trace that replays exactly. An unseeded explorer that only fails honestly and reproducibly is exactly what a required check should be. If a new flake class appears despite this, demote it then — with its trace attached to the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -364,6 +364,18 @@ jobs:
         working-directory: apps/gents-desktop
         run: npm run test:ui:fuzz -- --time-limit 30s
 
+      # Bombadil has no --seed; the recorded trace is the only exact-replay
+      # handle. Upload it on success too so a failing run can be bisected
+      # against a known-good trace (npm run test:ui:fuzz -- --reproduce <dir>).
+      - name: Upload Bombadil trace
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: desktop-ui-bombadil-trace
+          if-no-files-found: ignore
+          retention-days: 30
+          path: apps/gents-desktop/test-results/bombadil
+
       - name: Upload desktop UI artifacts
         if: failure()
         uses: actions/upload-artifact@v4
--- a/apps/gents-desktop/tests/bombadil/run-bombadil.mjs
+++ b/apps/gents-desktop/tests/bombadil/run-bombadil.mjs
@@ -11,8 +11,24 @@ import { runWithWatchdogRetry, stopProcess } from "./runner-control.mjs";
 
 const rootDir = resolve(dirname(fileURLToPath(import.meta.url)), "../..");
 const args = process.argv.slice(2);
+if (hasOption(args, "--seed")) {
+  console.error(
+    [
+      "[bombadil] Bombadil (0.6.x) has no PRNG seed; exploration randomness lives in the",
+      "[bombadil] host binary and cannot be replayed from a number. Exact replay uses the",
+      "[bombadil] recorded trace instead — every run writes one and prints it when it exits:",
+      "[bombadil]",
+      "[bombadil]   npm run test:ui:fuzz -- --reproduce <output-directory>",
+      "[bombadil]",
+      "[bombadil] CI uploads every run's trace as the 'desktop-ui-bombadil-trace' artifact",
+      "[bombadil] (on success too, so failures can be bisected against a known-good trace).",
+    ].join("\n"),
+  );
+  process.exit(2);
+}
 const headed = consumeFlag(args, "--headed");
 const keepRunning = consumeFlag(args, "--no-exit-on-violation");
+const replayRequested = hasOption(args, "--reproduce");
 const port = await choosePort(process.env.BOMBADIL_VITE_PORT);
 const origin = `http://127.0.0.1:${port}`;
 const harnessUrl = `${origin}/tests/ui-harness/harness.html`;
@@ -20,6 +36,7 @@ const defaultOutputPath = process.env.BOMBADIL_OUTPUT_PATH
   ? resolve(process.env.BOMBADIL_OUTPUT_PATH)
   : resolve(rootDir, "test-results", "bombadil", String(Date.now()));
 let bombadilProcess = null;
+let lastAttemptOutputPath = null;
 
 const vite = spawn(
   process.execPath,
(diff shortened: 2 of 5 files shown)


jackzampolin:
Verification sweep complete: **24/25 green** on the new spec (30s each, unseeded exploration). The single failure was run 5 — the pre-existing teardown race (`process group did not drain within 2000ms after SIGKILL` on a run whose fuzz phase passed with zero violations), fixed in 67d3b798 and pinned by two new unit tests. Runs 10–25 executed against the fixed teardown: **16 consecutive green**. Combined with the 25/25 baseline sweep, that's 49 green fuzz runs today with every failure accounted for, reproduced, and fixed.

PR #1006, opened 2026-08-02
Close tool-call CAS gaps (#1002); carry tool outcomes as typed data (#997)

### LINKED ISSUE #997: Carry tool outcomes as typed data, not sentinel-encoded strings — successful output can still forge a failure
## Problem

A tool call's outcome reaches the persistence layer as a **`String`**. The hook
(`DefraSessionHook::on_tool_result`) receives `result: &str` and has to
reverse-engineer what happened by pattern-matching sentinels embedded in that
text:

- `__gents_tool_lifecycle__:timedOut` / `:cancelled` — managed terminals
- `__gents_tool_lifecycle__:unparseableArgs` — argument parse failure
- `__gents_tool_lifecycle__:toolCallError` / `:jsonError` — dispatch failure (added in #995)

Every one of these is a workaround for the same missing thing: the dispatcher
already *knows* the typed outcome (`Result<String, ToolError>` plus the managed
terminal and deadline/cancel state) and then throws that type away, forcing the
hook to recover it from prose.

## Why this is worth fixing

**1. The channel is forgeable.** Tool output is untrusted arbitrary text. The
markers are collision-*resistant*, not unforgeable: a tool whose output contains
the literal `__gents_tool_lifecycle__:` sentinel — a log tail, a source listing, an
MCP or subagent relay quoting one, or a deliberate forgery — still classifies. It
can currently fabricate a `failed` lifecycle state and, via
`parse_command_policy_denial`, a **structured command-policy denial** that never
happened. That corrupts lifecycle and audit state, which is exactly the state the
run timeline and adapter projections replay.

This was found in review of #995. That PR closed the realistic half (a
human-readable `ToolCallError:` prefix, which ordinary output forges by accident)
by moving to the collision-free namespace. The residual — deliberate forgery of
the sentinel itself — cannot be closed without a typed channel.

**2. Markers must be manually stripped at every boundary.** They are internal
bookkeeping and must never reach the provider or the durable transcript. Today
that is enforced by remembering to call `model_facing_tool_result` /
`unparseable_args_notice` at each site. #995 shipped a bug of exactly this kind
during development: the new markers classified correctly but were not stripped in
the hook, so they would have landed in the persisted transcript and been replayed
to the provider on the next request. Caught in self-review — but the design makes
that class of bug easy to write and invisible until it reaches a model.

**3. It scatters classification logic.** `on_tool_call` is ~260 lines of thirteen
near-identical copy-pasted dispatch blocks, and outcome classification is spread
across `classify_managed_tool_result`, `unparseable_args_notice`,
`tool_dispatch_failure`, `classify_runtime_failure`, and `classify_runtime_error`
— several of which re-parse the same string.

## Proposed shape

Carry the outcome as data alongside the text:

```rust
enum ToolOutcome {
    Completed(String),
    Failed { class: FailureClass, denial: Option<CommandPolicyDenial>, text: String },
    TimedOut { deadline_at: Option<DateTime<Utc>> },
    Cancelled,
}
```

- `dispatch_tool` returns `ToolOutcome` instead of a marker-encoded `String`.
- `on_tool_result` takes `&ToolOutcome`; the `force_argument_invalid` /
  `classify_runtime_failure` branch chain collapses into a match.
- Model-facing text becomes a single accessor, so stripping is structural rather
  than a call each site must remember.
- The `__gents_tool_lifecycle__` markers and their parsers are deleted.

Then successful output cannot impersonate a failure at all, because the outcome no
longer travels in a channel tool output can write to.

## Scope

Touches the `on_tool_call`/`on_tool_result` hook surface and its callers, so it is
deliberately **not** bundled into #995. Worth doing alongside (or before)
collapsing the thirteen duplicated dispatch blocks, since both edit the same
region.

Found during a multi-agent review of the runtime, and confirmed by review of #995.

### LINKED ISSUE #1002: Runtime takes two tool-call transitions the Lean model proves impossible (timeout CAS, fabricated ChildTerminal::Dead)
Two confirmed defects where the runtime contradicts a **proven** property. Found in a multi-agent review, each independently verified.

## 1. `timeout()` lacks the running-state CAS every sibling transition has

`tool_call_lifecycle/transition/native.rs:393-411` filters only on `_docID`:

```
filter: {{ _docID: {{ _eq: "{escaped_doc_id}" }} }}
```

Every other terminal transition CAS-guards on `running` and adopts the durable terminal when it loses the compare — `complete` (native.rs:160-163), `fail` (native.rs:244-247), `cancel_during_run` (bridge.rs:350-353), `bridge_complete`/`bridge_failure`, and `recover_tool_call_row` (recovery.rs:1791-1794). `complete`'s own comment states the intent: *"Interrupt/timeout won the race — adopt the durable terminal."*

So `timeout()` can overwrite a row another actor already drove to a terminal state, violating `terminal_irreversible` (state **and** failureClass preserved from terminal states).

Concrete race: a native tool is running with an expired deadline while its parent request is interrupted; the periodic `reconcile_terminal_parent_owned_tools` sweep terminalizes the row, and the timeout path then overwrites that terminal with `timedOut`, losing the recorded cancellation cause.

Fix: add the `lifecycle_state: { _eq: "running" }` clause and the lost-compare adoption path the siblings use.

## 2. Parent-deadline expiry fabricates `ChildTerminal::Dead` for a live child

`hook.rs:723-725` (`timeout_expired_tool_calls`) and the parent-deadline arms in `hook/persistence/subagent_bridge.rs:132` and `:218` terminalize a foreground subagent bridge with `bridge_failure(ChildTerminal::Dead)` **without loading any child terminal evidence** — the child request may still be running.

The evidence-checked path exists immediately alongside it (`subagent_bridge.rs:239-326`: `load_child_terminal_row` → `child_request_completed` / `project_child_terminal`); these deadline arms bypass it.

This contradicts the model twice over:
- `Proofs/Background/Transition.lean:65-67` requires `h_second_term : pre.terminalOf.isFailure` for `bridge_failure`, and `Bridge.lean:17-25` maps a non-terminal child to `.running`. The transition is excluded for a live child.
- It lands in `.failed`, whereas the `ToolExecution` model's legal deadline transition is `.timedOut`.

Fix: load child terminal evidence before projecting, or use the deadline transition the model licenses (`.timedOut`) and leave the child's own terminalization to the liveness sweep.

## Why these are high value here

Both are cases where production takes a transition the Lean model proves impossible — the exact class the proof investment exists to eliminate, and the same failure shape as the request-contract drift fixed in #995 (edges published as illegal that production actually performed). Following the repo flow, check first whether the model is right and the code is wrong (likely) or whether the model needs to admit the transition, and drive the fix from there.

Both should get conformance coverage that exercises the real writers, not hand-written state tables — see `production_request_writers_only_reach_contracted_edges` in `tests/conformance/request_lifecycle.rs` for the pattern.

jackzampolin:
Fixes #1002. Closes #997 (bundled here at the operator's direction — the two edit the same `hook/persistence` region, and doing them together avoids the conflict the issue warned about).

## #1002 — two transitions the Lean model proves impossible

Method per the repo flow: **direction decided in Lean first**. In both cases the model is right and the code was wrong; zero Lean changes, `lake build` clean, zero `sorry`s.

### 1. `timeout()` lacked the running-state CAS

`ToolExecution.Transition.timeout` requires `pre.state = .running`; the Rust mutation filtered on `_docID` alone while every sibling terminal transition CAS-guards on `running` and adopts the durable terminal on a lost compare. Fix: add the `lifecycle_state: "running"` clause + `sync_after_lost_running_compare`, returning the compare outcome (`Result<bool>`) like the bridge transitions.

**Regression (watched failing first):** `timeout_adopts_terminal_written_by_terminal_parent_sweep` (`tests/conformance/tool_execution.rs`) drives the real writers against an embedded node — a native tool running past its deadline under an interrupted parent, `reconcile_terminal_parent_owned_tools` terminalizes first (`cancelled`/`interrupted`), then the straggler `timeout()` fires. Before the fix it overwrote the terminal with `timedOut`/`deadline`; now it loses the compare and adopts, preserving state **and** cause.

### 2. Parent-deadline expiry fabricated `ChildTerminal::Dead` for a live child

`bridge_failure` is licensed only with observed child failure evidence (`h_second_term : pre.terminalOf.isFailure`, `Bridge.lean` maps a non-terminal child to `.running`). The issue's suggested check confirmed: **the model already licenses this case as the tool-leg `timeout`** — `ToolExecution.Transition.timeout` has no `childRequestId` guard (unlike `complete`), `ComposedState.tool_step` lifts it into the composed machine, and `coherent_tool_deadlineExceeded_iff_request_deadlineExceeded` equates the bridge row's deadline with the parent's. So the fix is the one the issue anticipated: the deadline arms (`hook.rs` expiry sweep, both `subagent_bridge.rs` await loops, and the same arm in the workflow fan-out await) now call `timeout()` and leave the child's terminalization to the subagent-liveness sweep. Payload wording is unchanged (row `timedOut` already mapped to payload status `"dead"`).

**Regression (watched failing first):** `parent_deadline_sweep_times_out_foreground_bridge_without_child_evidence` (`hook/tests.rs`) drives `timeout_expired_tool_calls` over a live (`processing`) child: before the fix the bridge landed `failed`; now `timedOut`/`deadline`, child untouched, and the open-deadline negative control stays `running`.

## #997 — typed `ToolOutcome` channel

- `ToolOutcome { Completed, Failed { class, denial, text }, TimedOut, Cancelled }`; `dispatch_tool` returns it, and the new `call_tool_managed` executor owns the deadline/cancel envelope for background executions. Classification (including command-policy-denial parsing) happens once, at dispatch time.
- `on_tool_result` takes `&ToolOutcome`; the `classify_managed_tool_result` / `unparseable_args_notice` / `classify_runtime_failure` / `model_facing_tool_result` chain collapses into a match. Model-facing text is a single accessor, so stripping is structural rather than a call each site must remember (#995 shipped a bug of exactly that kind during development).
- The `__gents_tool_lifecycle__` markers, their parsers, and the `RuntimeManagedTool` wrapper are **deleted**. The in-tool marker returns (fs-runner, bash) become plain text; the envelope's `biased` select polls the fired cancellation token / elapsed deadline before the tool's ready result, so the managed terminal wins deterministically (unit-fenced by `managed_call_prefers_fired_cancellation_over_ready_output`).
- This also closes a latent leak: the fan-out composite's Skip payload could previously return a raw `cancelled` marker into the threaded transcript.

**Forgery fence:** `successful_tool_output_cannot_impersonate_any_failure` (classifier) and `forged_lifecycle_sentinel_in_tool_output_persists_as_completed` (end-to-end through the hook: row `completed`, no fabricated failure class or denial columns, text verbatim). A literal fail-first run wasn't possible for these — the fix changes `on_tool_result`'s signature, so the tests don't compile against the old code — but the pre-fix behavior is documented in the code itself: the deleted helpers test stated "the marker is collision-RESISTANT, not unforgeable … closing that last gap means carrying the typed outcome through the hook".

## Honest residuals (out of scope, same shape, different licensed transitions)

- `fail_in_flight_tool_calls` synthesizes `ChildTerminal::Failed` for bridges without child evidence (request-failure path).
- Workflow sibling-abort (`cancel_workflow_bridges`) and the unclaimed-cross-deployment-spawn arm still fabricate `Dead`; their licensed targets are cancellation and the recovery sweep's `failed/serviceUnavailable` respectively, and each needs its own model decision (the #995 `recoveryReachable` kind of judgement).

## Gates

- `lake build` clean, zero `sorry`s (no model changes needed — the model was right in both directions).
- `cargo fmt --all -- --check` clean.
- `cargo test -p gents` (full package) and `cargo check --workspace --all-targets` running at time of PR creation (shared build lock with a sibling session); results to be confirmed on this PR — CI enforces both.


jackzampolin:
Gate confirmation (local, at d021bcb7):

- `cargo test -p gents` — **pass**, every target green (1192 lib, 326 conformance, 99 e2e_subagent, 0 failures anywhere)
- `cargo check --workspace --all-targets` — **pass**
- `cargo fmt --all -- --check` — **pass**
- `lake build` — clean, zero `sorry`s (no model changes needed)

Two follow-up commits since the PR opened:

1. `c7d03160` — the e2e fence `foreground_spawn_subagent_parent_deadline_marks_bridge_dead` pinned the pre-fix fabricated `failed` row; updated to the licensed `timedOut` (payload status `"dead"` unchanged) and renamed `..._times_out_bridge`.
2. `d021bcb7` — `panicking_background_tool_terminalizes_and_notifies` flaked once under cross-worktree load during the first full-suite run. Captured and root-caused: the test read messages **once** after polling the row to terminal, but the completion notification is a separate mutation after the terminal write (production models the window as `completionPending:tool_panicked` + the side-effect recovery sweep). Pre-existing race, not introduced by this branch. Fixed by awaiting via the shared `wait_for_tool_completion_message` helper, whose fixed 500ms retry budget is widened to a 5s deadline. Verified with 12 concurrent full-suite binary runs — the same contention that reproduced the failure on iteration 1 before the fix.

Issue #937, opened 2026-07-29
Lean → Rust: background work for subagents and native long-running tools


jackzampolin:
## Summary

Close the gap between the formal backgrounding models and production Rust so **both**:

1. **Background subagents / agents** (R5-style cross-deployment `spawn_subagent` bridges, `await_mode = "background"`), and  
2. **Native long-running tools** (R6-style bash and other background tool executions, registries, partial output, restart interrupt)

…are driven and conformance-tested from the Lean (and where relevant TLA+) specs, not only from ad hoc integration tests.

This is **not** a harness-only cleanup. [#936](https://github.com/source-inc/gents/pull/936) made R5 crash scenarios honest; it deliberately did **not** re-implement backgrounding semantics from Lean. That larger foundation-flow work is this issue.

## Why

- Background work is load-bearing for multi-agent orchestration and long-running tool sessions.
- Recovery and cancel semantics differ by kind:
  - **Background subagent bridges** stay running across restart while a live parent waits; completion is projected from durable child terminals.
  - **Native background tools** are interrupted on restart when the parent is still live (`interrupted_on_restart`), with completion notifications and coalesced wakes.
- Partial Lean/TLA coverage exists (especially subagent bridge projection/cancel). Native tool backgrounding and a unified “background work” story are under-specified or under-bound in Rust conformance relative to the models.
- Without Lean → conformance → Rust binding, regressions can pass local e2e while drifting from the proven contract.

### In scope

**A. Formal surface (start here)**

- Audit and extend Lean under `crates/gents/proofs/Proofs/Background/`:
  - Existing: `Bridge`, `Transition`, `State`, `Executable`, properties (`Projection`, `Cancellation`, `Unique`, `Budget`, `Foreground`, `DelegationGraph`, `Structure`)
  - Gap-fill for **native background tool** lifecycle (spawn, running, partial output, terminal, cancel, restart interrupt) as a first-class model or a clearly factored submodule—not only subagent bridges
- Align Recovery sweeps that touch background work (`Proofs/Recovery/Sweeps/SubagentLiveness.lean`, tool-call recovery contracts) with both kinds
- TLA where cross-deployment/crash is the right tool:
  - `SubagentCompletion`, `SubagentCancelPropagation` (already green under default configs)
  - Decide whether native-tool background needs a dedicated TLA model or is fully covered in Lean + single-node recovery tests
- Zero `sorry`s; model-check any TLA changes and report states/depth/properties

**B. Conformance witnesses**

[...]

PR #1007, opened 2026-08-02
fix: admission slot accounting — waiter leak, crash-orphan recovery order, drain race (#1001)

### LINKED ISSUE #1001: Admission slot accounting: queue-waiter leak, crash-orphaned running calls survive a session, drain race
Three confirmed defects in admission/slot accounting, found in a multi-agent review and each independently verified. They share a subsystem and should be fixed together.

## 1. Queue-waiter slot leak on persist failure (medium)

`admission/controller.rs:111` — `try_enter_queue()` increments `waiters` via CAS. The only decrement is `leave_queue()` in `QueuedCallGuard::drop` (`controller.rs:303`), but the guard is not constructed until `controller.rs:166` — *after* `persist_call_queued(...).await.map_err(...)?` at lines 163-165.

On a persistence error the `?` returns early with `waiters` still incremented and no guard in existence, so that unit of queue capacity is never released for the lifetime of the controller. `try_enter_queue` rejects when `current >= max_queue_depth` (line 238), so each leaked waiter permanently shrinks the queue; `max_queue_depth` persistence failures wedge the backend into permanent `QueueFull` (lines 148-156) even when it is completely idle.

Fix: acquire the guard before the fallible persist, or release on the error path.

## 2. Crash-orphaned running InferenceCall rows survive a full session (medium)

`admission/recovery.rs:150-166` — `recovery_outcome` returns `None` unless the parent request is interrupted or terminal (`if !request_is_terminal(parent) { return None; }`).

At startup, `agent/runtime/startup.rs` calls `InferenceCall::recover_all` (line 661) **before** `RequestLifecycle::recover_all` (line 681). After a crash the parents of stale running calls are exactly the ones in `claimed`/`processing` — non-terminal — so every crash-orphaned call is skipped. The request sweep then terminalizes those parents moments later, but the inference-call sweep never runs again: `periodic_recovery.rs:69-97` registers only request-terminal-repair, subagent-liveness, and terminal-parent-tool sweeps, and Lean pins the inference-call sweep cadence to `.startup`.

Net effect: orphaned `running` rows persist until the *second* restart. Since reconstructed slot counts derive from persisted running call rows (`Proofs/InferenceCall/SlotAccounting.lean`), backend capacity is understated for the entire uptime following a crash.

Fix: order request recovery before inference-call recovery at startup, or register a periodic inference-call sweep (and update the Lean cadence accordingly).

## 3. `is_drained()` race allows transient over-admission (low)

`admission/controller.rs:61-63` — `is_drained()` reads the `running` counter, which is only incremented *after* semaphore acquisition (`controller.rs:217` after `try_acquire_owned` at line 84; `controller.rs:193` after `acquire_owned().await` at 172). A permit-holder inside the acquire→increment window is invisible to drain detection.

A concurrent `AdmissionRegistry::reconcile` (`registry.rs:73-77`: `active.close(); if active.is_drained() { install new controller }`) can therefore treat a closed controller as drained while a permit is live. `Semaphore::close` does not revoke issued permits and the replacement controller starts with a fresh `Semaphore::new(max_concurrent)`, so the backend can briefly carry `max_concurrent + K` concurrent calls — violating the capacity invariant S7 proves.

Fix: count permits at acquisition intent rather than after, or have drain detection consult the semaphore's outstanding permits.

## Why these matter together

S7/S8/S9 (`capacity_invariant_preserved`, `slot_accounting_preserved`, `terminal_implies_released`) are proven, and the README names "scheduler slot leaks" as a bug class the proof work exists to eliminate. All three defects live in the gap between the model and the runtime bookkeeping the model abstracts over — worth checking whether the Lean slot model should be extended to cover queue-waiter accounting and controller replacement, not just running-row counts.

jackzampolin:
Fixes the three verified defects from #1001. All three live in the gap between the Lean slot model (persisted running rows) and the in-memory bookkeeping the model abstracted over, so the model was extended first, then the Rust brought into conformance.

## Lean (model first)

- **`Proofs/InferenceCall/ControllerBookkeeping.lean`** — models one call's path through `BackendAdmissionController::acquire` as a phase machine with per-phase waiter / in-flight / semaphore-permit contributions.
  - `permit_implies_in_flight` + `drained_no_outstanding_permits` + `replacement_preserves_capacity`: a drained controller holds no permits — including one assigned to a parked waiter — so installing a replacement preserves the capacity bound **only** when in-flight is counted from acquisition intent.
  - `late_admission_counting_unsound`: the pre-fix counting (increment after acquisition) is rejected by the model, not merely unmentioned.
  - `persist_error_releases_waiter` / `terminal_phase_releases_bookkeeping`: every terminal admission outcome — including queued-persist failure — holds no waiter or in-flight unit (the in-memory S9 analog).
- **`Proofs/Recovery/StartupOrder.lean`** — models the parent gate of `InferenceCall::recover_all`. `inference_before_request_leaves_call_live` states defect 2 end-to-end (wrong order ⇒ a running orphan keeps its slot after startup); `request_before_inference_converges` is the ordering contract the runtime now implements. The sweep cadence deliberately stays `.startup` — a periodic inference sweep would race the permit-drop's spawned terminal persistence and is not needed once the order is right.

`lake build` clean, zero `sorry`s.

## Rust

1. **Queue-waiter leak** (`admission/controller.rs`): `QueuedCallGuard` is constructed *before* the fallible `persist_call_queued` and arms terminal-persist-on-drop only once the row is durable. A persist error now releases the waiter unit instead of permanently shrinking the queue toward `QueueFull`.
2. **Crash-orphan recovery order**: new `startup_recovery` module runs the ordered startup sweeps (tool calls → requests → inference calls), consumed by startup's `log_recovery` (tool calls stay first so the #937 restart-disposition classifier still observes crash-time parent liveness). Crash-orphaned `running` rows now terminalize on the **first** restart.
3. **`is_drained()` race** (`admission/controller.rs`): `running` is replaced by an `in_flight` counter incremented at acquisition intent and released through an RAII guard on every non-admitted exit (handed off to `AdmissionPermit` on success). A permit in the acquire→resume window — including one assigned to a parked waiter whose task hasn't run — is never invisible to reconcile's drain check. SeqCst on the closed flag + counter closes the close/check vs. count/check-closed interleaving.

## Regression tests (each watched failing before the fix)

| Defect | Test | Pre-fix failure |
|---|---|---|
| 1 | `admission::tests::queued_persist_failure_releases_queue_capacity` | waiter counter left at 1; asserts the counter and that queue capacity still admits |
| 2 | `conformance::startup_recovery_order_terminalizes_crash_orphaned_calls` | orphaned call left `running` after driving the real ordered sweep over a seeded crash state |
| 3 | `admission::tests::assigned_permit_is_visible_to_drain_detection` | `is_drained()` reported true while a parked waiter (noop-waker-driven) held an assigned permit |

## Gates

- `cargo test -p gents` — all suites pass (1194 lib, 326 conformance, all e2e)
- `cargo check --workspace --all-targets` — clean, zero warnings
- `cargo fmt --all -- --check` — clean
- `lake build` — clean, zero `sorry`s

One unrelated e2e failure (`panicking_background_tool_terminalizes_and_notifies`) occurred in a run that overlapped a disk-full window on the build host; it passed 3/3 reruns and in the final full-suite run.

Closes #1001

### DIFF
--- a/crates/gents/proofs/Proofs/InferenceCall.lean
+++ b/crates/gents/proofs/Proofs/InferenceCall.lean
@@ -3,3 +3,4 @@ import Proofs.InferenceCall.Transition
 import Proofs.InferenceCall.Executable
 import Proofs.InferenceCall.Properties
 import Proofs.InferenceCall.SlotAccounting
+import Proofs.InferenceCall.ControllerBookkeeping
--- a/crates/gents/proofs/Proofs/InferenceCall/ControllerBookkeeping.lean
+++ b/crates/gents/proofs/Proofs/InferenceCall/ControllerBookkeeping.lean
@@ -39,10 +39,17 @@ The load-bearing facts are:
   every terminal outcome — including the queued-persist failure path —
   contributes zero to both counters (the S9 analog for in-memory bookkeeping).
 
+The model's `release` action drops the permit and in-flight contributions
+atomically. Rust refines that edge by returning the semaphore permit
+*before* decrementing `in_flight` on every release path — the release can
+synchronously install a replacement controller, so the reverse order would
+let a drained controller briefly hold an outstanding permit.
+
 No contract JSON is emitted for this module; the Rust fence is
 `crates/gents/src/admission/tests.rs`
 (`queued_persist_failure_releases_queue_capacity`,
-`assigned_permit_is_visible_to_drain_detection`), which drives the real
+`assigned_permit_is_visible_to_drain_detection`,
+`drained_signal_implies_permit_returned`), which drives the real
 controller through these paths and asserts the modeled contributions.
 -/
 
--- a/crates/gents/proofs/Proofs/InferenceCall/ControllerBookkeeping.lean
+++ b/crates/gents/proofs/Proofs/InferenceCall/ControllerBookkeeping.lean
@@ -1,0 +1,268 @@
+import Proofs.InferenceCall.SlotAccounting
+import Mathlib.Algebra.Order.BigOperators.Group.Finset
+
+open scoped BigOperators
+
+/-!
+# In-memory admission controller bookkeeping (#1001)
+
+`Proofs/InferenceCall/SlotAccounting.lean` reasons about *persisted* running
+rows. The runtime additionally keeps two in-memory counters on each
+`BackendAdmissionController` (`crates/gents/src/admission/controller.rs`) that
+the persisted model abstracts over:
+
+- the **queue-waiter** counter bounding `max_queue_depth`, and
+- the **in-flight** counter that drain detection (`is_drained`) reads before
+  `AdmissionRegistry::reconcile` installs a replacement controller.
+
+Issue #1001 found two defects that live exactly in that gap:
+
+1. a fallible durable write between the waiter increment and the guard that
+   decrements it leaked queue capacity on persist failure, and
+2. in-flight admissions were counted only *after* semaphore acquisition, so a
+   permit-holder in the acquire→count window was invisible to `is_drained`,
+   letting reconcile install a fresh full-capacity controller while an old
+   permit was live.
+
+This module models one call's path through `acquire` as a phase machine and
+assigns each phase its waiter / in-flight / semaphore-permit contribution.
+The load-bearing facts are:
+
+- `permit_implies_in_flight` / `drained_no_outstanding_permits`: with
+  in-flight counted from `enteredAcquire` (acquisition *intent*), a drained
+  controller holds no permits, so `replacement_preserves_capacity` keeps the
+  per-backend bound after a controller swap.
+- `late_admission_counting_unsound`: counting in-flight only at `admitted`
+  (the pre-fix Rust behavior) violates that invariant — the model rejects the
+  buggy counting rather than merely not mentioning it.
+- `persist_error_releases_waiter` and `terminal_phase_releases_bookkeeping`:
+  every terminal outcome — including the queued-persist failure path —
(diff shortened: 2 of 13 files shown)


jackzampolin:
Review follow-up (P1: permit released after drained signal) — confirmed and fixed in 8c320c92.

Verified: `release_in_flight()` in `AdmissionPermit::drop` ran before the `_permit` field dropped (fields drop after the Drop body, which locks the terminal-failure observer — so the window was unbounded under lock contention, not just microseconds). That state — permit outstanding, in-flight zero — is exactly what `ControllerBookkeeping.permit_implies_in_flight` forbids; the Rust wasn't refining the model's atomic `release` edge. The same reversed ordering existed on `start_permit`'s persist-error path via reverse parameter-drop order.

Fix: the permit is `Option`-stored and taken/dropped first in `Drop`; `start_permit`'s error arm and the queued path's error arm drop the permit explicitly before the in-flight guard releases. Regression test `drained_signal_implies_permit_returned` uses the suggested stall (holding the observer lock pins the drop mid-body deterministically) and asserts the drained signal implies all permits returned — watched failing before the fix (`available_permits: 0` at drained). Full gates re-run green: `cargo test -p gents`, `cargo check --workspace --all-targets`, fmt, `lake build` (docstring updated to name the release-edge refinement obligation and the new fence).

One nuance for the record: by Drop time the permit-holder's provider stream has already finished, so actual concurrent inference work never exceeded `max_concurrent` — the violation was in outstanding-permit accounting. But the model rightly forbids the state, and the observer-lock stall made the window unbounded, so the strict ordering is the correct contract.

PR #1013, opened 2026-08-03
Add managed local-agent onboarding to desktop

nasdf:
## Summary
- add an optional first-run flow to create, name, and start a standard local agent while preserving remote pairing as a skip path
- persist the local-agent opt-in, restore the managed server before desktop client bootstrap, coalesce concurrent restoration attempts, and handle compatible external servers or identity conflicts
- guide newly created local agents into inference setup when they still use the seeded placeholder backend
- host the server in-process through a reusable CLI server host and embed the filesystem runner without bundling a CLI binary or sidecar
- add a system tray while the local agent is active, with actions to open Gents, stop the agent, or quit, plus close-to-tray behavior
- safely stop and restart the desktop client around local provisioning to avoid database contention

## Testing
- Manually ran through onboarding with local agent server.
- Closed and restarted app to ensure server comes back up.
- Manually tested the system tray works on linux.


crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
```diff
@@ -0,0 +1,306 @@
...
+        BridgeError::new(
+            BridgeErrorCode::Unsupported,
+            "managed server requires a local agent home",
+        )
+    })?;
+    let stored = load_preference(&state).await?;
+
+    {
+        let managed = state.managed_server.lock().await;
+        if managed.server.is_some() {
```
jackzampolin, line 58: **TOCTOU race — start is not single-flight.** This checks `managed.server.is_some()` under the lock, but releases it for the whole provision+start await below. Two concurrent calls (restore-on-mount racing the onboarding click, or a second window) both pass this check, both set `starting`, and both call `start_server` — the loser fails on the port bind and stamps `last_error`, leaving status `"failed"` while a server is actually running. The frontend `WeakMap` only coalesces `restoreManagedServer`, not restore-vs-onboarding.

The bridge already has the `client_lifecycle` single-flight pattern — the start path could hold an equivalent guard, or bail early when `starting` is observed true.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
```diff
@@ -0,0 +1,306 @@
...
+use tauri::{AppHandle, Emitter, Runtime, State};
+
+use gents_desktop_core::local_runtime::{init_standard_local_runtime, DesktopInitOptions};
+
+use crate::config::ManagedServerPolicy;
+use crate::error::{BridgeError, BridgeErrorCode};
+use crate::state::DesktopAppState;
+use crate::types::{ManagedServerStartRequest, ManagedServerStatus};
+
+const MANAGED_SERVER_EVENT: &str = "desktop://managed-server-updated";
```
jackzampolin: This duplicates `contract::MANAGED_SERVER_UPDATED_EVENT` — could import that instead so the contract stays the single source of truth (the tray in `apps/gents-desktop/src-tauri/src/bridge/mod.rs` hardcodes the string a third time).

jackzampolin (commented):
Solid PR — the design (policy-gated, fail-closed default, loopback-only, opt-in with a remote-pairing skip path) is right, and the subtle bits (two-phase auto-start commit, iroh ephemeral-port refresh) are commented where they live. Review notes below; the themes:

**CI never actually ran.** Both failures are infrastructure: `lean-proofs` died in 11s at setup with `sccache: Server startup failed: Address in use` (fallout from the sccache-recycle step in a4d98b3c), and `rust-and-cli` failed only because it gates on lean-proofs. So the Rust suite has never compiled this branch — worth merging main in and re-running before anything else.

**Would like addressed before merge:**
1. Backend start is not single-flight — TOCTOU double-start race (inline comment).
2. Constant duplication that will drift silently: event names re-hardcoded in three places, port 9191/`init.json` re-hardcoded in the bridge, seeded endpoint/model duplicated in TS (inline comments).
3. `ManagedServerStatus.state` stringly-typed, with a hand-written TS union shadowing the generated type (inline comment).

**Author's discretion:** quit-hang timeout, seeded-placeholder heuristic vs the documented default backend, stop discarding the server handle on error, unit tests for `status_from` / identity-conflict, naming the two-call commit protocol in the adapter.

Also: nice defensive fixes along the way (empty `peers.json` tolerance, provisioning rollback preserving the original error, DID-conflict detection on port 9191). Skipping the Lean spec is legitimate here — host plumbing, no lifecycle/invariant changes.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    emit_status(&app, &state).await;
+    desktop_managed_server_status(state).await
+}
+
+async fn matching_external_server(
+    agent_home: &std::path::Path,
+) -> Result<Option<ManagedServerStatus>, BridgeError> {
+    let payload = match gents_desktop_core::local_runtime::fetch_runtime_connection_payload(
+        "http://127.0.0.1:9191/status",
```
jackzampolin: Hardcoded `http://127.0.0.1:9191/status` (and `init.json` just below) — these silently drift if `DEFAULT_HTTP_PORT` or `INIT_CONFIG_FILE_NAME` ever change. Since `gents_server` is now a dependency, could derive the URL from the same constant `ServerConfig::standard` uses, and use the exported `server_host::initialized_home()` helper (currently unused) for the init.json check.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
```diff
@@ -0,0 +1,306 @@
...
+) -> Result<ManagedServerStatus, BridgeError> {
+    ensure_allowed(&state)?;
+    let server = {
+        let mut managed = state.managed_server.lock().await;
+        managed.starting = false;
+        managed.last_error = None;
+        managed.server.take()
+    };
+    if let Some(server) = server {
+        server
```
jackzampolin, line 200: Minor: if `shutdown()` errors, the `RunningServer` handle was already `take()`n and is dropped — the server thread may still be live but is no longer tracked, so a retry of stop becomes a no-op. Consider putting the handle back (or at least noting the abandonment) on the error path.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
```diff
@@ -0,0 +1,306 @@
...
+    if state.policy.managed_server != ManagedServerPolicy::Allowed {
+        return Err(BridgeError::new(
+            BridgeErrorCode::Unsupported,
+            "managed local server hosting is disabled for this desktop host",
+        ));
+    }
+    Ok(())
+}
+
+fn status_from(
```
jackzampolin, line 225: `status_from` is a pure function encoding the running/starting/failed/stopped/disabled priority order, and `matching_external_server`'s DID-conflict branch is the one security-relevant check in this file — both are cheaply unit-testable and currently untested on the Rust side.

apps/gents-desktop/src-tauri/src/bridge/mod.rs
```diff
@@ -33,8 +43,102 @@ pub fn run() {
...
+            }
+            "quit" => shutdown_and_quit(app),
+            _ => {}
+        })
+        .build(app)?;
+    tray.set_visible(false)?;
+
+    let tray_id = tray.id().clone();
+    let app_handle = app.handle().clone();
+    app.listen("desktop://managed-server-updated", move |event| {
```
jackzampolin: Hardcoded event strings here (`desktop://managed-server-updated` above at the emit site, and `desktop://managed-server-tray-stop`) — the former exists as `contract::MANAGED_SERVER_UPDATED_EVENT`, and the tray-stop event isn't in the contract's event inventory at all. Worth adding it so `contracts/desktop-bridge.json` stays a complete inventory.

apps/gents-desktop/src-tauri/src/bridge/mod.rs
```diff
@@ -33,8 +43,102 @@ pub fn run() {
...
+#[cfg(desktop)]
+fn show_main_window<R: tauri::Runtime>(app: &tauri::AppHandle<R>) {
+    if let Some(window) = app.get_webview_window("main") {
+        let _ = window.show();
+        let _ = window.set_focus();
+    }
+}
+
+#[cfg(desktop)]
+fn shutdown_and_quit<R: tauri::Runtime>(app: &tauri::AppHandle<R>) {
```
jackzampolin, line 135: `shutdown()` joins the server thread with no timeout, so a wedged server makes "Quit Gents" do nothing. A `tokio::time::timeout` falling through to `app.exit(0)` would keep quit reliable.

crates/gents-desktop-bridge/src/types/views/bootstrap.rs
```diff
@@ -29,3 +29,14 @@ pub struct DesktopBootstrapSummary {
     pub peer_directory_exists: bool,
     pub saved_peers: Vec<SavedPeerView>,
 }
+
+#[derive(Debug, Clone, Serialize, TS)]
+#[serde(rename_all = "camelCase")]
+pub struct ManagedServerStatus {
+    pub state: String,
```
jackzampolin: `state` as `String` means ts-rs generates `state: string`, and a hand-written union type with the same name in `packages/gents-desktop-client/src/api/types.ts` shadows it. If this were a Rust enum (`#[serde(rename_all = "lowercase")]`), the generated type would carry the union and the hand-written duplicate could be deleted — exactly the drift the typegen exists to prevent.

packages/gents-desktop-fleet/src/components/FleetDashboard.tsx
```diff
@@ -16,6 +16,9 @@ import { AddPeerForm } from "./AddPeerForm.js";
 import { FleetRow } from "./FleetRow.js";
 import { NetworkPanel } from "./NetworkPanel.js";
 
+const SEEDED_LOCAL_ENDPOINT = "http://127.0.0.1:8080/v1";
```
jackzampolin: These mirror Rust's `DEFAULT_INIT_ENDPOINT` / `DEFAULT_INIT_MODEL_NAME` in `gents-cli`. If the init defaults ever change, this detection silently stops matching and the seeded-placeholder flow breaks with no error. At minimum a comment on both sides pointing at each other; better, surface the seeded values through the bridge so there's one source of truth.

packages/gents-desktop-fleet/src/components/FleetDashboard.tsx
```diff
@@ -26,7 +29,13 @@ function needsInferenceSetup(deployment: DeploymentView): boolean {
     ) ?? deployment.inferenceBackends[0];
   if (!backend) return true;
   if (backend.enabled === false) return true;
-  return backend.models.length === 0;
+  if (backend.models.length === 0) return true;
+  return (
+    deployment.source === "local-standard" &&
+    backend.endpoint === SEEDED_LOCAL_ENDPOINT &&
+    backend.models.length === 1 &&
+    backend.models[0] === SEEDED_LOCAL_MODEL
```
jackzampolin: This treats the seeded endpoint+model as "not set up" for `local-standard` agents — but that's precisely the configuration `gents init`'s help text tells users to run (`llama-server -hf google/gemma-4-12B...`). A user actually serving that model gets a permanent "still needs a working model backend" callout. A probe result would be a truer signal; at minimum this deserves a comment and a deliberate product call.

packages/gents-desktop-fleet/src/LocalRuntimeConnect.tsx
```diff
@@ -8,25 +8,35 @@ import { formatPeerConnectionError } from "./peerConnectionErrors.js";
...
     bootstrap?.initAgentDid?.trim() || bootstrap?.defaultAgentHome || "";
 
   async function connect() {
     setError(null);
     try {
+      await onStartServer?.(agentName);
       await onConnect(agentName);
+      // The second call commits launch restoration only after the client has
+      // successfully provisioned and connected the local peer.
+      await onStartServer?.(agentName);
```
jackzampolin: The start → connect → start commit protocol is load-bearing (the second call is what persists auto-start) but lives implicitly in a UI component's click handler — easy to break in a refactor that "dedupes" the call. A named adapter method (e.g. `commitManagedServerAutoStart()`) would make the contract explicit; the comment helps, but a name is harder to delete by accident.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
nasdf: Addressed in a7d98325. Added a dedicated managed_server_lifecycle mutex and hold it across the full provision/start/readiness sequence and stop, so concurrent restore/onboarding calls serialize. The waiter then observes the installed in-process handle and follows the existing preference-commit path instead of racing the port bind.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
nasdf: Addressed in a7d98325. managed_server.rs now imports contract::MANAGED_SERVER_UPDATED_EVENT; the duplicate local constant was removed. The Tauri tray listener also consumes the same contract constant.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
nasdf: Addressed in a7d98325. Added ServerConfig::status_url() and now build the probe URL from ServerConfig::standard, so it shares the standard address/port source. The init check now uses server_host::initialized_home(); the only remaining init.json read extracts the initialized DID after that shared check succeeds.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
nasdf: Reviewed, but intentionally left unchanged in this commit. RunningServer::shutdown consumes the handle because it sends shutdown and joins the owned thread; preserving a retryable handle would require changing that ownership API to retain an optional join handle and defining behavior after a failed join. I kept this separate from the required review fixes rather than adding a partial retry contract.

crates/gents-desktop-bridge/src/tauri_commands/managed_server.rs
nasdf: Addressed in a7d98325. Added focused Rust tests covering the starting, failed, stopped, and disabled status priority and the security-relevant rejection of an external server whose DID differs from the initialized home. The identity comparison was extracted into a pure helper so the conflict branch is directly testable.

apps/gents-desktop/src-tauri/src/bridge/mod.rs
nasdf: Addressed in a7d98325. Both tray events now come from bridge contract constants. Added MANAGED_SERVER_TRAY_STOP_EVENT to the event inventory, bumped the additive contract version to 0.8, and regenerated contracts/desktop-bridge.json.

apps/gents-desktop/src-tauri/src/bridge/mod.rs
nasdf: Addressed in a7d98325. Quit now bounds managed-server shutdown with a five-second tokio::time::timeout; errors and timeouts are logged, and both paths still proceed to app.exit(0).

crates/gents-desktop-bridge/src/types/views/bootstrap.rs
nasdf: Addressed in a7d98325. ManagedServerStatus.state is now the serde-lowercase Rust ManagedServerState enum. Regenerated ts-rs output produces the exact six-value union, and the handwritten shadow union was removed in favor of re-exporting the generated type.

packages/gents-desktop-fleet/src/components/FleetDashboard.tsx
nasdf: Addressed in a7d98325 by removing the duplicated seeded endpoint/model constants and the inference based on them. The frontend now considers a backend configured when it exists, is enabled, and has at least one model, without mirroring CLI defaults.

packages/gents-desktop-fleet/src/components/FleetDashboard.tsx
nasdf: Addressed in a7d98325. The seeded-placeholder heuristic was removed, so the documented default llama-server endpoint/model is treated as a valid configured backend rather than forcing a permanent setup callout. The UI test was inverted to protect this behavior.

packages/gents-desktop-fleet/src/LocalRuntimeConnect.tsx
nasdf: Addressed in a7d98325. Added the named adapter operation commitManagedServerAutoStart and threaded it through the host/fleet components as a distinct callback. The connect flow now reads start, connect, commit, making the load-bearing second call explicit in the API and test assertions.

Issue #330, opened 2026-05-20
Flaky test: `startup_recovery::run_agent_shutdown_is_prompt_while_request_waits_for_backend_capacity`


jackzampolin:
## Symptom

Intermittent failure on `cargo test -p defra-agent --lib`. Passes when run in isolation. First flagged during #327 (`Backgrounded tools panel: prototype + impl`) final verification — the implementer noted it was pre-existing and not on that branch's surface.

## Source

`crates/defra-agent/src/agent/runtime/tests/startup_recovery.rs:308`

```rust
async fn run_agent_shutdown_is_prompt_while_request_waits_for_backend_capacity() { ... }
```

## Why this matters

Per the project's no-tolerated-flakes discipline (the entire formal-verification investment exists to eliminate this class of failure): a "passes in isolation, fails in suite" intermittent is exactly the signal that there's a hidden shared-state or ordering assumption in the test fixture. Should not be shrugged at.

## Suspected shape

The test exercises `run_agent_shutdown` while a request is queued waiting for backend admission capacity. The intermittency likely points at:
- A `tokio::time::sleep` / `tokio::select!` race between the shutdown signal and the admission-waiter's poll loop
- A `tokio::test` runtime that's reused across tests, leaking task state

But these are guesses — needs a real repro, then a flamegraph or `RUST_LOG=trace` capture of a failing run.

## Repro to try

```bash
# In a loop until it fails:
for i in {1..50}; do
  cargo test -p defra-agent --lib startup_recovery::run_agent_shutdown_is_prompt_while_request_waits_for_backend_capacity -- --include-ignored --nocapture 2>&1 | tail -20 || break
done

# Or as part of the full lib suite (the failure mode reported in #327):
for i in {1..20}; do
  cargo test -p defra-agent --lib 2>&1 | tail -30 | grep -E '(FAILED|test result)'
done
```

## Out of scope here

This issue tracks the flake. The fix should live in its own PR — not bundled into an adjacent feature like #327.

jackzampolin:
Recurred 2026-08-04 during `cargo test -p gents` on branch `issue-1015-compaction-fenced-json` (base `4411456f`), on a machine saturated with two other concurrent worktree builds plus a cold Lean/Mathlib proofs build:

```
thread 'agent::runtime::tests::startup_recovery::run_agent_shutdown_is_prompt_while_request_waits_for_backend_capacity' panicked at crates/gents/src/agent/runtime/tests/support.rs:486:9:
[...]

PR #1020, opened 2026-08-04
fix(runtime): decode compaction summaries with unterminated fences

### LINKED ISSUE #1015: Compaction summary parser rejects complete JSON when a Markdown fence is not closed
## Problem

The compaction summary decoder only removes a Markdown fence when both the opening and closing fence are present. DeepSeek V4 Flash frequently returned a complete JSON object beginning with ```` ```json ```` but omitted the final ```` ``` ````. `parse_summary_response` then passed the opening fence to `serde_json` and aborted the entire agent request.

Observed in the full Terminal-Bench 2.1 run from PR #988 (`gents-d4f-max-988-full-c16-clean-v1`, Gents `4411456f`):

- `feal-linear-cryptanalysis`
- `largest-eigenval`
- `regex-chess`
- `torch-pipeline-parallelism`

All four responses contained a complete, useful summary object. The failure was `expected value at line 1 column 1` because the response still began with ```` ```json ````.

## Desired behavior

Decode the JSON object when the response is:

- bare JSON;
- a correctly closed `json` or untyped Markdown fence;
- an otherwise complete JSON object after an opening fence whose closing fence is absent;
- surrounded by bounded non-JSON whitespace or fence text.

Do not use a greedy first-`{`/last-`}` extraction that can silently select an unrelated object from arbitrary prose. The accepted envelope should remain narrow and deterministic.

Parse diagnostics must be bounded; a malformed model response must not be copied into multi-megabyte error strings, response documents, and logs.

## Acceptance criteria

- Unit tests cover bare, closed-fenced, unterminated-fenced, and malformed responses.
- The four captured response shapes parse successfully.
- Invalid output still fails with a useful bounded diagnostic.
- No durable transcript or provider-input lifecycle semantics change; this is decoder plumbing and does not require a Lean transition change.

## References

- `crates/gents/src/compaction/summary.rs`
- PR #988
- #717 for budget-safe fallback when a response remains unusable

jackzampolin:
## Summary

Closes #1015.

The compaction summary decoder only stripped a Markdown fence when both the opening and closing fence were present. DeepSeek V4 Flash frequently returns a complete JSON summary that opens with ```` ```json ```` but never emits the closing ```` ``` ````; `parse_summary_response` then handed the opening fence straight to serde_json and aborted the entire agent request (`expected value at line 1 column 1`). Observed on four tasks in the full Terminal-Bench 2.1 run from #988.

## Changes

- **Closing fence is now optional.** `strip_markdown_fence` strips a leading ` ```json ` or untyped ` ``` ` fence and, if present, a trailing ` ``` `.
- **The envelope stays narrow.** The payload must be exactly one JSON object followed by nothing but whitespace (or the closing fence), enforced with `serde_json::Deserializer::end()` — no greedy first-`{`/last-`}` extraction that could pick an unrelated object out of prose. A regression test pins this down.
- **Bounded diagnostics.** Parse errors now embed a 256-byte, char-boundary-safe excerpt plus the total byte count instead of the full response, so a malformed multi-megabyte reply is no longer copied into error strings, response documents, and logs.

No durable-transcript or provider-input lifecycle semantics change — this is decoder plumbing at the summary boundary, so no Lean transition change is required (per the issue).

## Testing

- 8 new unit tests: bare JSON, closed `json`/untyped fences, unterminated `json`/untyped fences (the #1015 shape), JSON embedded in prose (rejected), trailing prose after the object (rejected), and a 4 MiB malformed response whose rendered diagnostic must stay under 2 KiB. Written first and observed failing against the old parser.
- `cargo test -p gents`: full package suite green (one unrelated pre-existing flake reproduced once and passed on re-run — reopened #330 with the capture).
- `cargo check --workspace --all-targets`: clean.

### DIFF
--- a/crates/gents/src/compaction/summary.rs
+++ b/crates/gents/src/compaction/summary.rs
@@ -20,24 +20,52 @@ pub(super) fn compaction_request_prompt() -> &'static str {
 }
 
 pub(super) fn parse_summary_response(raw_summary: &str) -> Result<SummaryResponse> {
-    let trimmed = raw_summary.trim();
-    let json = trimmed
-        .strip_prefix("```json")
-        .and_then(|value| value.strip_suffix("```"))
-        .map(str::trim)
-        .or_else(|| {
-            trimmed
-                .strip_prefix("```")
-                .and_then(|value| value.strip_suffix("```"))
-                .map(str::trim)
-        })
-        .unwrap_or(trimmed);
+    let json = strip_markdown_fence(raw_summary);
 
-    let mut summary: SummaryResponse = serde_json::from_str(json)
-        .with_context(|| format!("parsing compaction summary response: {json}"))?;
+    let mut deserializer = serde_json::Deserializer::from_str(json);
+    let mut summary = SummaryResponse::deserialize(&mut deserializer)
+        // `end()` rejects anything but whitespace after the object, keeping the
+        // accepted envelope narrow: no extracting an object out of prose.
+        .and_then(|value| deserializer.end().map(|()| value))
+        .with_context(|| {
+            format!(
+                "parsing compaction summary response: {}",
+                bounded_excerpt(json)
+            )
+        })?;
     dedupe_paths(&mut summary.files_read);
     dedupe_paths(&mut summary.files_modified);
     Ok(summary)
+}
+
+/// Strips a `json` or untyped Markdown fence around the summary object. The
+/// closing fence is optional: models sometimes emit a complete object after an
+/// opening fence and stop without closing it (#1015).
+fn strip_markdown_fence(raw: &str) -> &str {
+    let trimmed = raw.trim();
+    let Some(body) = trimmed
+        .strip_prefix("```json")
+        .or_else(|| trimmed.strip_prefix("```"))
+    else {
+        return trimmed;
+    };
+    let body = body.trim();
+    body.strip_suffix("```").map(str::trim_end).unwrap_or(body)
+}
+
+/// Malformed model responses can be multi-megabyte; diagnostics carrying them
+/// verbatim would be copied into error strings, response documents, and logs.
+const DIAGNOSTIC_EXCERPT_BYTES: usize = 256;
+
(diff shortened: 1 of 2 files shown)


jackzampolin (commented):
Review notes

The unique parser change looks sound: it accepts only the two intended fence forms, still requires the JSON value to consume the entire remaining input, keeps prose rejection intact, and bounds malformed-output diagnostics. I did not find a blocker in that delta.

Before merge, please fix the PR ancestry. This PR targets main but carries the entire still-open #988 stack, so GitHub currently presents 78 changed files rather than the compaction-parser change. #1022 already uses agent/add-atif-trace-projection as its base. Please either retarget this PR to that branch now, or wait for #988 to merge and rebase onto the resulting main, then rerun CI.

Please also establish the integration order with #1021: both branches edit crates/gents/src/compaction/tests.rs, so whichever lands second should rebase and run the combined compaction suite rather than resolving that overlap mechanically.

jackzampolin:
Retargeted the base to `agent/add-atif-trace-projection`, matching #1022 — the PR now shows only the two parser files (the branch is exactly one commit ahead of #988's head, which is a direct ancestor, so the diff is clean with no rebase needed).

Integration order with #1021: since this PR is now stacked behind #988 while #1021 targets `main` directly, #1021 will presumably land first. When #988 merges and this PR retargets onto the resulting `main`, I'll rebase over #1021's `compaction/tests.rs` changes by hand and rerun the full compaction suite (`cargo test -p gents compaction`) plus the package gate before merge, rather than taking a mechanical conflict resolution.

jackzampolin (commented):
Follow-up review

Re-reviewed at c5d1a687 after the base retarget. The PR is now a clean two-file parser diff against #988, GitHub reports it mergeable, and rename-guard, Lean, desktop, and rust-and-cli are all green. The head is the same implementation previously reviewed, and the stated integration plan with #1021 addresses the remaining overlap concern.

No remaining review blocker; #1020 is ready from the code-review side.

PR #1021, opened 2026-08-04
fix(runtime): retry empty compaction completions on a bounded immediate ladder

### LINKED ISSUE #1016: Retry empty internal compaction completions with a bounded immediate policy
## Problem

Internal compaction uses `CompletionRetryPolicy::no_retry()`. A provider turn that emits reasoning but no visible summary therefore aborts the user request immediately, even though the owned completion loop can safely retract and resample because no tool effect occurred.

Observed eight times in the full Terminal-Bench 2.1 run from PR #988 (`gents-d4f-max-988-full-c16-clean-v1`, Gents `4411456f`):

- `extract-moves-from-video`
- `gpt2-codegolf`
- `llm-inference-batching-scheduler`
- `make-mips-interpreter`
- `path-tracing`
- `path-tracing-reverse`
- `qemu-alpine-ssh`
- `schemelike-metacircular-eval`

Every request failed with `compaction summary inference failed ... completion produced no visible output: transport retry budget exhausted after 0 attempt(s)`.

The current no-retry rule was added to prevent an internal sub-completion from inheriting the scheduled origin's 5s/30s/120s ladder and blocking inline compaction for minutes. That does not require zero recovery attempts.

## Desired behavior

Give internal tool-free sub-completions a small, fixed, immediate, deadline-aware recovery budget. It must not inherit the parent execution origin's scheduled ladder. Because compaction has no tools and no effects, the retry is the already-modelled retract-and-resample transition.

## Acceptance criteria

- An empty/no-visible-output first compaction attempt is retracted and immediately resampled at least once.
- The retry count is small and fixed, with no multi-minute scheduled backoff inheritance.
- Repeated empty output fails deterministically after the internal budget.
- Ordinary transport failures follow the explicitly documented internal policy.
- Tests prove the number of provider calls and that no tool effect can be replayed.
- Existing CompletionRetry Lean/conformance invariants remain green; update the policy fence that currently asserts exactly one compaction call.

## References

- `crates/gents/src/compaction.rs`
- `crates/gents/src/agent/completion_retry.rs`
- `crates/gents/src/agent/loop_stream.rs`
- PR #648
- #717 for a budget-safe fallback after retries are exhausted

jackzampolin:
Closes #1016.

## Problem

Internal compaction sub-completions forced `CompletionRetryPolicy::no_retry()`, so a provider turn that emitted reasoning but no visible summary aborted the whole user request on the first attempt (`compaction summary inference failed ... completion produced no visible output: transport retry budget exhausted after 0 attempt(s)`). Observed eight times in the full Terminal-Bench 2.1 run from #988. The no-retry rule existed to keep compaction off the scheduled origin's minutes-scale 5s/30s/120s ladder (#648) — it never required zero recovery.

## Change

- **`CompletionRetryPolicy::internal_immediate()`** — a small, fixed, immediate, deadline-aware recovery budget for internal tool-free sub-completions with no caller-level retry: transport ladder `[0s, 1s, 2s]` (the crate-wide three-retry convention; the first slot lands at the 100ms jitter floor), `max_resample: 0`, `allow_repair: false`. Deadline fail-fast applies to every slot, including the immediate one.
- **`DefraCompactor::new` forces `internal_immediate()`** instead of `no_retry()`. Because compaction is tool-free, the empty-output recovery is exactly the already-modelled `CompletionRetry.retract` transition — retract the streamed reasoning, resample the identical request; no tool effect exists to replay.
- **Title generation keeps `no_retry()`** — `generate_title_with_fallback` already retries at its own layer (#648's original concern), and the doc comments now spell out that split.
- **No Lean changes** — the Lean `Budget` is parametric and retract/resample plus its invariants are already proven; only budget values changed, so the CompletionRetry conformance surface is untouched.

## Acceptance criteria → tests

- *Empty first attempt is retracted and immediately resampled*: `empty_compaction_completion_is_retracted_and_immediately_resampled` — succeeds in exactly 2 provider calls; asserts the resample re-issues byte-identical chat history and that no request offers tools (nothing to replay).
- *Small fixed budget, no scheduled-ladder inheritance*: `internal_immediate_carries_a_small_fixed_ladder_with_no_resample_or_repair`, `internal_immediate_retracts_three_times_then_fails` (delay-range fenced per slot), and the reworked #648 fence `transient_compaction_failures_follow_the_internal_immediate_policy`, which hands the compactor a `scheduled_default` config and proves the override (virtual-time bounded).
- *Deterministic failure after the budget*: `repeated_empty_compaction_completions_fail_after_the_internal_budget` — exactly 1 + 3 provider calls, then the empty-output error surfaces; the scripted mock panics on any call past the script, so over-retrying can't pass silently.
- *Deadline awareness*: `internal_immediate_is_deadline_aware`.

All tests were written first and watched fail against `no_retry()` (the empty-output test reproduces the Terminal-Bench error string verbatim) before the policy change.

## Verification

- `cargo test -p gents` — full package suite green (lib 1236 passed + all integration binaries, 0 failures).
- `cargo check --workspace --all-targets` — clean.

### DIFF
--- a/crates/gents/src/agent/completion_retry.rs
+++ b/crates/gents/src/agent/completion_retry.rs
@@ -159,13 +159,34 @@ impl CompletionRetryPolicy {
     }
 
     /// No retry at all: an empty transport ladder, no resample, no repair. For
-    /// internal sub-completions (compaction, title generation) that are not a
-    /// user execution origin — they fail fast and are re-driven, if at all, by
-    /// their own caller, and must not inherit the scheduled ladder's
-    /// minutes-scale, deadline-less backoff (#648).
+    /// internal sub-completions that already retry at their own layer (title
+    /// generation's `generate_title_with_fallback`) — the inner completion
+    /// must not also inherit the scheduled ladder's minutes-scale,
+    /// deadline-less backoff (#648).
     pub fn no_retry() -> Self {
         Self {
             transport_backoff: Vec::new(),
+            max_resample: 0,
+            allow_repair: false,
+        }
+    }
+
+    /// Small, fixed, immediate, deadline-aware recovery budget for internal
+    /// tool-free sub-completions with no caller-level retry (compaction). Like
+    /// [`Self::no_retry`] it never inherits an execution origin's ladder
+    /// (#648), but it keeps the already-modelled retract-and-resample
+    /// recovery available: an empty-output turn or transient transport
+    /// failure gets three attempts on the crate-conventional three-slot
+    /// ladder — the first immediate (jitter floors it at 100ms) — before
+    /// failing deterministically (#1016). No resample or repair: those are
+    /// parse-400 remedies for tool-calling requests.
+    pub fn internal_immediate() -> Self {
+        Self {
+            transport_backoff: vec![
+                Duration::ZERO,
+                Duration::from_secs(1),
+                Duration::from_secs(2),
+            ],
             max_resample: 0,
             allow_repair: false,
         }
--- a/crates/gents/src/agent/completion_retry/tests.rs
+++ b/crates/gents/src/agent/completion_retry/tests.rs
@@ -365,6 +365,59 @@ fn interactive_default_allows_a_single_short_retry_then_fails() {
 }
 
 #[test]
+fn internal_immediate_carries_a_small_fixed_ladder_with_no_resample_or_repair() {
+    // #1016: internal tool-free sub-completions (compaction) recover on a
+    // fixed 0s/1s/2s ladder — the crate-wide three-retry convention with an
+    // immediate first slot — and never on the origin ladders.
+    let policy = CompletionRetryPolicy::internal_immediate();
+    assert_eq!(
+        policy.transport_backoff,
+        vec![
+            Duration::ZERO,
+            Duration::from_secs(1),
+            Duration::from_secs(2)
+        ]
+    );
+    assert_eq!(policy.max_resample, 0);
+    assert!(!policy.allow_repair);
+}
+
+#[test]
+fn internal_immediate_retracts_three_times_then_fails() {
+    // The empty-output path is the no-effect mid-stream failure: each retry
+    // must be a retract-and-resample (never close-and-continue), the first at
+    // the 100ms jitter floor, and the fourth failure is terminal.
+    let mut state = CompletionRetryState::new(CompletionRetryPolicy::internal_immediate());
+    let now = Utc::now();
+
+    let expected_ranges = [(100, 100), (750, 1_250), (1_500, 2_500)];
+    for (low_ms, high_ms) in expected_ranges {
+        match state.on_mid_stream_failure(false, now, None) {
+            MidStreamDirective::RetractAndResample { delay } => {
+                assert_in_range(delay, low_ms, high_ms);
+            }
+            other => panic!("expected RetractAndResample, got {other:?}"),
+        }
+    }
+    match state.on_mid_stream_failure(false, now, None) {
+        MidStreamDirective::Fail { reason } => assert!(reason.contains("exhausted")),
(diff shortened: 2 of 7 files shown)


jackzampolin (commented):
Review notes

There is a blocking production-path deadline bug in the new retry policy. BehaviorDaemon::new constructs DefraCompactor from completion_factory::loop_config, whose deadline is None, and DefraCompactor stores that config for the daemon lifetime. run_inference assigns request_deadline only to the outer loop_config; its turn-compaction callback clones self.compactor, and the pre-inference compaction path in daemon/request.rs also calls self.compactor directly. Consequently the summary run_loop_to_text receives deadline: None in both real compaction paths.

The internal_immediate_is_deadline_aware unit test proves CompletionRetryState behaves correctly when a deadline is supplied, but production never supplies it. A request at its deadline can therefore still consume the 0/1/2-second ladder. Please pass the active claimed deadline into each compaction call (or create a request-scoped compactor/config) and cover both daemon compaction entry points with a regression that verifies an already-expired deadline does not retry the provider.

This PR also needs the same ancestry correction as #1020: it targets main while carrying all of open PR #988, producing an 80-file diff. Please retarget it to agent/add-atif-trace-projection, or rebase onto main after #988 lands, before treating its CI result as the merge gate.

jackzampolin:
Both review items addressed in b15cfcd0 plus a base retarget:

**Deadline bug** — confirmed as described: `DefraCompactor` held a daemon-lifetime config with `deadline: None`, and `run_inference` assigned the claimed deadline only to the outer loop config, so both compaction paths recovered with no deadline. Fix:

- `CompactionOptions.deadline` carries the request-scoped claimed deadline; `DefraCompactor::compact` min-combines it with any config deadline before running the summary loop.
- Both entry points (`daemon/request.rs` pre-inference, `daemon/inference.rs` turn callback) now build options via `BehaviorDaemon::compaction_options_for_request(deadline)`, which takes the deadline as a **required** argument — a call site that omits it no longer compiles. This also absorbed the redundant inline `strategy`/`force_summarize` overrides at both sites.
- Regression `expired_deadline_stops_compaction_recovery_at_one_provider_call`: an already-expired `options.deadline` with an empty first turn stops at exactly one provider call and fails naming the deadline, not budget exhaustion.

On test shape: there is no existing harness that drives `handle_request`/`run_inference` against a scripted provider, so per-entry-point behavioral e2e coverage would mean building one. I covered the shared compactor boundary behaviorally and made the per-site wiring compile-enforced instead. If you want the daemon-level harness anyway, I'd suggest that as a follow-up issue.

**Ancestry** — retargeted to `agent/add-atif-trace-projection`; the diff is now the 4 (now 6) files of this change. If #988 lands first, this can be retargeted back to `main` and rebased.

Gates on the new head: `cargo test -p gents` fully green (1237 lib + all integration binaries), `cargo check --workspace --all-targets` clean.

jackzampolin (commented):
Follow-up review

Re-reviewed b15cfcd0. The deadline finding is resolved: both production daemon entry points now build request-scoped CompactionOptions through a required-deadline helper, DefraCompactor applies the earlier of the request and stored-config deadlines, and the expired-deadline regression proves the shared recovery path stops at one provider call. I am comfortable with compile-enforced entry-point wiring plus the shared behavioral test here; a new daemon/provider harness is not required for this PR.

The ancestry is also correctly scoped. One integration gate remains because #1022 and #1020 were merged into agent/add-atif-trace-projection after this head's CI run began. I verified the current base and b15cfcd0 merge cleanly with git merge-tree, including the shared compaction/tests.rs edits, but the recorded checks exercised the old #988 base (4411456f), not the current base (0cd9252d). Please update/rebase this branch onto the current base and rerun cargo test -p gents plus cargo check --workspace --all-targets. Once that combined-base CI is green, I have no remaining review blocker.

PR #1022, opened 2026-08-04
Harbor: verifier-score MaxTurn workspaces instead of recording agent exceptions

### LINKED ISSUE #1019: Harbor should verifier-score MaxTurn workspaces instead of recording agent exceptions
## Problem

The Gents Harbor runner turns every terminal Gents error response into a non-zero agent command. Harbor then records a `RuntimeError` and skips verification, even when the terminal condition is the configured agent turn ceiling and the task filesystem may contain useful or complete work.

In PR #988's full Terminal-Bench 2.1 run, `gcode-to-text` and `make-doom-for-mips` reached `MaxTurnError: reached max turn limit: 250`. Both were counted as infrastructure exceptions rather than verifier-scored attempts.

## Desired behavior

Distinguish expected agent-budget exhaustion from infrastructure/runtime corruption. For MaxTurn, preserve the terminal response and ATIF trajectory, return control to Harbor, and run the verifier against the resulting workspace. Provider, persistence, compaction, and projection failures should remain explicit agent exceptions.

## Acceptance criteria

- `MaxTurnError` yields a completed Harbor agent phase with metadata identifying budget exhaustion.
- Harbor runs the verifier and records the actual workspace reward.
- Response and ATIF artifacts retain the terminal error and configured turn limit.
- Genuine infrastructure failures still make the agent phase fail.
- Tests cover complete, MaxTurn, and provider/compaction error responses.

## References

- `scripts/harbor/run_gents.sh`
- `scripts/harbor/gents_agent.py`
- PR #988

jackzampolin:
Closes #1019. Stacked on #988 (`agent/add-atif-trace-projection`) — this branch sits directly on its head, so only the three commits here are new.

## Problem

The Harbor runner turned every terminal Gents error response into a non-zero agent command, so Harbor recorded a `RuntimeError` and skipped verification — even when the terminal condition was the configured turn ceiling and the task filesystem held real work. In #988's Terminal-Bench 2.1 run, `gcode-to-text` and `make-doom-for-mips` hit `MaxTurnError` at 250 turns and were counted as infrastructure exceptions instead of verifier-scored attempts.

## Change

**`scripts/harbor/run_gents.sh`** now classifies the terminal response instead of failing on any error status:

- `completed` and `max_turns_exhausted` exit 0, so Harbor runs the verifier and records the actual workspace reward. Budget exhaustion is detected by the `MaxTurnError:` token in the persisted `error_message` — produced exclusively by the owned loop's max-turn guard (`agent stream failed: PromptError: MaxTurnError: (reached max turn limit: N)`); the match is scoped to `"error_message":` lines so model content and the nested `request.failure_reason` can never classify on their own.
- Provider, persistence, compaction, projection, and unexpected-status failures still exit non-zero and remain explicit agent exceptions.
- Every classified case writes `/logs/agent/gents-outcome.json` (`outcome`, `response_status`, `max_turns`, `request_id`); `response.json` and `trajectory.json` are written before classification, so the terminal error and configured turn limit are retained in the artifacts.
- New fixture-driven `self-test` mode covering complete/completed, MaxTurn, provider, compaction, false-positive, full-envelope, and unexpected/missing-status responses.

**`scripts/harbor/gents_agent.py`** stamps `outcome`, `budget_exhausted`, and `terminal_error` into Harbor trial metadata in `populate_context_post_run`, with degraded-to-warning handling for missing or corrupt artifacts.

**Hardening:** the runtime max-turns test (`crates/gents/src/agent/loop_stream/tests.rs`) now asserts the rendered error carries the `MaxTurnError:` token, so a rig wording change breaks CI instead of silently reverting MaxTurn trials to infrastructure exceptions. `GENTS_MAX_TURNS` is validated as a JSON-safe integer.

**CI:** the `rename-guard` job gains `bash -n`, `shellcheck -s sh`, and the runner self-test.

## Verification

- `scripts/harbor/run_gents.sh self-test` — 10/10 classifications pass; `shellcheck -s sh` clean
- `cargo test -p gents --lib exceeding_max_turns_terminates_with_error` — passes with the new token assertion
- `cargo check --workspace --all-targets` — clean
- `populate_context_post_run` exercised against fixture logs (MaxTurn, completed, missing-outcome cases)

Design spec: `docs/superpowers/specs/2026-08-04-harbor-maxturn-verifier-design.md`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,12 @@ jobs:
       - name: Reject stale product identifiers
         run: scripts/rename-to-gents.sh guard all
 
+      - name: Check Harbor runner syntax
+        run: bash -n scripts/harbor/run_gents.sh
+
+      - name: Run Harbor runner self-test
+        run: scripts/harbor/run_gents.sh self-test
+
   rust-and-cli:
     needs: [rename-guard, lean-proofs]
     if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,9 @@ jobs:
       - name: Run Harbor runner self-test
         run: scripts/harbor/run_gents.sh self-test
 
+      - name: Run Harbor agent adapter tests
+        run: python3 scripts/harbor/test_gents_agent.py
+
   rust-and-cli:
     needs: [rename-guard, lean-proofs]
     if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,9 @@ jobs:
         run: scripts/rename-to-gents.sh guard all
 
       - name: Check Harbor runner syntax
-        run: bash -n scripts/harbor/run_gents.sh
+        run: |
+          bash -n scripts/harbor/run_gents.sh
+          shellcheck -s sh scripts/harbor/run_gents.sh
 
       - name: Run Harbor runner self-test
         run: scripts/harbor/run_gents.sh self-test
--- a/crates/gents/src/agent/loop_stream/tests.rs
+++ b/crates/gents/src/agent/loop_stream/tests.rs
@@ -1476,6 +1476,15 @@ async fn exceeding_max_turns_terminates_with_error() {
         !crate::error::classify_completion_error(error).is_retryable(),
         "max-turns exhaustion must be non-retryable; got {last:?}"
     );
+    // The Harbor adapter (scripts/harbor/run_gents.sh) classifies budget
+    // exhaustion by this token in the persisted error message
+    // (`agent stream failed: {error}` in agent/daemon/inference.rs). If rig's
+    // display wording changes, MaxTurn trials silently revert to Harbor
+    // infrastructure exceptions instead of verifier-scored attempts.
+    assert!(
+        error.to_string().contains("MaxTurnError:"),
+        "max-turns error display must carry the MaxTurnError: token; got {error}"
+    );
 }
 
 #[tokio::test]
--- a/crates/gents/src/agent/loop_stream/tests.rs
+++ b/crates/gents/src/agent/loop_stream/tests.rs
@@ -1482,9 +1482,7 @@ async fn exceeding_max_turns_terminates_with_error() {
     // display. If rig's wording changes, MaxTurn trials silently revert to
     // Harbor infrastructure exceptions instead of verifier-scored attempts.
     assert!(
-        error
-            .to_string()
-            .starts_with("PromptError: MaxTurnError: "),
+        error.to_string().starts_with("PromptError: MaxTurnError: "),
         "max-turns error display must start with the anchored Harbor prefix; got {error}"
     );
 }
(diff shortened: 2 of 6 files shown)


jackzampolin (commented):
Review notes

The shell self-test passes locally and the Python adapter compiles, but the classifier currently overmatches externally supplied provider errors. For every error response it accepts any error_message line containing MaxTurnError:. A CompletionError/ProviderError may include upstream response text containing that token; that infrastructure failure would then be returned as success and verifier-scored, contrary to the safe-default design.

Please narrow the match to the owned-loop error shape (for example the agent stream failed: PromptError: MaxTurnError: prefix/type path that the Rust test now pins), and add a fixture such as agent stream failed: CompletionError: ProviderError: upstream mentioned MaxTurnError: that must remain agent_error.

Non-blocking coverage note: the acceptance behavior also depends on populate_context_post_run merging outcome, budget_exhausted, and terminal_error, but current automated coverage exercises only the shell classifier and the Rust display token. A small temp-directory Python test for that metadata projection would protect the Harbor-facing contract. The PR base is correctly scoped to #988.

jackzampolin:
Both review items addressed in 3af1ba93:

**Overmatch (blocking):** Confirmed real — added the adversarial fixture (`agent stream failed: CompletionError: ProviderError: upstream mentioned MaxTurnError: (reached max turn limit: 250)`) first and watched it misclassify as `max_turns_exhausted`. The classifier now anchors on the full quoted key-plus-prefix `"error_message": "agent stream failed: PromptError: MaxTurnError: ` via `grep -F`; since JSON escapes quotes inside string values, that byte sequence can only introduce the real field, so provider errors echoing the token stay `agent_error` (fixture now passes, 11/11). The runtime max-turns test was strengthened from `contains("MaxTurnError:")` to `starts_with("PromptError: MaxTurnError: ")` so the exact rendered prefix the anchor depends on is pinned against rig changes.

**Metadata coverage (non-blocking):** Added `scripts/harbor/test_gents_agent.py` — stdlib-only unittest that stubs harbor/certifi and exercises `populate_context_post_run` over max-turn, completed, missing-artifact, and corrupt-outcome cases (4/4 pass). Wired into the `rename-guard` CI job.

Verified: self-test 11/11 + shellcheck clean, adapter unittest 4/4, `cargo test -p gents --lib exceeding_max_turns_terminates_with_error` passes with the strengthened assertion.

jackzampolin (commented):
Follow-up review

Re-reviewed 3af1ba93. The anchored owned-loop prefix fixes the provider-error overmatch, the adversarial fixture passes, and the new adapter tests cover the requested metadata contract. I locally verified the 11-case runner self-test, ShellCheck, all four Python adapter tests, and exceeding_max_turns_terminates_with_error. The substantive review findings are addressed.

One mechanical merge blocker remains: rust-and-cli is red because cargo fmt --all --check wants the starts_with assertion collapsed onto one line in crates/gents/src/agent/loop_stream/tests.rs. I reproduced that exact formatting diff locally. Please run cargo fmt --all, push the result, and let the required check rerun; after that I have no remaining review concern.

Issue #985, opened 2026-07-31
bash tool: omitted timeout_secs applies 10s (schema advertises ceiling as default); background await_mode still capped by command ceiling


jackzampolin:
## Summary

Two coupled defects make `await_mode: background` bash calls unusable for the long-running jobs they exist for, observed on gents v0.10.1 in a live GLM-5.2 session.

### 1. Advertised default ≠ applied default

`toolset/bash_tools.rs` advertises the schema:

```
"timeout_secs": { "default": self.default_timeout.as_secs(), "maximum": self.default_timeout.as_secs(), ... }
```

…so the model is told the default is the ceiling (120s on this deployment). But when the model omits `timeout_secs`, serde applies `default_command_timeout_secs()` → `DEFAULT_COMMAND_TIMEOUT_SECS = 10` (`toolset.rs:63`). Omission = 10s, not the advertised 120s.

### 2. Background calls share the foreground lifetime ceiling

`run_command` clamps with `Duration::from_secs(args.timeout_secs.max(1)).min(self.default_timeout)` (`bash_tools.rs:133/:213`), and `--command-timeout-secs` flows in as that ceiling (`serve.rs:306`). `await_mode: background` changes await semantics only — the spawned command is still killed at ≤ ceiling. There is no way to run a legitimately long (hours) background job through the tool surface; the model is forced into `setsid nohup` detach workarounds that escape lifecycle tracking (no cancel, no completion notification, orphaned process on failure).

## Evidence (production session, 2026-07-31)

Agent amy (studio-1), behavior `default`, `backgroundable_tool_names: ["bash_unrestricted"]`, unit runs `--command-timeout-secs 120`. AgentToolCall rows:

- seq 56: `await_mode=background`, no `timeout_secs` arg → started 21:31:26.06, completed 21:31:36.07, result `{"ok":false,"status":"timeout"}` — killed at exactly 10s.
- seq 85: same shape → started 21:34:26.24, completed 21:34:36.25 — again exactly 10s.
- The model then spent ~25 messages hunting for a phantom config knob (the schema told it the default was 120), and finally shipped a `setsid nohup` detached launcher to get an hours-long batch running outside the tool surface.

## Suggested fixes

1. Make the serde omission default equal the advertised schema default (the ceiling), so schema and behavior agree.
[...]

Issue #899, opened 2026-07-28
Rationalize session timeout hierarchy behind one user-facing budget


jackzampolin:
## Problem

gents exposes and enforces several timeout/deadline layers that are individually
reasonable but difficult to reason about as a system. A high-level session can
be configured for an hour while a lower-level operation still defaults to ten
seconds, silently turning a slow but valid operation into partial evidence or
repeated retries.

This surfaced in production session-classification work:

- classifier requests had a 90-minute validity window;
- the inference profile had a 3,600-second request deadline and 3,600-second
  stream-liveness timeout;
- the submitting CLI waited up to 3,600 seconds;
- foreground Bash was independently hard-capped at 10 seconds;
- large evidence reads for sessions with 586 messages / 430 actions and 930
  messages / 726 actions crossed that cap;
- the model responded by retrying narrower windows, increasing turn count and
  risking incomplete evidence despite ample session-level budget.

Increasing one upper-level timeout did not help because a stricter lower-level
timeout won first.

## Timeout inventory to map

At minimum, document the semantics, owner, default, precedence, and terminal
effect of:

- request `valid_until` / admission expiry;
- request execution deadline (`deadline_duration_secs`);
- provider request/connect/read timeouts, if distinct;
- provider stream-liveness inactivity timeout;
- per-turn / max-turn limits;
- native Bash foreground command cap;
- approved CLI-tool cap;
- native tool-call lifecycle deadline and recovery timeout;
- MCP connect, list-tools, and tool-call timeouts;
- background-process spawn, wait, and cancellation deadlines;
- local `gents request submit --timeout-secs` / chat / Codex-shim wait timeouts;
- subagent claim/spawn/child deadlines, including cross-deployment timeouts;
- persistence/materialization polling and retry windows;
- transport retry/resample backoff budgets.

For each layer, distinguish wall-clock budget, inactivity timeout, per-attempt
cap, and operator/client wait timeout. Those are not interchangeable.

## Design goal

Provide one user-facing session/execution budget that gives ordinary users
coherent behavior, while retaining expert overrides for individual layers.

A plausible precedence model is:

1. explicit per-request overrides;
2. inference-profile / behavior budget;
3. runtime/operator default;
4. derived lower-level values.

Derived deadlines should obey monotonic invariants:

- a child operation cannot outlive its parent request;
- a lower-level default must not unexpectedly make a larger declared session
  budget ineffective;
- client wait windows include the execution budget plus bounded materialization
  grace;
[...]

PR #1026, opened 2026-08-04
Harbor: bound foreground commands with a watchdog below the whole-request deadline

### LINKED ISSUE #1018: Harbor foreground commands need a watchdog below the whole-request deadline
## Problem

The PR #988 Harbor profile sets both request and foreground-command timeouts to 86,400 seconds to avoid artificial benchmark cutoffs. In the full Terminal-Bench 2.1 run, five trials remained active for 19-22 hours with no inference work running because a foreground tool or service workflow had not returned.

Examples observed live:

- `write-compressor`: `./enc` consumed CPU for more than 18 hours;
- `db-wal-recovery`: recursive binary `grep` over `/` ran for many hours;
- `qemu-startup`: an exhaustive syscall/fork scan remained active;
- `install-windows-3.11`: a QEMU probe remained active;
- `mailman`: the request remained blocked while spawned service processes persisted.

vLLM reported zero running and zero waiting requests while these trials occupied Harbor slots.

## Desired behavior

The benchmark adapter needs a bounded foreground-command policy that allows long legitimate builds but returns control to the model when a command becomes pathological. The timeout must be visible to the model and attributable in the trace, not converted into an opaque Harbor hang.

## Acceptance criteria

- Harbor's default foreground-command ceiling is materially lower than the whole-request deadline and documented separately.
- A timed-out command produces a normal tool outcome so the model can recover or choose a narrower command.
- Request cancellation/Harbor teardown promptly terminates the command process tree.
- The effective command deadline is recorded in runtime evidence.
- A regression process that runs indefinitely cannot occupy a full benchmark slot for 24 hours.
- Coordinate with #899's general timeout hierarchy and #729's bounded filesystem-search work rather than adding an unexplained global constant.

## References

- `scripts/harbor/run_gents.sh`
- `scripts/harbor/README.md`
- #899
- #729
- PR #988

jackzampolin:
Closes #1018.

## Problem

The Harbor profile set `GENTS_COMMAND_TIMEOUT_SECS=86400`, equal to the whole-request deadline, which turned the #985 foreground ceiling into a no-op. In the full Terminal-Bench 2.1 run, five trials sat 19–22 hours on pathological foreground commands (`./enc`, recursive `grep` over `/`, QEMU probes) with vLLM reporting zero active requests.

Under #985 semantics the single knob was both the omitted-`timeout_secs` default and the hard cap, so it could not express "return control quickly by default, but let the model opt into a longer run for a known-long build."

## What changed

**Runtime (`crates/gents`)** — the foreground default and the model-requestable maximum are now separate values:
- `resolve_command_timeout` takes a `foreground_max`: omitted `timeout_secs` → default; explicit requests clamped to `[1, max]`; background path (`spawn_process`, 10h budget) unchanged; still `min()`'d against the request deadline.
- `NativeTool::BashReadOnly`/`BashUnrestricted` carry `timeout_max`; builder methods renamed to `bash_*_with_timeouts(…, timeout, timeout_max)`.
- The bash tool schema advertises both (`default`, `maximum`) so the policy is model-visible, and the invariant `max = max(default, configured_max)` is enforced at every layer — a misconfigured pair can never push the cap below the default.

**CLI (`crates/gents-cli`)** — new `gents server --command-timeout-max-secs` (optional; unset ⇒ max = default, byte-for-byte the previous #985 behavior). The startup tracing line records both effective values, so the pair lands in the retained `gents-server.log` evidence; per-call attribution already exists via `timeout_ms`/`timed_out`/`status: "timeout"` in persisted command results.

**Harbor adapter (`scripts/harbor`)** — `GENTS_COMMAND_TIMEOUT_SECS` default 86400 → **600**, new `GENTS_COMMAND_TIMEOUT_MAX_SECS` default **3600**, threaded through both `run_gents.sh` and the `GentsAgent` env passthrough; `GENTS_REQUEST_TIMEOUT_SECS` stays 86400. README documents the command pair separately from the request deadline.

## Acceptance criteria (#1018)

- Harbor ceiling materially below the request deadline, documented separately: 600/3600 vs 86400 ✅
- Timed-out command → normal tool outcome the model can recover from: pre-existing #985 machinery, verified ✅
- Cancellation/teardown kills the process tree: pre-existing `managed_exec` setsid + group SIGTERM→SIGKILL, covered by `unrestricted_bash_timeout_kills_descendants_and_returns_promptly` ✅
- Effective deadline recorded in runtime evidence: per-call `timeout_ms` + both knobs logged at startup ✅
- No unexplained global constant: an optional operator flag defaulting to prior behavior, shaped so #899's budget derivation can subsume it ✅

## Design docs

Lean not required: pure configuration resolution, no lifecycle transition or invariant changes — fenced by conformance tests instead (`command_timeout_resolution_matches_advertised_schema`, `bash_schema_advertises_decoupled_default_and_max`, `command_timeout_max_ceiling_reaches_selected_bash_tool`, args tests). Spec and plan are in `docs/superpowers/`.

## Testing

- `cargo test -p gents`: 1791 passed, 0 failed (at the merge HEAD including the updated base)
- `cargo test -p gents-cli --bins`: 523 passed, 0 failed
- `cargo check --workspace --all-targets`: clean
- `sh -n` + `run_gents.sh` self-test: pass; `test_gents_agent.py`: 4/4

Note: 23 pre-existing `gents-cli` integration failures (a `reasoning_effort` manifest `Some("")` round-trip defect) exist in this stack's base — they pass at main's merge-base and have zero diff overlap with this branch; they should be fixed in the base stack before it reaches main.

### DIFF
--- a/crates/gents-cli/src/cli/args.rs
+++ b/crates/gents-cli/src/cli/args.rs
@@ -552,6 +552,11 @@ pub(crate) struct ServeArgs {
         help = "Foreground Bash command timeout in seconds: applied when a call omits timeout_secs and also the foreground cap (backgrounded runs use the built-in background lifetime budget)"
     )]
     pub(crate) command_timeout_secs: u64,
+    #[arg(
+        long,
+        help = "Foreground cap in seconds for explicit Bash timeout_secs requests. Defaults to --command-timeout-secs; values below it are raised to it (#1018)"
+    )]
+    pub(crate) command_timeout_max_secs: Option<u64>,
     #[arg(long = "cli-tool")]
     pub(crate) cli_tools: Vec<String>,
     #[arg(
--- a/crates/gents-cli/src/cli/args.rs
+++ b/crates/gents-cli/src/cli/args.rs
@@ -549,7 +549,7 @@ pub(crate) struct ServeArgs {
     #[arg(
         long,
         default_value_t = 120,
-        help = "Foreground Bash command timeout in seconds: applied when a call omits timeout_secs and also the foreground cap (backgrounded runs use the built-in background lifetime budget)"
+        help = "Foreground Bash command timeout in seconds: applied when a call omits timeout_secs, and the foreground cap unless --command-timeout-max-secs raises it (backgrounded runs use the built-in background lifetime budget)"
     )]
     pub(crate) command_timeout_secs: u64,
     #[arg(
--- a/crates/gents-cli/src/cli/args/tests.rs
+++ b/crates/gents-cli/src/cli/args/tests.rs
@@ -216,9 +216,14 @@ fn server_p2p_admission_flags_parse() {
 #[test]
 fn server_command_timeout_defaults_and_parses() {
     assert_eq!(parse_server(&[]).command_timeout_secs, 120);
+    assert_eq!(parse_server(&[]).command_timeout_max_secs, None);
     assert_eq!(
         parse_server(&["--command-timeout-secs", "300"]).command_timeout_secs,
         300
+    );
+    assert_eq!(
+        parse_server(&["--command-timeout-max-secs", "3600"]).command_timeout_max_secs,
+        Some(3600)
     );
 }
 
--- a/crates/gents-cli/src/commands/serve.rs
+++ b/crates/gents-cli/src/commands/serve.rs
@@ -304,9 +304,18 @@ pub(crate) async fn serve(mut args: ServeArgs) -> Result<()> {
         ),
     };
     tool_ceiling = tool_ceiling.with_command_timeout_secs(args.command_timeout_secs);
+    if let Some(max_secs) = args.command_timeout_max_secs {
+        tool_ceiling = tool_ceiling.with_command_timeout_max_secs(max_secs);
+    }
+    let effective_command_timeout_max_secs = args
+        .command_timeout_max_secs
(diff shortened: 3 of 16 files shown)


jackzampolin (commented):
Review summary

Reviewed 784c45ec against the current agent/add-atif-trace-projection base. I found no code-level blocker.

The policy remains consistent across the full path:

- CLI omission preserves the existing coupled default/max behavior; an explicit max is normalized to at least the default.
- ToolCeiling threads the normalized pair into both read-only and unrestricted Bash tools.
- The model-visible schema advertises the same default and maximum used by runtime resolution.
- Foreground omission uses the default, explicit requests clamp to the max, and background execution retains its separate bounded lifetime.
- run_command still intersects the resolved command timeout with the request deadline and returns ordinary timeout metadata/output for a command-cap expiry.
- Harbor now supplies 600s/3600s while retaining the 86400s request deadline.

Local verification passed: cargo fmt --all --check; the three command-timeout runtime tests; the schema and background-exemption tests; the CLI args test; runner self-test (11/11); ShellCheck; and adapter tests (4/4).

The PR is mergeable. Required GitHub CI is still in progress, so that remains the merge gate. The noted full gents-cli integration failures are base-stack debt rather than overlap with this patch, but they still need resolution before the stack reaches main.

Issue #730, opened 2026-07-15
Flaky: e2e_triggers::app_collection_pairing_fires_event_trigger_via_reconcile fails under parallel suite, passes isolated


jackzampolin:
Observed 2026-07-15 on `fix/search-tool-walk-bounds-729` (base `9539bc9c`, diff touches only `defra-native-fs-runner` + `toolset/{file_tools,native_runner}.rs` — no trigger/P2P surface).

`cargo test -p defra-agent` (full parallel suite, cold worktree): `app_collection_pairing_p2p_e2e::app_collection_pairing_fires_event_trigger_via_reconcile` FAILED (suite binary finished in 7.38s, 16 others passed). Immediate isolated rerun: passed in 8.47s.

Same class as #441 (trigger e2e timing under parallel suite load). Per the no-tolerated-flakes rule: filing for capture; needs a root-cause pass on its wait/deadline structure rather than a shrug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jackzampolin:
Cross-link: #743 is a third instance of this parallel-suite-only flake class (with #441). See https://github.com/source-inc/defra-agent/issues/743#issuecomment for the proposal to fix the class with one systemic pass over async-condition test waiting instead of per-test hunts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jackzampolin:
Completed by the P2P e2e hardening now on main: the suite serializes these transport tests, crosses the bounded conflict-retry boundary, and waits for actual collection convergence instead of sampling transient state.

PR #1034, opened 2026-08-05
Validate GraphQL collection identifiers at the trust boundary

### LINKED ISSUE #1003: GraphQL identifiers are interpolated unvalidated — agent-writable source_collection, and one unescaped filter
## The gap

CLAUDE.md's first sharp edge is *"Always `graphql::escape_graphql_string()` for anything interpolated into a GraphQL string."* Compliance is high across the codebase. But the rule only covers **string literals**, and the runtime also interpolates values into positions where escaping does not apply and would not help — GraphQL **identifiers**.

`trigger_engine/event_source.rs` interpolates `EventTrigger.source_collection` as a bare identifier in several runtime queries:

- `seed_seen_docs_for_collection` — `query {{ {collection}(limit: {limit}) {{ _docID }} }}` (lines 264-268)
- `fields_for` — `__type(name: "{name}")` (lines 105-112)
- `probe_filter` and `fetch_source_doc` (lines 416-420, 447-457)

Line 455 escapes an adjacent doc id, so escaping was considered for the neighbouring value but not for the identifier itself. Because identifiers sit outside string literals, `escape_graphql_string` would not sanitize them even if applied.

Validation on that field is a presence check only: `agent/document_view/snapshot.rs:578-583` (`if source_collection.is_empty() ... continue`) and `self_config/mod.rs:431-439`.

**`source_collection` is agent-writable via self-config.** That makes this a privilege-boundary question rather than a style question: an agent can influence the shape of a query the runtime issues, in a system whose whole premise is that DIDs and document ACLs bound what a principal may touch.

## Second, unverified site

`streaming/queries.rs:104` interpolates `response_key` into a filter **without** `escape_graphql_string`, unlike the neighbouring `request_id` sites in `interrupt.rs` and `background_tools.rs` which all escape. Callers (`streaming.rs:327`, `streaming.rs:619`) pass a `request_id`. Those look like content-addressed doc ids that cannot contain a quote, which would make it safe in practice — but that has not been confirmed, and it is the sole unescaped interpolation of its kind.

Note: the adversarial verification of both findings was cut short, so unlike the other review findings these are **reported but not independently confirmed**. Verify before fixing.

## Proposed work

1. Settle the `response_key` question: confirm every caller provides a doc id whose grammar excludes quotes. If so, escape it anyway for uniformity and note why; if not, it is a live injection.
2. Add an allowlist or grammar check for `source_collection` — validate against the set of collections actually present in the schema at reconcile time, rather than a non-empty check. An identifier cannot be escaped, so validation is the only defense.
3. Consider making the rule structural rather than conventional: a typed `CollectionName` newtype validated at construction would let the compiler carry the guarantee, instead of every future call site having to remember. CLAUDE.md's sharp-edge list is exactly the kind of knowledge that decays as a codebase grows.

## Repro / audit command

```
grep -rn 'format!' crates/gents/src --include=*.rs | grep -E 'query|mutation'
```
then check each interpolated value for either escaping (string positions) or validation (identifier positions).

jackzampolin:
Closes #1003

## The trust boundary, not the syntax

`EventTrigger.source_collection` is **agent-writable via self-config** and lands in GraphQL **identifier** positions in four trigger-engine queries (`seed_seen_docs_for_collection`, `fields_for`, `probe_filter`, `fetch_source_doc`). Identifiers sit outside string literals, so `escape_graphql_string()` — the codebase's standing sharp-edge rule — cannot apply there, and the only validation was a presence check. In a system whose premise is that DIDs and document ACLs bound what a principal may touch, that let a principal shape the queries the runtime issues. Escaping can't fix identifier positions; validation is the only defense.

## What changed

**Canonical validator** (`graphql.rs`): `validate_graphql_name` enforces the GraphQL Name grammar (`[_A-Za-z][_0-9A-Za-z]*`, ASCII only); `validate_collection_identifier` additionally rejects the spec-reserved `__` prefix so a "collection" of `__Type`/`__schema` cannot aim runtime queries at the introspection surface. `defra_query`'s existing `validate_identifier` now delegates to it — one seam, no query-builder abstraction.

**Enforced at three layers**, so no single bypass reaches a query string:
1. **Write time** — the self-config apply path rejects a non-conforming `source_collection` (covered by an e2e test driving injection payloads through `configure_automation`).
2. **Reconcile time** — snapshot resolve quarantines non-conforming triggers into `unavailable_event_triggers` with a `tracing::warn`. This covers trigger documents that never pass self-config, e.g. replicated from a peer.
3. **Query-build time** — the event source re-validates in `reconcile_subscriptions` (invalid names never enter the desired-collections set) and at every query-building function, so a snapshot assembled by any other path still cannot reach `format!`.

**The issue's second, unverified site is now verified and fixed**: `streaming/queries.rs` interpolated `response_key` (line 104) *and* `doc_id` (line 60) unescaped. The new tests demonstrate a hostile key previously produced a GraphQL parse error — a live syntax break, not a theoretical one. Both are now escaped; a hostile value matches nothing instead of breaking the query.

**Sweep result**: audited the other dynamic-collection interpolation sites (`defra_query/query.rs`, `defra_write`, `config_client/patch.rs`). They are already validated, or take collection names from the static `SelfConfigTarget` enum — no other agent-writable value reaches identifier position.

## Notes for review

- A grammar-valid but *unknown* collection name (e.g. `NoSuchCollection`) still reaches a query and fails harmlessly at execution with a warn/skip — it cannot change query shape. A schema-catalog allowlist was considered and left out: reconcile-time resolve has no reliable catalog snapshot for app collections that may land later, and the grammar check alone closes the injection channel.
- A validated Name cannot contain `"` or `\`, so the `__type(name: "{name}")` string-literal site in `fields_for` is safe by the same check.
- Tests were written first and observed failing (including the streaming parse-error repro) before the fixes landed.

Gates: `cargo test -p gents` fully green; `cargo check --workspace --all-targets` clean; rustfmt clean.

### DIFF
--- a/crates/gents-protocol/src/graphql.rs
+++ b/crates/gents-protocol/src/graphql.rs
@@ -101,6 +101,47 @@ pub struct GraphqlSessionShape {
     pub tool_results: Vec<AgentToolResultRow>,
 }
 
+/// Validate `name` against the GraphQL `Name` grammar:
+/// `[_A-Za-z][_0-9A-Za-z]*` (ASCII only). Anything interpolated into a
+/// GraphQL document in identifier position MUST pass this check first —
+/// escaping does not exist for identifiers.
+pub fn validate_graphql_name(name: &str) -> Result<()> {
+    let mut chars = name.chars();
+    match chars.next() {
+        Some(c) if c.is_ascii_alphabetic() || c == '_' => {}
+        Some(c) => {
+            return Err(anyhow!(
+                "invalid identifier {name:?}: must start with a letter or underscore, got {c:?}"
+            ))
+        }
+        None => return Err(anyhow!("invalid identifier: empty string")),
+    }
+    if let Some(c) = name
+        .chars()
+        .find(|c| !c.is_ascii_alphanumeric() && *c != '_')
+    {
+        return Err(anyhow!(
+            "invalid identifier {name:?}: only ASCII letters, digits, and underscore are allowed, got {c:?}"
+        ));
+    }
+    Ok(())
+}
+
+/// Validate a value used as a **collection name** in identifier position
+/// (e.g. `EventTrigger.source_collection`). On top of the Name grammar this
+/// rejects the `__` prefix, which the GraphQL spec reserves for
+/// introspection — a "collection" of `__Type` or `__schema` would aim a
+/// query at the introspection surface instead of a document collection.
+pub fn validate_collection_identifier(name: &str) -> Result<()> {
+    validate_graphql_name(name)?;
+    if name.starts_with("__") {
+        return Err(anyhow!(
+            "invalid collection name {name:?}: the __ prefix is reserved for GraphQL introspection"
+        ));
+    }
+    Ok(())
+}
+
 pub fn escape_graphql_string(value: &str) -> String {
     let mut escaped = String::with_capacity(value.len());
     for ch in value.chars() {
@@ -585,7 +626,14 @@ pub fn graphql_input_literal(value: &Value) -> Result<String> {
         Value::Object(map) => {
             let rendered = map
                 .iter()
-                .map(|(key, value)| Ok(format!("{key}: {}", graphql_input_literal(value)?)))
+                .map(|(key, value)| {
+                    // Keys render in identifier position, where escaping
+                    // cannot apply. Values reaching here are caller-supplied
+                    // (self-config patches are agent-authored), so an
+                    // unvalidated key is an injection into the mutation.
+                    validate_graphql_name(key)?;
+                    Ok(format!("{key}: {}", graphql_input_literal(value)?))
+                })
                 .collect::<Result<Vec<_>>>()?;
             Ok(format!("{{ {} }}", rendered.join(", ")))
         }
@@ -1015,6 +1063,30 @@ mod tests {
         assert!(rendered.contains("enabled: true"));
         assert!(rendered.contains(r#"name: "alpha""#));
         assert!(rendered.contains(r#"tags: ["a", "b"]"#));
+    }
+
+    #[test]
+    fn graphql_input_literal_rejects_object_keys_that_are_not_graphql_names() {
+        // Object keys land in identifier position in the rendered literal,
+        // where escaping does not apply. A self-config patch value is
+        // agent-controlled, so a non-Name key is an injection, not a typo.
+        let hostile = serde_json::json!({
+            "endpoint": {
+                r#"x: 1 }, api_key: "leaked" }) { _docID } #"#: 1
+            }
+        });
+        let err = graphql_input_literal(&hostile)
+            .expect_err("a non-Name object key must be rejected, not spliced");
+        assert!(
+            err.to_string().contains("identifier"),
+            "error should name the identifier rule: {err}"
+        );
+
+        // Well-formed nested keys still render.
+        let ok = serde_json::json!({ "outer": { "inner_1": "v" } });
+        assert!(graphql_input_literal(&ok)
+            .expect("valid Name keys render")
+            .contains("inner_1: \"v\""));
     }
 
     #[test]
--- a/crates/gents-protocol/src/graphql.rs
+++ b/crates/gents-protocol/src/graphql.rs
@@ -142,6 +142,90 @@ pub fn validate_collection_identifier(name: &str) -> Result<()> {
     Ok(())
 }
 
+/// Validate a caller-supplied GraphQL **filter-object fragment** — a value
+/// spliced into a query whole, as `EventTrigger.filter` is.
+///
+/// This is the third interpolation position, and neither of the other two
+/// defenses reaches it: escaping would destroy the object syntax, and the
+/// value is not an identifier. What makes it checkable is that a break-out
+/// must unbalance the fragment — to escape it has to close a `]`, `}` or
+/// `)` it never opened. So: require one balanced object literal, built only
+/// from the tokens a filter legitimately needs.
+///
+/// Deliberately strict. `(`, `)` and `#` never appear in a filter object,
+/// but each is load-bearing for an attack (`)` closes the enclosing field's
+/// argument list, `#` comments out the query's own tail), so they are
+/// rejected outright rather than balance-tracked.
+pub fn validate_graphql_filter_fragment(filter: &str) -> Result<()> {
+    let trimmed = filter.trim();
+    if trimmed.is_empty() {
+        return Err(anyhow!("invalid filter: empty"));
+    }
+    if !trimmed.starts_with('{') {
+        return Err(anyhow!("invalid filter: must be an object literal"));
+    }
+
+    let mut stack: Vec<char> = Vec::new();
+    let mut in_string = false;
+    let mut escaped = false;
+    let mut closed_at: Option<usize> = None;
+
+    for (index, ch) in trimmed.char_indices() {
+        if in_string {
+            match ch {
+                _ if escaped => escaped = false,
+                '\\' => escaped = true,
+                '"' => in_string = false,
+                _ => {}
+            }
+            continue;
+        }
+        // A depth-0 token after the object already closed means the value is
+        // not a single fragment — trailing text is someone else's query.
+        if closed_at.is_some() && !ch.is_whitespace() {
+            return Err(anyhow!(
+                "invalid filter: trailing content after the object literal"
+            ));
+        }
+        match ch {
+            '"' => in_string = true,
+            '{' | '[' => stack.push(ch),
+            '}' | ']' => {
+                let opener = stack
+                    .pop()
+                    .ok_or_else(|| anyhow!("invalid filter: unbalanced {ch:?} at byte {index}"))?;
+                let expected = if ch == '}' { '{' } else { '[' };
+                if opener != expected {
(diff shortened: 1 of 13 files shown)


jackzampolin:
### Code review

Found 1 issue:

1. `EventTrigger.filter` is agent-writable and spliced raw into a runtime query, while the doc comment one line below the validation this PR adds asserts it is "operator-authored and validated at apply time." That claim was true at `c714304f` but was invalidated by `1b7d22e5`, which added `configure_automation` support for EventTrigger and put `"filter"` in `EventTrigger::writable_fields()`. Nothing on the self-config path validates it — the filter syntax probe the comment refers to lives only in the CLI apply path (`gents-cli/src/desired_state/validate.rs`), and it probes `filter: {filter}` rather than the runtime's `_and: [ {filter} ]` embedding, so passing it would not imply safety here anyway.

https://github.com/source-inc/gents/blob/7ecf966c32385dfa1040ff1272d40565971d5def/crates/gents/src/trigger_engine/event_source.rs#L409-L434

Confirmed with a PoC rather than inferred. Replicating `probe_filter`'s construction verbatim with `filter` set to `{} ] }, limit: 1) { _docID } AgentBehavior(filter: { _and: [ {} ] }, limit: 1) { system_prompt } PocEvent(filter: { _and: [ {}` yields a well-formed query with zero GraphQL errors and returns `{"AgentBehavior": [{"system_prompt": "TOP-SECRET-PROMPT"}], ...}` — a collection the trigger has nothing to do with, read through the trigger engine's own node handle, which is not scoped to the writing agent's identity. This is the same injection class as #1003, in a sibling field of the same struct, reached through the same tool.

To be precise about impact: `probe_filter` only reads `data[source_collection]` and returns a bool, so the injected selection's data is fetched and discarded, and aliasing it onto the expected response key collides with the real field under GraphQL field-merging. So this is confirmed "the runtime executes attacker-chosen reads across ACL boundaries," not a proven exfiltration channel back to the agent.

Not fixed here — this PR is scoped to identifier positions, and both real fixes are larger: removing `filter` from `writable_fields()` touches a Lean-mirrored set (`proofs/Proofs/SelfConfig/Apply.lean`), and validating the fragment needs its own quote-aware balanced-delimiter check or a JSON representation migration. Filed as #1038 with the PoC and three fix options. `5c201bf` corrects the false comment and the `graphql` module doc (which implied its two defenses covered every interpolation position) so the stale claim doesn't outlive the review that found it.

Also checked, not reported: the schema-catalog allowlist and typed-`CollectionName`-newtype suggestions from #1003 (both scored as non-issues — the former is disclosed and justified in the PR body, the latter runs against CLAUDE.md's "queries are built inline by convention"), and the `__`-prefix asymmetry between `defra_query`'s `validate_identifier` and `validate_collection_identifier` (pre-existing, gated by an operator-configured scope allowlist).

jackzampolin:
### Ultracode re-review — one P0 found and fixed here, two follow-ups filed

Ran a 16-agent workflow over the branch (five lenses: security completeness beyond the diff, right-sizing, behavioral regression, test quality, consistency), each finding put through adversarial refutation. Five of ten findings survived. I verified the load-bearing ones against the code myself before touching anything.

**P0, fixed in `c893d9a8`: object-valued self-config patches injected into the config mutation.**

`gents-protocol::graphql::graphql_input_literal` rendered JSON object keys as bare identifiers — `format!("{key}: {}", ...)` — and `ensure_admissible` checked field *names* only, iterating `for (field, _) in patch` and ignoring values. The self-config tool schema is `additionalProperties: true`, so patch values are agent-authored arbitrary JSON. Reproduced end-to-end: a patch on `InferenceBackend.endpoint` wrote `api_key`, the secret `read_doc_in_txn` deliberately refuses to even select, and an object-valued `EventTrigger.filter` forged an `AgentBehavior` in a different collection under the agent DID. Strictly more powerful than the bug this PR was opened for, in the adjacent field of the same tool call — so it was fixed here rather than deferred.

Two changes, one per layer. `graphql_input_literal` now validates every object key as a GraphQL Name (structural: no mutation can be built with an injected key regardless of caller). `ensure_admissible` now rejects object values and lists with non-scalar elements — which **restores the model rather than adding a rule**: Lean's `FieldValue` is a `String` (`proofs/Proofs/SelfConfig/Apply.lean`), so a non-scalar is a state the spec cannot express. Every writable column across all six targets is a scalar or list of scalars, so nothing legitimate is refused. No Lean change is owed for the same reason, and the 322 conformance tests stay green as the fence.

**On "are we adding the right things, or spraying code?"** — the review's verdict, which I agree with:

- The validators moved *into* `gents-protocol` and are re-exported by `gents`, so the mutation renderer shares one definition. `crates/gents/src/graphql.rs` is net **-52 lines**; this deleted a copy rather than adding a third.
- **Removed one genuinely redundant check**: `fetch_source_doc` validated a binding that `fields_for` re-validates three lines later and `?`-propagates before either query is built. Gone.
- **Kept the other four**, and the reasoning matters: they looked redundant, but their alternative safety argument runs four hops across ~400 lines *and crosses a snapshot re-borrow* where `build_intents_for_all_matching` may read a newer generation than the one that populated `desired_collections`. Deleting `probe_filter`'s guard would be especially wrong with its sibling `filter` an open injection point.
- **The three layers are not the same gate thrice.** Layer 1 is the privilege boundary; layer 2 catches documents that never pass self-config (replicated from a peer — the point of a P2P control plane); layer 3 is where the string becomes query text.
- **Trimmed the comment I added in `5c201bf`** — it narrated review history ("this doc used to claim…"), which is talking to the reviewer, not the next reader. Now states the constraint, plus the review's correct point that the CLI probe *executes* the filter rather than validating it.
- **Added the patch-path case** to the e2e boundary test; it was create-only, and the sibling `ensure_stored_automation_owned` in the same match arm is deliberately create-exempt, which is exactly the asymmetry that invites drift.

**Filed, not fixed here** (both pre-existing and outside this PR's scope): #1044 — `gents-cli`'s apply-time EventTrigger probes splice `source_collection` unvalidated and unescaped, the same class fixed in `streaming/queries.rs`. #1045 — `BoundedWriteTool` splices collection and field names into identifier positions; low severity, not a privilege boundary, since `write_tools` is apply-managed.

Refuted and not acted on: the claim that all five per-call-site validators are dead (partly true, but only one deletion is safe), and that the `probe_filter` comment read as reassurance (it did not).

Gates: `cargo test -p gents` and `-p gents-protocol` green, `cargo check --workspace --all-targets` clean, rustfmt clean. One lib test flaked once mid-run and passed on rerun — that is the already-filed #330, unrelated to this change.

jackzampolin:
Pushed `42db0556`. #1038 is closed in this PR; the P1 is answered but deliberately not coded around — reasoning below.

**#1038 (filter injection) — fixed.** `EventTrigger.filter` is the third interpolation position: spliced in as a whole object fragment, where escaping would destroy the syntax and identifier validation does not apply. What makes it checkable is that a break-out *must* unbalance the fragment — to escape, a payload has to close a `]`, `}` or `)` it never opened. `validate_graphql_filter_fragment` requires exactly one balanced object literal built only from the tokens a filter needs, and rejects `(`, `)` and `#` outright: none appear in a legitimate filter, and each is load-bearing for an attack. Applied at write time, resolve time (replicated triggers), and in `probe_filter` itself. The confirmed PoC payload is a test case; `{ kind: { _eq: "signup" } }` and friends still commit.

**P1 (collection authorization) — confirmed real, and I'm not fixing it in app code.** You're right that an EventTrigger on `AgentMessage` gets the full document read as node root and rendered into an agent-controlled template, with no injection required. I traced it and agree.

I started building the obvious gate — reject `source_collection ∈ ALL_COLLECTION_NAMES` at the self-config boundary — and then dropped it, for three reasons:

1. It is exactly the thing `config_client/mod.rs:9-13` rules out: *"authorization is enforced at the node, not by app-level ownership checks."* A denylist is a second notion of what a principal may read, which drifts from real ACP the moment ACP lands and gives false assurance until then.
2. It would break legitimate use. `Host`, `InferenceBackend` and `Task` are all real `source_collection` values today via the CLI apply path — operator-authored triggers on runtime collections are a supported pattern, and provenance isn't recoverable at resolve time.
3. It cannot express the case that actually matters once ACP exists: principal A watching principal B's *app* collection.

Worth flagging for whoever picks this up: **making the read identity-scoped would also change nothing today.** `grep -rn '@policy' crates/` returns zero hits — no collection carries an ACP policy, so DefraDB treats every document as readable and an identity-scoped read returns exactly what a root read returns. Scoping the call before registering policies is ceremony that reads like a fix.

Filed as #1048 with the full analysis and the ordering the real fix needs: attach policies, decide which identity a trigger read assumes, then scope the reads. I also considered a Lean fence and rejected it as the same trap one level up — the model has no authorization predicate, so adding one now would be inventing ACP semantics ahead of the database's.

**On the test suite**: `e2e_triggers::app_collection_pairing_fires_event_trigger_via_reconcile` failed twice while I was verifying this, and my first read was that I had caused it. Running it 6× against the change showed 6/6 pass, and the failures match the already-filed P2P flake (#730/#1023). Recording that because the intermediate evidence pointed the wrong way.

Gates: `cargo test -p gents` (all 17 targets) and `-p gents-protocol` green, `cargo check --workspace --all-targets` clean, `cargo fmt --all --check` clean.

Issue #1036, opened 2026-08-05
Investigate: replicator POST hang + reconciler install failing every sweep in the #984 conflicted state


jackzampolin:
Split out of #984's secondary observations (live, 2026-07-31, pre-restart, while the home held a merge-parked unique-index-conflicting `PeerEndpoint`):

- `POST /api/v0/p2p/replicators` hung until client timeout while GETs served fine.
- The pairing reconciler's data-plane replicator install failed every sweep — the peer's replicator list stayed empty despite a live `DataPlanePairingDesired` row (silent after the first transition-aware WARN).

Initial triage during the #984 fix (upstream handler read, pinned rev f9e21c6):

- The POST handler's initial doc replay is `tokio::spawn`ed, so the hang is **not** the replay itself. Remaining awaits on that path: `require_permission`, `transport.dial` (skipped when already connected), and `sync_coordinator.create_replicator_info` — an actor call into the coordinator. Most plausible wedge: the coordinator actor busy/blocked (possibly retry-driving pushes involving the conflicted doc), starving the replicator-info call while read-only GETs (served from other state) stay healthy.
- The reconciler failing every sweep is consistent with its install POSTs timing out against that wedged endpoint, i.e. a consequence, not an independent gents bug — but this is unconfirmed.

The boot fix for #984 does not touch this path. Needs a repro under the conflicted state (a parked doc plus an active replicator install) to confirm the coordinator-wedge hypothesis; if confirmed, the fix is upstream (defradb.rs) and should be filed there with specifics.

Related: sourcenetwork/defradb.rs#1308 (reindex/merge asymmetry), sourcenetwork/defradb.rs#1309 (untracked sweep tasks hold the store past shutdown).

jackzampolin:
Closing after backlog audit. This investigation was tied to a merge-parked unique-index conflict on the old DefraDB f9e21c6 pin. The current pin includes substantial replicator retry, reconnect, and shutdown work and was exercised by the later 19-node restart/remesh qualification in #1157. There has been no recurrence of this exact state; a reproduction against the current pin should be filed as a fresh upstream-focused defect with current evidence.

PR #1037, opened 2026-08-05
fix(migration): merge-parked unique-index conflict must not brick boot

### LINKED ISSUE #984: Unique-index conflict persisted at merge time bricks the home at next boot (eager materialization hard-fails)
Follow-on from #983, observed live 2026-07-31 — severity higher than the error-loop itself.

Sequence:
1. A re-paired client replayed a stale `PeerEndpoint` doc; merge hit the unique `did` index conflict and handled it gracefully: `unique index conflict during merge: existing holder wins the deterministic pick; the incoming document is persisted unindexed` (db_index::index_manager).
2. Server ran fine until restart.
3. On next boot: `Error: ensure_migrations — eager materialization failed: PeerEndpoint: storage error: can not index a doc's field(s) that violates unique index.` — the home no longer boots at all.

So the merge path persists a document the boot path refuses to accept: a single foreign/stale doc becomes a poison pill that permanently bricks the home on the next restart. Boot-time materialization should apply the same deterministic-winner tolerance as merge (park the loser unindexed and continue), or the merge path should quarantine rather than persist into the collection.

Also observed in the same conflicted state, pre-restart (possibly consequences, noting for the investigation): `POST /api/v0/p2p/replicators` hung until client timeout while GETs served fine, and the pairing reconciler's data-plane replicator install failed every sweep — the peer's replicator list ended up empty despite a live `DataPlanePairingDesired` row (silent after the first transition-aware WARN).

Recovery required a full home reset.

jackzampolin:
Closes #984

## The bug

A re-paired client replays a stale `PeerEndpoint` doc; the P2P merge hits the unique `did` index conflict and handles it **by design** (defradb.rs#1111): the lexicographically smallest public docID keeps the index entry, the loser is persisted unindexed, replicas converge. The server runs fine — until restart. Boot's `ensure_migrations` → eager materialization calls defradb's `materialize_collection`, which does `remove_all` + `bulk_index` with **strict** unique saves over every persisted doc. The parked loser re-raises `UniqueConstraintViolation`, gents treated any collection failure as fatal, and since materialization runs on every boot, the home was bricked until a full reset.

The asymmetry is the bug: one subsystem's documented graceful outcome was another subsystem's poison pill.

## The fix (boot tolerant + observable)

`materialize_collections` now classifies exactly this error class as merge-parked residue:

- **Safe to skip**: the failed reindex transaction upstream never commits, so the on-disk index state — the merge-time deterministic winner — survives intact. Skipping the collection's eager pass preserves the exact pick the merge made; boot re-derives nothing.
- **Observable**: a `tracing::warn!` names the collection, unique index, winner docID and parked docID(s). The parked docs are recovered by a non-index scan grouped by unique-field values with the same smallest-docID winner rule the merge uses, so the reported split matches what merge persisted. Details are recorded in a new `MaterializationStats::parked_unique_conflicts` and surfaced through `MigrationReport::warnings`. Diagnostics are best-effort and can never fail boot.
- **Everything else stays fatal**: only the unique-violation class (defradb's shared error constant) is tolerated; all other materialization failures still hard-fail as before.

## The repro test (the heart of the PR)

`crates/gents/tests/e2e_lifecycle/merge_parked_unique_boot_e2e.rs` drives the real incident sequence: two P2P nodes create `PeerEndpoint` docs with the same unique `did` (distinct docIDs), a replicator replays the client's doc onto the home, the test asserts the merge parked state (scan sees 2 docs, the unique index resolves to the smallest docID), then restarts the home via `simulate_process_crash` — which re-runs the production boot path.

- **Red run** (tolerance disabled): fails with the incident error verbatim — `eager materialization failed: PeerEndpoint: storage error: can not index a doc's field(s) that violates unique index.`
- **Green run**: boot succeeds; the report names PeerEndpoint and the parked docID; both docs survive; the winner is unchanged after boot.

The reopen retries on lock errors only, with a bounded deadline: `EmbeddedNode::shutdown` doesn't join the pending-DAG sweep loops, whose coordinator Arc holds the store for up to one 60s sweep interval (filed as sourcenetwork/defradb.rs#1309). A real process restart has no such lag, so the retry models process exit, not a race in the code under test.

Unit tests in `gents-migration` fence the error classifier (violation anywhere in the anyhow chain → tolerated; anything else → fatal).

## Lean

Boot-time materialization sits outside the proof boundary: "DefraDB storage-engine correctness remains external" is an explicit model boundary, and the recovery-convergence theorems (`Proofs/Recovery/`) model request-state repair, not schema/document materialization. This change alters no legal transitions and no fenced invariants, so no spec change. Recovery-semantics review notes:

- **Winner identity**: boot does not re-pick a winner — it preserves the merge-time index state (rollback semantics), and the *reporting* applies the identical smallest-public-docID rule.
- **No loss / no double-materialization**: parked docs remain persisted and scan-readable; the skipped eager pass commits nothing, so nothing is partially applied. Docs left at older schema versions continue to be lens-migrated on read.

## Upstream halves (filed, not silently half-fixed)

- sourcenetwork/defradb.rs#1308 — `bulk_index` should resolve unique conflicts with the merge's deterministic-winner rule (order-independent, converges to exactly the merge outcome), and expose a typed error so embedders don't string-match.
- sourcenetwork/defradb.rs#1309 — `setup_p2p`'s untracked sweep tasks hold the store past shutdown.
- #1036 — follow-up for #984's secondary observations (replicator POST hang, reconciler install failures), triaged as a probable coordinator wedge, independent of this fix.

## Verification

- `cargo test -p gents-migration` ✅ · `cargo test -p gents` (full package suite) ✅ · `cargo check --workspace --all-targets` ✅
- Red/green runs of the repro as above.

### DIFF
--- a/crates/gents-migration/src/engine.rs
+++ b/crates/gents-migration/src/engine.rs
@@ -43,6 +43,9 @@ pub async fn ensure_migrations_with_registry(
     verify_managed_lineages(node, registry, &mut report).await?;
 
     report.materialization = materialize::materialize_all(node, registry).await?;
+    report
+        .warnings
+        .extend(report.materialization.parked_unique_conflicts.iter().cloned());
 
     info!(
         baseline_registered = report.baseline_registered,
--- a/crates/gents-migration/src/engine.rs
+++ b/crates/gents-migration/src/engine.rs
@@ -43,9 +43,13 @@ pub async fn ensure_migrations_with_registry(
     verify_managed_lineages(node, registry, &mut report).await?;
 
     report.materialization = materialize::materialize_all(node, registry).await?;
-    report
-        .warnings
-        .extend(report.materialization.parked_unique_conflicts.iter().cloned());
+    report.warnings.extend(
+        report
+            .materialization
+            .parked_unique_conflicts
+            .iter()
+            .cloned(),
+    );
 
     info!(
         baseline_registered = report.baseline_registered,
--- a/crates/gents-migration/src/materialize.rs
+++ b/crates/gents-migration/src/materialize.rs
@@ -5,8 +5,10 @@
 //! merged in #1232): advances document blobs, stored version keys, and
 //! secondary indexes without creating CRDT commits or replication events.
 
+use std::collections::BTreeMap;
+
 use defra_node::EmbeddedNode;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::error::{Error, Result};
 use crate::registry::Registry;
@@ -15,7 +17,16 @@ use crate::report::MaterializationStats;
 /// Eagerly materialize every managed collection.
 ///
 /// Returns stats on success. Any collection-level failure is a hard error —
-/// after lineage verification every managed collection is expected to exist.
+/// after lineage verification every managed collection is expected to exist —
+/// with one deliberate exception: a unique-index violation raised by a
+/// document a P2P merge already parked unindexed (#984). The merge path
+/// resolves live unique conflicts deterministically (smallest public docID
+/// keeps the index entry; the loser persists unindexed), so the same document
+/// re-raising the violation during the boot-time reindex is expected residue,
+/// not corruption. The failed reindex transaction rolls back, preserving the
+/// merge-time winner; boot skips the collection's eager pass, records the
+/// parked state in the report, and continues. A foreign stale doc must never
+/// be a poison pill.
 pub async fn materialize_all(
     node: &EmbeddedNode,
     registry: &Registry<'_>,
@@ -35,6 +46,7 @@ pub async fn materialize_collections(
         collections_failed: 0,
         skipped_upstream_missing: false,
         read_through_scans: 0,
+        parked_unique_conflicts: Vec::new(),
     };
 
     let mut failures: Vec<String> = Vec::new();
@@ -44,6 +56,17 @@ pub async fn materialize_collections(
             Ok(n) => {
                 stats.documents_materialized += n;
                 debug!(collection = %name, documents = n, "materialized collection");
+            }
+            Err(e) if is_unique_index_violation(&e) => {
+                let detail = describe_parked_unique_conflict(node, name, &e).await;
+                warn!(
+                    collection = %name,
+                    detail = %detail,
+                    "eager materialization skipped: a merge-parked unique-index \
+                     conflict makes a full reindex impossible; the merge-time \
+                     deterministic winner keeps the index entry and boot continues"
+                );
+                stats.parked_unique_conflicts.push(detail);
             }
             Err(e) => {
                 stats.collections_failed += 1;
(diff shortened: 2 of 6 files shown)


jackzampolin:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Five parallel reviewers audited CLAUDE.md adherence, scanned for bugs, and attacked the recovery semantics specifically; all findings scored below the confidence threshold. The load-bearing claims verified against the pinned defradb.rs source (f9e21c6):

- **No loss / no double-materialization**: the failed reindex in [`materialize_collection_inner`](https://github.com/sourcenetwork/defradb.rs/blob/f9e21c682c93fd64e85434b7804b7a4112c51216/crates/db/src/migration/reindex.rs) never reaches `write_txn.commit()` on the error path; all four storage backends buffer writes until commit, so the pre-boot index state (merge-time winner) is restored exactly.
- **Winner identity is preserved, not re-derived**: boot's tolerance keeps the merge-time index entry via rollback; the reported winner (scan + smallest-public-docID) provably agrees with the actual holder, because each merge conflict in [`save_resolving_unique_conflict`](https://github.com/sourcenetwork/defradb.rs/blob/f9e21c682c93fd64e85434b7804b7a4112c51216/crates/db-index/src/index_manager/mod.rs) compares against the current live holder — the held entry is always the running minimum over the conflicting set, equal to the global minimum regardless of arrival order or 3+-way conflicts.
- **Lean fences unaffected**: recovery-convergence models request-state repair; DefraDB storage-engine correctness is an explicit external model boundary. No legal transition or fenced invariant changes.
- **Read-path safety of the skipped eager pass**: lens transforms apply on every read via the lensed fetchers; eager materialization is a write-back cache, not a correctness requirement.

jackzampolin:
### Adversarial review round — disposition

One round, as scoped. Five parallel reviewers attacked CLAUDE.md adherence, bug surface, git history, prior-PR conventions, code-comment accuracy, and the recovery semantics specifically. Six findings; none crossed the confirmation bar, but four conventions were applied anyway in 5549faf8:

**Applied**
- Test GraphQL values now go through `escape_graphql_string` (CLAUDE.md sharp edge is unconditional; most sibling e2e files escape even literals).
- `describe_parked_unique_conflict` validates collection/field names as `[A-Za-z0-9_]` identifiers before interpolating — escaping cannot apply to identifier positions, and this file's own pre-#947 history carried the same guard.
- Parked-doc detail hedges provenance ("consistent with a P2P merge having parked…") since the storage error carries no provenance signal.
- The lens-first design doc's Ship Matrix now records the #984 carve-out to phase C's hard-fail contract (precedent: #947 updated it in-commit).

**Rebutted (no change)**
- *"Lock-retry loop is a flake."* It is a bounded deterministic wait on a structurally-bounded condition: `run_pending_dag_resync` checks the shutdown flag at each wake on a 60s cadence and is untracked by `P2PLifecycleInner::shutdown` (sourcenetwork/defradb.rs#1309), so the in-process lock lag is hard-bounded at one sweep interval; the 90s deadline has 1.5× headroom and the retry matches one exact error string. A real process restart has no lag at all.
- *"Classifier tolerates non-merge unique violations."* No provenance signal exists in the error, and local writes are strictly unique-enforced (`on_document_create` uses the non-healing save path), so duplicate live unique values among persisted docs can only arise via merge today. The tolerance is loud (warn + report), never silent, and the typed-error/provenance fix is upstream (sourcenetwork/defradb.rs#1308).
- *"Parked details dropped when another collection fails fatally."* Real but immaterial: `ensure_migrations` returns `Err` in that case, no caller reads a report on the Err path (verified across gents, gents-cli, desktop bootstrap), boot still fails for the genuine fatal reason, and the tracing warn retains the parked docIDs.

Verification after fixes: `cargo test -p gents-migration` ✅, the e2e repro ✅, `cargo check --workspace --all-targets` ✅ (full `cargo test -p gents` passed pre-fix; the fixes touch only the test file, a diagnostics guard, a log string, and a doc).

Issue #1041, opened 2026-08-05
Flaky: codex_shim_does_not_clobber_session_behavior_id — server connection refused mid-test under load


jackzampolin:
# Flaky: codex_shim_does_not_clobber_session_behavior_id — server connection refused mid-test under load

## Observation

During stress verification for #990/#989/#933 (branch `fix/flaky-cli-suite`), iteration 19 of a 20× `cargo test -q -p gents-cli --test cli_codex_shim` loop (running concurrently with a second worktree's build/test loop — heavy host load) failed:

```
---- codex_shim_does_not_clobber_session_behavior_id stdout ----
Error: posting GraphQL to http://127.0.0.1:53888/api/v0/graphql
Caused by:
    ...
    3: Connection refused (os error 61)
test result: FAILED. 29 passed; 1 failed; 10 ignored; finished in 19.18s
```

The test had progressed past `wait_for_port`/readiness, so the server was up and then stopped listening mid-test — either the serve process died (panic/abort) or the listener was lost. 1 occurrence in ~24 suite runs under load; 0 in 20 base-commit runs on a quieter host.

## Evidence gap

The failure path doesn't print the server's captured stdout/stderr (`serve.captured_output()` exists in the helpers and is already used by other tests' timeout paths), so whether the server panicked — and why — is unknown. First step: make GraphQL-post failures in `cli_codex_shim` helpers append the server's captured output like the timeout paths do, then re-stress to capture the actual death.

## Repro

Full `cli_codex_shim` target in a loop under concurrent host load; observed once at ~4% per suite run on a loaded host.

jackzampolin:
Base-commit repro captured: a load-matched 30×/30× A/B (main @ b5105682 vs #1042's branch, loops started simultaneously on the same host) hit this family **twice on base, zero on the branch**:

- `base2-5`: `codex_shim_config_value_write_model_mutates_behavior` — `Error: connecting to codex-shim websocket` (connection refused at ws connect)
- `base2-19`: `codex_shim_does_not_clobber_session_behavior_id` — same ws-connect refusal signature

So this is confirmed pre-existing on main and has (at least) two signatures: refusal at initial ws connect, and refusal on mid-test GraphQL posts. Both are evidence-starved for the same reason — the helper failure paths don't dump the server's captured output. The capture fix in #1042 should make the next occurrence diagnosable.

jackzampolin:
[...]

Issue #989, opened 2026-08-01
Flaky: codex_shim_derives_git_info_and_persists_early_rename fails under full-workspace parallel test run


jackzampolin:
## Summary

`codex_shim_derives_git_info_and_persists_early_rename` (`crates/gents-cli/tests/cli_codex_shim.rs`) failed once during a full sequential gate (`cargo test -p gents && cargo test -p gents-cli`) on 2026-07-31 (branch `agent/973-grok-oauth`, local macOS, worktree checkout), suite result `29 passed; 1 failed; 10 ignored`.

## Evidence limits

The gate's per-suite output was piped through `tail -6`, so the panic message was truncated — only the failure name survived. Immediately afterwards:

- the test passed standalone (`cargo test -p gents-cli --test cli_codex_shim -- codex_shim_derives_git_info_and_persists_early_rename`), and
- the full `cli_codex_shim` suite passed (`30 passed; 0 failed; 10 ignored`).

So this looks like contention-dependent flakiness (parallel test processes / git or filesystem racing in the shim's early-rename path), not a deterministic regression. Note the same working tree also carried an unrelated in-progress change to bash tool timeouts (#985); nothing in that diff touches the codex shim, but flagging for completeness.

## Ask

Per project policy (flaky tests are defects): reproduce under load (e.g. `cargo test -p gents-cli` while the gents suite runs), capture the panic, and fix the underlying race. If it reproduces, attach the full failure output here.

jackzampolin:
Status from #1042: the original panic was truncated (`tail -6`) and was not individually reproduced during stress verification, so this is a **class fix, not a confirmed individual fix** — hence no `Closes`. Every operation on this test's failing surface (`thread/start` ×2, `thread/name/set`'s `upsert_AgentConversation`, `thread/read`) flowed through the codex-shim store's unretried `EmbeddedNode::execute` path, and the test's error-reporting shape matches #933's captured startup transaction conflict exactly. #1042 routes that whole store path through the bounded DefraDB conflict retry (#440). The test passed in every one of 24+ full-target stress runs on the branch. If it recurs post-merge it is a different defect — please capture the full panic this time (the gate's `tail -6` is what cost us the evidence here).

jackzampolin:
[...]

Issue #440, opened 2026-06-09
Centralize DefraDB transaction-conflict retry; remove test-helper band-aid


jackzampolin:
## Problem

`EmbeddedNode::execute()` (used everywhere as `node.execute(graphql)`) is **auto-commit** — each call is its own implicit transaction — and does **not** retry internally. DefraDB uses optimistic concurrency at the storage layer and surfaces races between concurrent writers to overlapping data as a retryable error:

```
commit error: datastore error: storage error: transaction conflict. Please retry
```

Retry is the DB's *prescribed* remedy (no transaction config avoids the conflict — explicit transactions would conflict too). But today **nothing in the runtime retries DB conflicts** — the existing `error::is_retryable()` covers only inference/streaming (rig) errors, not DefraDB writes. Both the runtime write path and tests are exposed.

## Symptom

Flaky integration test `cli_codex_shim::codex_shim_completes_blank_materialized_terminal_message` — it seeds an `AgentResponse`/`AgentRequest` while the live runtime is concurrently writing the same docs, hitting the transient conflict intermittently.

## Stopgap (already merged)

A bounded retry on `"transaction conflict"` was added to the shared test helper `crates/defra-agent-cli/tests/support/graphql.rs::graphql_query` (commit `9f05ce4b`). This is a **band-aid in the test layer only**.

## Proper fix

1. **Centralize DB conflict-retry** at the defra-agent↔node boundary: a shared `execute_with_conflict_retry` (bounded attempts + backoff) that classifies `"transaction conflict"` the way `is_retryable()` classifies inference errors. Use it in the runtime write path (`crates/defra-agent/src/document_config/*.rs` upserts/updates and other `node.execute` mutation sites), and replace the test-helper band-aid with it.
2. **Review tests that double-write live docs**: the codex_shim seed races the runtime by design. Even with centralized retry, prefer driving completion through the runtime rather than seeding docs the runtime is concurrently producing.

## Upstream

Filed a companion request on defradb.rs for an opt-in built-in retry on `execute()` so consumers don't each reimplement it (link to follow in a comment).

## Affected
- `crates/defra-agent/src/document_config/*.rs` (mutation `node.execute` sites)
- `crates/defra-agent-cli/tests/support/graphql.rs` (current band-aid)

jackzampolin:
Upstream companion request: sourcenetwork/defradb.rs#1031 (opt-in built-in retry on `EmbeddedNode::execute`).

PR #1042, opened 2026-08-05
Kill the CLI-suite flake class: parallel-run contention (#990, #933, #989)

### LINKED ISSUE #933: Flaky cli_codex_shim: thread/start surfaces startup transaction conflict
## Symptom

While stress-validating the test-setup synchronization for #927 against `main`, the complete `cli_codex_shim` integration target passed seven consecutive iterations, then iteration 8 failed before the background-interrupt scenario began: `thread/start` surfaced a raw DefraDB transaction conflict.

This is distinct from #927: that issue covers explicit config transaction commits (`ConfigApplyTxn::commit`, HTTP 409). This failure came from the embedded Codex-shim query path during `thread/start`.

## Exact observation

Observed on 2026-07-29 from isolated branch `agent/fix-cli-config-tx-conflict`:

- base: `16619abf`
- branch head: `8e1e20a0`
- command: `for run_no in {1..20}; do cargo test -q -p gents-cli --test cli_codex_shim || exit $?; done`
- first seven target iterations passed (30 passed, 10 explicitly ignored each)
- iteration 8 failed:

```text
running 40 tests
iiiiii 6/40
codex_shim_interrupt_completes_with_running_background_tool --- FAILED
.........i.iii...................

---- codex_shim_interrupt_completes_with_running_background_tool stdout ----
Error: Codex shim returned error for request 103: Codex shim request `thread/start` failed: GENTS Codex shim query failed: [QueryResponseError { message: "commit error: datastore error: storage error: transaction conflict. Please retry", path: None, locations: None }]

test result: FAILED. 29 passed; 1 failed; 10 ignored; finished in 26.62s
```

The test waits for `AgentRuntime.process_state=ready` and then immediately initializes the shim and starts a thread. It does not wait for startup reconciliation to settle. Current `commands/codex_shim/store.rs::query_node_json` calls `EmbeddedNode::execute` directly rather than the shared bounded DefraDB conflict-retry boundary.

## Investigation leads

1. Synchronize the test with runtime startup quiescence if the scenario is not intended to cover concurrent startup writes.
2. Separately assess whether the real Codex-shim `thread/start` query path should use the centralized bounded conflict retry from #440; preserve existing error strings/retry semantics unless that production behavior change is explicitly justified.

The failing stress run included no production changes and no fix for this path.

### LINKED ISSUE #990: Flaky: demo_pair_delegate_materializes_remote_worker — demo pair flow hits default port 9191 under full-suite parallel runs
## Summary

`demo_pair_delegate_materializes_remote_worker` (`crates/gents-cli/tests/cli_demo.rs:453`) failed during a full local gate (`cargo test -p gents-cli`, run 2026-08-01 on `agent/973-grok-oauth` + local commit under review) and passed on immediate isolated rerun (43.8s). Same flake class as #989: full-workspace parallel test contention.

## Captured failure (full output this time)

```
demo pair did not converge:
...
  initializing node B (worker)…
  starting node B…
  pair failed: gents p2p failed: Error: sending GET request to http://127.0.0.1:9191/api/v0/p2p/info
Caused by:
    ...
    3: Connection refused (os error 61)
  delegate failed: not paired — run `pair` first
[tool done] spawn_subagent (...) => { "ok": false, "failure_class": "tool_not_allowed", ... "subagent spawning is not enabled for this b..." }
```

## Analysis

The pair step queried `http://127.0.0.1:9191/api/v0/p2p/info` — `DEFAULT_HTTP_PORT: u16 = 9191` from `crates/gents-cli/src/main.rs:45` — i.e. the demo pair flow fell back to the *default* HTTP port for node B instead of the per-test ephemeral port, or node B wasn't listening yet when pair probed it. Everything after (`delegate failed: not paired`, `spawn_subagent tool_not_allowed`) is downstream cascade of the failed pair.

Two candidate defects:
1. The demo `pair` path resolves node B's admin endpoint to the hardcoded default 9191 in some path rather than the ephemeral port the test assigned (real bug, would only pass when 9191 happens to be free/owned by the right process), or
2. node B startup lost a readiness race under parallel-suite load and pair has no retry.

## Ask

Reproduce under load, determine whether the 9191 fallback is a genuine endpoint-resolution bug in the demo pair flow or a readiness race, and fix per flaky-tests-are-defects policy.

### LINKED ISSUE #1040: Flaky: codex_shim projects_authorized_subagent — duplicate reasoning item after reset-tail under load
# Flaky: codex_shim_projects_authorized_subagent_and_enforces_read_only_child_thread — duplicate reasoning item after reset-tail under load

## Observation

During stress verification for #990/#989/#933 (branch `fix/flaky-cli-suite`), iteration 4 of a 20× `cargo test -q -p gents-cli --test cli_codex_shim` loop failed:

```
---- codex_shim_projects_authorized_subagent_and_enforces_read_only_child_thread stdout ----
Error: terminal durable reasoning completed a duplicate item after reset-tail
test result: FAILED. 29 passed; 1 failed; 10 ignored; finished in 19.16s
```

1 failure in ~24 fixed-branch suite runs under heavy host load (a concurrent second worktree build/test loop); 0 in 20 base-commit runs on a quieter host. A load-matched 30×/30× A/B result is recorded on the PR.

## Analysis: the cursor forgets emitted text at a reset-tail boundary

`turn/stream.rs` `ReasoningCursor::observe()`:

- On a boundary completion it **clears `observed_preview`**, sets `active_item_id = None`, and bumps `segment` — correct for a genuinely new reasoning stream that follows.
- But the response row **still carries the same durable reasoning text**. If any later pre-finalize poll cycle sees `marker_changed && !projection_settled` (the progress marker spans request/response/tool/inference rows, so heartbeat-ish runtime writes qualify) with `progress_seq` unchanged, `observe()` compares the unchanged text against the now-empty preview and emits it as a **full fresh delta**: `ItemStarted` on the next segment id + a complete replay + eventually a duplicate `ItemCompleted` when the terminal block calls `finish_reasoning` (the `terminal_reasoning_backfill_allowed` gate passes because the reopened item is open).
- The existing guard at the `progress_boundary && current == previous_preview` comparison only covers the same-call case; the cleared-preview state from an *earlier* call is unguarded.

A plausible fix: have the cursor retain the completed segment's text (e.g. `completed_preview`) and treat `current == completed_preview` as no-delta, so an unchanged durable text can never re-open a segment. Projection/transcript logic is proven-adjacent territory — consider whether the reasoning-projection contract (#745 lineage, `Proofs.Conformance.Contracts.Json.CodexShim`) should encode "an unchanged durable reasoning text emits at most one completion" before the code fix, per the Lean-first flow.

## Repro

Full `cli_codex_shim` target in a loop under concurrent host load (a parallel workspace build is enough); observed at ~4% per suite run on a loaded host.

jackzampolin:
# Kill the CLI-suite flake class: full-workspace parallel-run contention

Three flaky CLI integration tests shared one class — they only failed under full-suite parallel load, when server startup is slow enough that gaps the tests never see in isolation open up. Each got a root-cause fix in product code; no test bodies changed, no timing slack added. Plus one deterministic gate breakage found and fixed along the way, and a dev-workflow improvement.

## #990 — `demo_pair_delegate_materializes_remote_worker` (candidate (a): real endpoint-resolution bug)

The demo pair/delegate flow shelled out to `p2p status` / `network create|grant` / `pairings invite|join` with only `--home`. Those subcommands resolve the admin endpoint via `resolve_graphql_endpoint`: explicit `--graphql` → the node's persisted `runtime_state.json` → **hardcoded default port 9191**. The server writes `runtime_state.json` only after its runtime reaches Ready (`serve.rs`), but `/healthz` — the only thing `pair()` gates on — is served by the DefraDB node HTTP server from node build time. Under parallel-suite load that window stretches to seconds, `p2p status --home <node>` finds no state file, probes `http://127.0.0.1:9191/api/v0/p2p/info`, and dies with connection refused. Everything downstream (`delegate failed: not paired`, `spawn_subagent tool_not_allowed`) was cascade.

**Fix** (`commands/demo/fleet.rs`): the fleet constructs both endpoints itself — pass `--graphql` explicitly on every CLI invocation in `pair()`/`delegate()`. `pairings invite` builds its token from the live HTTP API (`/node/identity`, `/p2p/shareable-address`), so no `runtime_state.json` dependency remains anywhere in the pair path. Both node A and node B were exposed (node A is also only healthz-gated before the shell accepts `pair`); both are covered.

## #933 — `cli_codex_shim` `thread/start` raw transaction conflict

`codex_shim/store.rs::query_node_json` executed reads **and auto-committed upserts** (`thread/start`'s `create_codex_thread`/`ensure_agent_session`, early rename's `upsert_AgentConversation`) via bare `EmbeddedNode::execute` — no conflict retry. Right after `AgentRuntime.process_state=ready` the runtime's startup reconciliation writes are still in flight; an auto-commit that loses that race surfaces `commit error: … transaction conflict. Please retry` raw to the Codex client. This is the same defect class already fixed on adjacent paths: startup config writes (#935/#926) and the shim's own turn-submission path (`create_agent_request_with_retry`) both have bounded retries.

**Fix**: route `query_node_json` through the centralized bounded conflict retry (`gents::retry::execute_graphql_with_conflict_retry`, #440 — 3 retries, 100 ms doubling backoff), and give `execute_committed` the same bounded retry around its begin/execute/commit cycle. Retrying is safe: a conflicted transaction aborted without committing. Error strings unchanged; exhaustion still returns the raw error. This is a production fix, not test hygiene — a user connecting Codex right after server start hit the same raw conflict.

## #989 — `codex_shim_derives_git_info_and_persists_early_rename` (same class; honest status: not individually reproduced)

The original panic was truncated (`tail -6`), so there is no captured message to match. But every operation on that test's failing surface — `thread/start` ×2, `thread/name/set` (an `upsert_AgentConversation`), `thread/read` — flows through the exact unretried path fixed for #933, and the test's error-reporting shape (`read_typed_response` bail) matches #933's captured failure. The fix covers that whole class. Linked rather than closed; leaving an honest status comment on the issue — if it recurs post-merge it is a different defect and deserves a fresh capture.

## Also: deterministic gate breakage from #1011 (found while reproducing)

`cargo test -p gents-cli` standalone has been red since yesterday: #1011 (via #1024, "keep non-production storage backends test-only") moved the `redb` feature to `gents`' dev-deps, and the codex-shim background unit test (`background.rs`) was the one `EmbeddedNode` construction site still relying on the upstream default backend — which is `Redb`, now compiled out in a standalone `-p gents-cli` graph. Every other site pins RocksDb; this one now does too. PR CI never sees this (full CLI suites run only on main pushes), which is how it landed silently.

## Dev workflow: `make worktree`

Cold worktrees cost ~40 min of recompilation (sccache can't cache build scripts, proc-macro expansion, or linking) plus hours of Mathlib. `make worktree BRANCH=<branch>` bootstraps a sibling worktree with `target/` and `proofs/.lake` cloned from the current checkout via APFS clonefile (~8 s, copy-on-write; `.lake` copy gated on a `lake-manifest.json` match). Documented in CLAUDE.md as the default.

## Verification (final, at merge head — `8b3ec81d` is `d0f364ac` + a formatting-only commit)

Full sweep at `d0f364ac`: **40/44 green**, and every failure is the single filed, pre-existing #1041 port-theft class, each self-diagnosed by this PR's capture fix:

| Gate | Result |
|---|---|
| `cli_codex_shim` ×20 | 17/20 — all 3 failures: shim port stolen pre-bind (captured: "Address already in use"), #1041 |
| `cli_demo` ×20 | 19/20 — 1 failure: node A port theft (captured), #1041 |
| **`demo_pair_delegate_materializes_remote_worker` (#990)** | **20/20** (+15/15 in the pre-commit loop) |
| **`codex_shim_interrupt_completes_with_running_background_tool` (#933)** | **19/20, zero transaction conflicts** (1 loss = captured port theft) |
| **`codex_shim_derives_git_info_and_persists_early_rename` (#989)** | **20/20** |
| `cargo test -p gents-cli` (full parallel) ×2 | 2/2 |
| `cargo test -p gents` | green |
| `cargo check --workspace --all-targets` | green |

Also: matched-load A/B (30×/30×, earlier head): branch 30/30 vs base 28/30. Review round: 5 independent reviewers + adversarial verify; all confirmed findings fixed on this branch (see review comment and `672ded8c`/`1f635ae8`/`d0f364ac`).

The residual port-theft class (test-suite ports pre-allocated at test start, stolen by the OS ephemeral allocator before bind; defra-node's HTTP task then dies silently while the process lives) is pre-existing on main (2/30 in the base A/B), fully diagnosed with captures on #1041, and needs the respawn-retry pattern across the spawn helpers plus a product-side startup self-probe — follow-up scope.

## Follow-ups (now fixed in this PR where possible)

- #1040 — duplicate reasoning replay after reset-tail: **fixed here** (`83b86c51`). Contracts check confirmed the Lean model already forbids the duplicate — pure implementation alignment. Red-first repro proven against base; 8 new cursor unit tests.
- #1041 — mid-test server death: **diagnosis capture added here** (`c00255fb`, `ServeProcess::capturing()` + 84 wrapped call sites). Issue stays open for the root-cause fix; structural candidates documented on the issue.

Closes #990
Closes #933
Closes #1040

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,4 +43,5 @@ The few things you can't discover until they bite:
 - **Gate with the full package suite** (`cargo test -p gents`), not `--lib` — integration tests are separate compile units and `--lib` will happily pass while they don't build.
 - **Compile the whole workspace before pushing** (`cargo check --workspace --all-targets`) — the test gate above still skips examples, the desktop crates, and many test targets, so a new required struct field breaks their construction sites silently until CI (or the next unrelated PR). CI enforces this in `rust-and-cli`; run it locally first.
 - **Flaky tests are defects.** The formal-verification investment exists to eliminate that class; capture, file, and fix — never shrug.
+- **Create worktrees with `make worktree BRANCH=<branch>`**, not bare `git worktree add`. It clones `target/` and `proofs/.lake` from the current checkout via APFS clonefile; a cold worktree costs ~40 min of recompilation (sccache can't cache build scripts, proc-macro expansion, or linking) and hours of Mathlib. Note `cargo check` artifacts don't warm `cargo test` — warm with `cargo build --tests -p <crate>` if tests are the goal.
 - `tracing`, never `println`.
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,16 @@ help:
 	@echo "  make live-cli              Run live CLI smoke test"
 	@echo "  make live-agent            Run ignored live runtime tests"
 	@echo "  make live-desktop-smoke    Run live desktop smoke suites"
+	@echo
+	@echo "Worktrees:"
+	@echo "  make worktree BRANCH=<branch> [DIR=<dest>] [BASE=<ref>]"
+	@echo "                             Create a worktree with target/ and proofs/.lake"
+	@echo "                             cloned from this checkout (APFS clonefile)"
+
+.PHONY: worktree
+worktree:
+	@test -n "$(BRANCH)" || { echo "usage: make worktree BRANCH=<branch> [DIR=<dest>] [BASE=<ref>]" >&2; exit 2; }
+	@scripts/worktree-bootstrap.sh "$(BRANCH)" $(DIR) $(BASE)
 
 .PHONY: build build-cli build-cli-headless build-desktop build-desktop-ui
 build:
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ help:
 .PHONY: worktree
 worktree:
 	@test -n "$(BRANCH)" || { echo "usage: make worktree BRANCH=<branch> [DIR=<dest>] [BASE=<ref>]" >&2; exit 2; }
-	@scripts/worktree-bootstrap.sh "$(BRANCH)" $(DIR) $(BASE)
+	@WORKTREE_DIR="$(DIR)" WORKTREE_BASE="$(BASE)" scripts/worktree-bootstrap.sh "$(BRANCH)"
 
 .PHONY: build build-cli build-cli-headless build-desktop build-desktop-ui
 build:
--- a/crates/gents-cli/src/commands/codex_shim/store.rs
+++ b/crates/gents-cli/src/commands/codex_shim/store.rs
@@ -1,13 +1,21 @@
 use anyhow::Result;
 use gents::defra_node::{EmbeddedNode, QueryRequest};
+use gents::retry::{
+    defradb_conflict_retry_backoff, execute_graphql_with_conflict_retry,
+    is_defradb_transaction_conflict_text, DEFRA_DB_CONFLICT_MAX_RETRIES,
+};
 use gents_protocol::transcript::present_persisted_message;
 use serde_json::{json, Value};
 
 use super::progress::response_field_is_blank;
 use crate::materialized_message_query;
 
+/// All shim reads and auto-committed writes go through the runtime's bounded
+/// DefraDB conflict retry (#440): right after startup the runtime's own
+/// reconciliation writes are still in flight, and an unretried auto-commit
+/// surfaces a raw `transaction conflict` to the Codex client (#933).
 pub(super) async fn query_node_json(node: &EmbeddedNode, query: &str) -> Result<Value> {
-    let response = node.execute(query).await;
+    let response = execute_graphql_with_conflict_retry(node, query, "codex shim store").await;
     if response.has_errors() {
         anyhow::bail!("GENTS Codex shim query failed: {:?}", response.errors);
     }
@@ -24,6 +32,31 @@ pub(super) async fn query_node_json(node: &EmbeddedNode, query: &str) -> Result<
 /// `config skill` CLI path (#340). Mirrors the Local arm of
 /// `config_writes::txn::ConfigApplyTxn`.
 pub(super) async fn execute_committed(node: &EmbeddedNode, mutation: &str) -> Result<Value> {
+    let mut retry_index = 0;
+    loop {
+        match execute_committed_once(node, mutation).await {
+            Ok(value) => return Ok(value),
+            Err(error)
+                if retry_index < DEFRA_DB_CONFLICT_MAX_RETRIES
+                    && is_defradb_transaction_conflict_text(&format!("{error:#}")) =>
+            {
+                let backoff = defradb_conflict_retry_backoff(retry_index);
+                retry_index += 1;
+                tracing::warn!(
+                    retry_count = retry_index,
+                    max_retries = DEFRA_DB_CONFLICT_MAX_RETRIES,
+                    backoff_ms = backoff.as_millis() as u64,
+                    error = %error,
+                    "retrying Codex shim committed mutation after transaction conflict"
+                );
+                tokio::time::sleep(backoff).await;
+            }
+            Err(error) => return Err(error),
+        }
+    }
+}
+
+async fn execute_committed_once(node: &EmbeddedNode, mutation: &str) -> Result<Value> {
     let handle = node
         .runner()
         .begin_txn(false)
--- a/crates/gents-cli/src/commands/codex_shim/store.rs
+++ b/crates/gents-cli/src/commands/codex_shim/store.rs
@@ -30,7 +30,9 @@ pub(super) async fn query_node_json(node: &EmbeddedNode, query: &str) -> Result<
 /// COMMIT does. Routing skill enable/disable writes through here lets a running
 /// agent pick up Codex-driven toggles without a restart — matching the
 /// `config skill` CLI path (#340). Mirrors the Local arm of
-/// `config_writes::txn::ConfigApplyTxn`.
+/// `gents::config_client::txn::ConfigApplyTxn`, plus a bounded conflict retry
+/// that arm does not have (#933): a conflicted cycle commits nothing, so a
+/// successful call still emits exactly one Update event.
 pub(super) async fn execute_committed(node: &EmbeddedNode, mutation: &str) -> Result<Value> {
     let mut retry_index = 0;
     loop {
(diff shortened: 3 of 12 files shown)


jackzampolin:
**Matched-load A/B result (30×/30×, simultaneous start, same host):** branch **30/30 PASS**, base (main @ b5105682) **28/30** — base's two failures are the #1041 ws-connect-refused family. This settles the open question in the PR body: the two newly-surfaced flakes (#1040, #1041) are pre-existing on main and not amplified by this branch's retry change; the uniform store retry stays as-is.

Cumulative `cli_codex_shim` full-target runs on this branch: **53/55 green**, with the three targeted tests (#933 interrupt, #989 git-info/early-rename, plus the rest of the suite's prior captures) green in **55/55**. The two branch-side failures were single occurrences of #1040 and #1041 under doubled load, before the matched A/B.

Fixes for #1040 and #1041 are being added to this PR next (design fan-out in progress); the full verification sweep (20× demo loop, 2× full-parallel suite, `cargo test -p gents`, workspace check) runs once over the final tree and lands here after.

jackzampolin:
### Code review

Found 1 issue:

1. `make worktree` with `BASE=` but no `DIR=` silently creates the worktree in the wrong place from the wrong base (bug due to unquoted optional positionals in `Makefile`: `@scripts/worktree-bootstrap.sh "$(BRANCH)" $(DIR) $(BASE)` — an empty `$(DIR)` vanishes, so `BASE` shifts into the script's dest-dir parameter `$2`; `make worktree BRANCH=x BASE=origin/main` creates a worktree at a path literally named `origin/main`, branched from `HEAD`, with no error)

https://github.com/source-inc/gents/blob/3e8dcb935841d1e25d5765d466b3370fd40b1452/Makefile#L69-L71

Also verified during this review (below the report threshold, being fixed in a follow-up commit on this branch): two narrow edge cases in the new `ReasoningCursor` stale-tail logic (boundary-read seq recording; rolled-tail growth detection) and `capturing()` collapsing a log-read failure into empty output.

Issue #1032, opened 2026-08-05
Harbor runner loses failure evidence when the per-trial Gents server exits


jackzampolin:
## Problem

The Harbor integration does not supervise the per-trial Gents server after startup and deletes the runtime home before preserving enough evidence to diagnose an unexpected server exit.

Observed in the August 4 Terminal-Bench 2.1 C32 run:

- job: `gents-d4f-max-988-36a2a038-full-c32-clean-v2`
- Gents SHA: `36a2a0381a06b5436fa84bef8c86dd6aebf36d7f`
- trial: `db-wal-recovery__ZyXBWLs`
- agent execution ran from `2026-08-05T00:53:59Z` until `2026-08-05T06:13:47Z`
- output then contained `Terminated`
- `response wait` retried `http://127.0.0.1:9191/api/v0/graphql` four times and failed
- Harbor recorded a generic adapter `RuntimeError` and did not run the verifier
- no ATIF trajectory was emitted
- the container and `GENTS_HOME` were removed, so the server exit status, final process tree, partial persisted timeline, and store state were unavailable

It remains unknown whether the task sent a broad signal, the Gents runtime crashed, or Harbor/container lifecycle terminated it. The current adapter destroys the evidence needed to distinguish those cases.

## Current lifecycle gap

`scripts/harbor/run_gents.sh` checks `server_pid` during startup only. During the benchmark request it blocks in:

```sh
gents response wait ...
```

There is no race or supervisor linking the response waiter to the server process. When the server exits, the visible failure becomes a secondary GraphQL connection error.

In `scripts/harbor/gents_agent.py`, temporary runtime state including `GENTS_HOME` is removed before `_require_success` classifies the failed runner. The cleanup trap can project a partial trace only while the server is still alive.

## Desired behavior

Treat the Gents server and response waiter as one supervised trial lifecycle. Unexpected server exit should terminate the wait promptly and produce a first-class diagnostic bundle before destructive cleanup.

## Acceptance criteria

- After startup, the runner continuously supervises both the Gents server PID and response waiter.
- Unexpected server exit cancels the waiter promptly and records exit code versus terminating signal.
- Before deleting `GENTS_HOME`, failure handling captures:
  - server log tail;
  - Gents server exit status;
  - request and response documents if available;
  - partial ATIF/timeline projection when possible;
  - `gents status` or a clear note that GraphQL is unavailable;
  - a process-tree snapshot;
[...]

Issue #1031, opened 2026-08-05
Managed exec hangs after the direct child exits when a descendant retains output pipes


jackzampolin:
## Problem

A foreground managed command can remain active forever after its direct child has already exited when a daemonized descendant inherits the command's stdout or stderr pipe.

Observed in the August 4 Terminal-Bench 2.1 C32 run:

- job: `gents-d4f-max-988-36a2a038-full-c32-clean-v2`
- Gents SHA: `36a2a0381a06b5436fa84bef8c86dd6aebf36d7f`
- trial: `install-windows-3.11__dKrFwF4`
- the agent launched nginx/websockify as background services
- the direct foreground shell PID later disappeared
- nginx remained alive as a daemon and retained inherited output handles
- the Gents response waiter and native `bash_unrestricted` tool call remained active for more than 12 hours without inference

The same defect remains present at PR #1030 head `b96ae809f1cc314b97fa95c0d5e32e26e1d784a7`.

## Root cause

On Unix, the normal child-exit branch in `crates/gents/src/managed_exec/process.rs` leaves the command deadline/cancellation `select!` and then waits without a bound:

```rust
let stdout = join_capture(stdout_task).await;
let stderr = join_capture(stderr_task).await;
```

Those readers wait for EOF. A descendant retaining either pipe prevents EOF indefinitely. The capture-drain timeout is currently used only after cancellation or timeout when process-group reap fails, so the command watchdog cannot recover once the direct child has exited.

This is distinct from #1018: that issue covers a direct child that itself remains blocked. Here the direct child is already gone.

## Desired behavior

Direct-child exit must make the foreground tool terminal within a small bounded pipe-drain interval, even if descendants retain inherited handles. The tool should preserve available output and the direct child's exit status while making incomplete capture observable.

## Acceptance criteria

- The normal child-exit path applies a bounded drain to stdout and stderr capture.
- If EOF does not arrive, capture tasks are aborted or detached without keeping the tool call active.
- The outcome records that output capture was incomplete or forcibly bounded.
- The direct child's exit code remains authoritative.
- The active-exec registry entry is removed promptly.
- A regression test starts a background descendant that inherits stdout and stderr, lets the shell exit successfully, and proves `run_managed_exec` returns within the drain bound.
- The test covers both stdout and stderr retention.
- Any descendant-cleanup policy remains consistent with foreground versus explicit `spawn_process` semantics.

## References

- `crates/gents/src/managed_exec/process.rs`
[...]

Issue #719, opened 2026-07-15
Make compaction budgeting coherent and expose validated behavior-level tuning


jackzampolin:
## Context

The runtime has configurable compaction strategy and threshold, plus hardcoded CompactionOptions values:

- keep_recent_tokens = 20,000
- tool_result_max_chars = 2,000

Before exposing those values, the end-to-end compaction budget and strategy semantics need to be coherent.

### The outer and inner gates count different inputs

The daemon triggers compaction from a BuiltPrompt estimate plus request text. The compactor's internal needs_compaction check receives only history and compares it with the full context-window threshold.

Preamble, tool definitions, existing summaries, selected skills, request context, current request, and output reserve therefore do not consistently influence the compactor's target.

### Loaded history is stripped before strategy selection

agent/daemon/request.rs unconditionally calls strip_tool_results before invoking the configured compactor.

As a result:

- Summarize never sees the original bounded tool-result observations.
- StripThenSummarize strips already-created stubs a second time.
- tool_result_max_chars is effectively inert in the only production compact call path.

### There is no final fit postcondition

After compaction the prompt is rebuilt, but the runtime does not require the complete rebuilt provider input to fit the effective budget before inference proceeds.

StripToolResults can also return without reducing message count or persisting a compacted prefix, leaving a permanently over-budget session.

## Dependency

Depends on #718 for a shared provider-input TokenCounter. This issue defines how that counter is used.

### Phase 1: Lean budget and strategy contract

Model an abstract provider-input cost with:

- fixed prompt layers
- existing compaction summaries
- selected skill reminders
- tool definitions
- request-context message and current request
- output reserve and configured safety threshold
- conversation-history budget as the remainder

Define the legal result contract:

- compaction returns provider-valid history whose complete rebuilt input fits the effective budget; or
- compaction returns an explicit cannot-fit/failure outcome

Prove that no strategy mutates canonical AgentMessage history, prompt layer order is preserved, and pair-blind recent-window splits sanitize safely. Zero sorrys.

### Phase 2: Make strategy semantics real

Remove the unconditional pre-strip from request preparation or explicitly move it into the strategy implementation.

Pin the intended behavior:

- StripToolResults: replace tool results with stubs; succeed only if the complete rebuilt input then fits.
- Summarize: split unstripped-but-already-bounded history, pre-truncate tool results only for summary inference, summarize the old window, and retain the recent window.
[...]

PR #1046, opened 2026-08-05
fix(runtime): harden Terminal-Bench compaction recovery

jackzampolin:
## What changed

- bound managed foreground-process capture draining so a noisy command cannot wedge Terminal-Bench execution after exit
- retain useful request and server diagnostics when the Harbor agent loses its server process
- recover from repeated guided structured-output failures with one strict non-guided JSON fallback
- raise the independent compaction summary output budget from 4,096 to the existing validated 32,768-token ceiling
- preserve the parent reasoning profile for guided compaction and its fallback, with explicit regression coverage

## Why

The targeted Terminal-Bench 2.1 regression exposed both harness hangs and a dominant compaction failure mode on DeepSeek V4 Flash. In the prior six-task run, four trials terminated as compaction-provider infrastructure errors. The model was inheriting max reasoning while the internal summary completion had only 4,096 output tokens, producing EOF-truncated JSON or no usable visible output.

The runtime changes keep compaction bounded and diagnostic while giving the inherited reasoning profile enough output room to produce its structured continuation checkpoint.

## Validation

- `git diff --check`
- focused guided/fallback and independent-output-budget compaction tests
- `CARGO_INCREMENTAL=0 cargo test -p gents compaction::tests` (63 passed, 1 ignored)
- `CARGO_INCREMENTAL=0 cargo test -p gents`
- exact Linux/amd64 release build at `fa899256` on studio-2

## Benchmark status

A fresh six-task, c=6 targeted regression is currently running on studio-2 against workstation-1 DSV4F as `gents-d4f-fa899256-compaction32k-targeted-c6-v2`. All six tasks have entered inference. This PR stays draft until the targeted compaction-heavy run is evaluated.

Tracks #1031 and #1032. Related configurability follow-up: #719.

### DIFF
--- a/crates/gents/src/agent/loop_stream.rs
+++ b/crates/gents/src/agent/loop_stream.rs
@@ -101,10 +101,26 @@ impl StructuredOutputConfig {
             validate: Arc::new(|raw| {
                 serde_json::from_str::<T>(raw)
                     .map(|_| ())
-                    .map_err(|error| error.to_string())
+                    .map_err(|error| {
+                        format!(
+                            "{error}; raw_output_preview={}; finish_metadata=unavailable_at_rig_streaming_boundary",
+                            bounded_structured_output_preview(raw)
+                        )
+                    })
             }),
         }
     }
+}
+
+fn bounded_structured_output_preview(raw: &str) -> String {
+    const MAX_PREVIEW_BYTES: usize = 192;
+    let mut cut = raw.len().min(MAX_PREVIEW_BYTES);
+    while !raw.is_char_boundary(cut) {
+        cut -= 1;
+    }
+    let suffix = if cut < raw.len() { "…" } else { "" };
+    serde_json::to_string(&format!("{}{suffix}", &raw[..cut]))
+        .unwrap_or_else(|_| "\"<unavailable>\"".to_string())
 }
 
 #[derive(Debug)]
@@ -694,7 +710,9 @@ where
                     MidStreamDirective::Fail { reason } => {
                         Err(StreamingError::Completion(
                             CompletionError::ProviderError(format!(
-                                "completion produced no visible output: {reason}"
+                                "completion produced no visible output: {reason}; \
+                                 raw_output_preview=\"\"; \
+                                 finish_metadata=unavailable_at_rig_streaming_boundary"
                             )),
                         ))?;
                         unreachable!("Err(..)? above ends the stream");
--- a/crates/gents/src/compaction.rs
+++ b/crates/gents/src/compaction.rs
@@ -11,8 +11,8 @@ mod tests;
 
 use history::{extract_file_activity, pretruncate_tool_results, split_messages_for_summary};
 use summary::{
-    bounded_error_diagnostic, compaction_prompt, compaction_request_prompt, format_summary,
-    ContinuationCheckpoint,
+    bounded_error_diagnostic, compaction_json_fallback_prompt, compaction_prompt,
+    compaction_request_prompt, format_summary, parse_fallback_checkpoint, ContinuationCheckpoint,
 };
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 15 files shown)


jackzampolin:
Exact-harness compaction regression (`fa899256`, linux/amd64 SHA256 `ead97f1f1e1443a9f8458d91611ee292c9f0c2629eb538e2dc2f2347430aa95b`) completed cleanly on studio-2 against workstation-1 DSV4F (`reasoning=max`).

- Harbor 0.20.0, exact checked-out `scripts.harbor.gents_agent:GentsAgent` / `run_gents.sh`
- six historically long/compaction-heavy Terminal-Bench 2.1 tasks at c=6
- 6/6 terminal `result.json` files with `exception_info=null`
- 3 verifier passes, 3 ordinary verifier misses
- `install-windows-3.11`: 250 main inference turns, 7 persisted `call_kind=compaction` calls, all 7 completed, no failure reasons; exact harness verifier-scored MaxTurn instead of raising Harbor `RuntimeError`
- `make-doom-for-mips`: MaxTurn also verifier-scored with no exception
- zero compaction/provider/infrastructure/harness failures

The optional conversation-title one-off still emits a harmless `no visible output` warning because it inherits max reasoning under a 24-token title cap, then uses the deterministic fallback; no main inference or compaction call emitted that error.

The authoritative 89-task full Terminal-Bench 2.1 run is now live at c=32: `gents-d4f-fa899256-compaction32k-exact-harness-full-c32-v1`.

jackzampolin:
## Authoritative Terminal-Bench 2.1 result (`fa899256`)

The exact pinned c32 run completed cleanly on `studio-2`, using workstation-1's DSV4F endpoint and Harbor 0.20.0.

- Job: `/Users/admin/harbor-jobs/gents-d4f-fa899256-compaction32k-exact-harness-full-c32-v1`
- Commit: `fa89925691847a7756b24b93976209760e7efce3`
- Binary SHA-256: `ead97f1f1e1443a9f8458d91611ee292c9f0c2629eb538e2dc2f2347430aa95b`
- Config SHA-256: `ce79ec7d386bf41ddf5b50aa8b3856545a3847fa069926fdae5020fb52e12ebe`
- Settings: c=32, `reasoning_effort=max`, context 458,752, turn max-output 393,216, max 250 turns
- Result: **69/89 = 77.53%**
- Harbor: 89/89 completed, **0 errored**, 0 retries
- Gents/Harbor exceptions: **0**
- Outcomes: 81 normal completions, 8 normal max-turn outcomes (one of those still verifier-passed)
- Wall time: 4h45m31s

The observed score is 4.47 pp below the ~82% expectation, but the misses are model/verifier outcomes rather than harness attrition: seven verifier misses exhausted 250 turns and thirteen completed normally before failing verification. For comparison, the previous best full c32 attempt in this environment was 59/89 with four exceptions; this run gained ten passes and eliminated all exceptions. An 82% underlying pass rate remains within the single-run 95% binomial interval for 69/89.

### Compaction and recovery evidence

- No `compaction_provider_failure`, `CompactionProviderError`, exhausted guided-compaction fallback, `compaction failed`, or owned-loop failure appears in any of the 89 task logs.
- The longest task, `regex-chess`, ran 4h45m, completed normally, and verifier-passed after **15/15 observed compaction calls completed**.
- Other directly audited long regressions also compacted repeatedly: `path-tracing` 19, `gpt2-codegolf` 14, `install-windows-3.11` 7, and the exact six-task regression completed with zero exceptions or compaction failures.
- DSV4F did still emit unusable output occasionally. The runtime retracted and resampled 20 attempts: 19 in turn-0/internal calls and one ordinary main call at turn 227. All recovered. No retry ladder was exhausted.
- The literal `completion produced no visible output` terminal diagnostic appears only inside optional conversation-title one-off warnings (90 lines across 88 tasks). Every affected title used the deterministic fallback; none was a main-loop or terminal compaction error.
- Five transient first-stream HTTP failures across five tasks were retried successfully. They produced no trial exception.

### Capacity

Effective provider decode concurrency peaked at **31/32**, with the admission waiting gauge normally at zero. This is materially better batching than the earlier runs and did not create terminal queue failures. Optional title one-offs sometimes encountered `QueueFull` at startup and fell back deterministically; main and compaction work continued.

All ten GitHub checks on this exact head SHA are green. General compaction configurability remains tracked separately in #719.

Issue #1051, opened 2026-08-06
Workspace-root allowlist vs operator --tool-root ceiling: admitted personas can be unusable at serve time


noahsaso:
Live repro (persona e2e): operator registers WorkspaceRoot rows for two repos; the directory catalog publishes both in `allowed_roots`; a phone composes a persona with root B; admission validates root ∈ allowed_roots; the reconciler applies it. Then the first request on that persona fails at tool-surface resolution:

```
behavior <did>:fast file tool root /Users/noah/Developer/source/gents-mobile escapes operator tool root /Users/noah/Developer/source/defra-agent
```

because `gents server --tool-root` (defaulting to the server's CWD) is a separate, older confinement mechanism that the persona pipeline never consults. The whole point of the request channel was "validated picks can't produce broken behaviors" — this seam reintroduces exactly that: admitted-yet-unusable.

Asks, in increasing strength:
1. `gents config workspace-root set` warns (or refuses without --force) when the path is outside the running server's operator tool-root.
2. The catalog projection publishes only roots within the operator ceiling (never offer a pick the runtime will refuse) — needs the projection to know the effective tool-root.
3. Persona admission (`decide_persona_request`) gains an operator-ceiling conjunct so the request is rejected with a clear status_detail instead of a serve-time error.
4. Longer term: consider deriving the operator ceiling FROM the registered workspace roots (their common ancestor or the set itself) so there is one source of truth.

Workaround used: restart the server with `--tool-root` at the roots' common parent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Arv3YzUJH1oSvYL21kekca

jackzampolin:
Fixed by #1055 (merged 2026-08-10): one shared predicate (`directory_projection::filter_roots_to_ceiling`, mirroring the serve-time guard's exact semantics, fail-closed on unresolvable paths) applied at both the directory projection's `allowed_roots` publication and the persona reconciler's catalog loader — clients are never offered a root the runtime would refuse, and admitted personas can no longer be unusable at serve time.

PR #1055, opened 2026-08-06
fix(personas): ceiling-filter workspace roots at catalog + admission (#1051)

noahsaso:
Closes the admitted-yet-unusable seam from #1051: a persona could pick a workspace root the catalog published and admission accepted, then fail at serve time with `escapes operator tool root` because `--tool-root` is a separate confinement mechanism the persona pipeline never consulted.

## Fix
One shared predicate — `directory_projection::filter_roots_to_ceiling` — mirroring the serve-time guard's exact semantics (`resolve_configured_tool_root` + component-wise `starts_with`, fail-closed on unresolvable paths), applied at both loaders:
- The directory projection's `allowed_roots` publication: clients are never offered a root the runtime would refuse.
- The persona reconciler's `PersonaCatalogView`: the existing `root ∈ allowed_roots` admission conjunct now enforces the ceiling for free, rejecting with the standard named-value `status_detail`.

Ceiling threads from `DocumentResolveContext.tool_ceiling.root()` at both background-task spawn sites; no ceiling (`None`) behaves exactly as before. No schema changes, no Lean changes (the conjunct set is unchanged — only the set it validates against is now consistent with the runtime).

Deliberately out of scope (noted in #1051): the CLI set-time warning (needs cross-process ceiling knowledge — likely wants the effective ceiling published on the runtime row) and deriving the ceiling from registered roots.

## Tests
Filter matrix incl. the `/ceil/ws` vs `/ceil/wsx` sibling-prefix trap (component-wise containment); embedded-node test: ceiling-aware store excludes the out-of-ceiling root from the view and admission rejects a create naming it. Full suite 1681 passed / 0 non-environmental failures; workspace all-targets check clean; fmt clean.

### DIFF
--- a/crates/gents/src/agent/directory_projection.rs
+++ b/crates/gents/src/agent/directory_projection.rs
@@ -363,9 +363,10 @@ pub async fn reconcile_directory_tick(
 pub async fn run_directory_projection(
     node: Arc<EmbeddedNode>,
     source_did: String,
+    ceiling_root: Option<std::path::PathBuf>,
     cancel: CancellationToken,
 ) -> Result<()> {
-    let store = GraphqlDirectoryStore::new(node.clone());
+    let store = GraphqlDirectoryStore::new(node.clone(), ceiling_root);
     let mut subscription = node.subscribe(&[EventName::Update]);
     let mut interval =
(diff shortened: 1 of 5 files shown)


jackzampolin:
conflicts!

Issue #840, opened 2026-07-20
Persist rendered provider inputs durably by default


jackzampolin:
From the gents.md doc audit (2026-07-20): the flywheel's "training example by construction" claim needs rendered inputs durable, not just capturable.

Today \`rendered_request_capture_factory\` defaults to \`None\` (\`crates/defra-agent/src/agent.rs:225\`); the inference loop only invokes capture when externally installed (\`agent/daemon/inference.rs\`). There is a well-defined DTO in \`rendered_request.rs\` but no collection or standard persistence sink.

Scope: default-on capture sink → durable collection (encrypted to session participants), surfaced in the run timeline and adapter projections. There is exactly one assembly point (PromptAssembly-fenced), so capture placement is already settled.

Issue #722, opened 2026-07-15
Make trimmed tool results recoverable through the session-history tool


jackzampolin:
## Context

Defra-agent already has the right storage architecture for reversible context reduction:

- canonical `AgentMessage` history remains in DefraDB
- every tool call is persisted as an `AgentToolCall`
- oversized tool output is bounded for the model and the full output is spilled to `AgentToolResult`
- `strip_tool_results` replaces historical observations with compact stubs instead of mutating canonical history
- the opt-in `sessions` tool already lists this principal's persisted sessions

Today the resulting stub says, for example:

```text
[tool: read, call_id: call-1, 5000 bytes — see DefraDB AgentToolCall for full output]
```

That is useful to an operator, but it is not executable by the model. The `sessions` tool only supports `action: "list"`, and the background-only `read_tool_output` surface is scoped to calls spawned by the current parent request rather than arbitrary persisted history.

This is the DefraDB-native version of the spill-pointer idea noted in #395: the prompt is a bounded working set, while the database remains the lossless source of truth.

## Problem

Tool-result trimming is durable but model-irreversible:

1. A later turn can see that a historical result existed but cannot retrieve it.
2. If the original result exceeded the first-level tool-output limit, `AgentToolCall.result` contains only the bounded observation; the full bytes live in `AgentToolResult.output_text`.
3. The call-to-spill relationship is encoded only indirectly in model-facing prose (`[Full output: DefraDB doc ...]`), not as a structured database relationship suitable for a resolver.
4. Compaction therefore makes the summary the model's only practical memory even though the canonical evidence is still present locally and may be replicated.

The runtime should let an agent query its own persisted evidence without exposing a generic raw-document escape hatch.

### 1. Extend the existing `sessions` tool

Keep `action: "list"` unchanged and add a bounded read action, named `read_tool_result` unless implementation constraints favor an equivalent name.

The model-facing input should identify a persisted call through a stable runtime-owned reference. Candidate shape:

```json
{
  "action": "read_tool_result",
  "session_id": "...",
  "tool_call_id": "...",
  "offset_bytes": 0,
  "max_bytes": 16384
}
```

Using a single opaque `tool_result_ref` is also acceptable if the runtime creates and validates it. Do not require the model to construct GraphQL or query an unrestricted DefraDB document id.

The result envelope should include at least:

- session id, request id, tool call id, and tool name
- lifecycle/status and whether the result was ever delivered to the model
- returned byte range and total byte count
[...]

Issue #523, opened 2026-06-17
Rendered-request projection: materialized-view reconstruction via time-travel reads


jackzampolin:
## Goal

Materializing a past request should **automatically reconstruct the rendered provider request** — the exact bytes that were sent to the model — by reading each contributing input document **at the version that was live when the model call was made**, with no hot-path capture. This is the materialized-view shape: you ask the projection for request `R`, and it pulls the right historical versions of behavior/prompt/tool/model config and reassembles the wire request behind the scenes.

This is the feature behind the #503 reframing (fidelity as a *provable projection* of durable state, not an in-memory capture). It is **not** built yet.

## What's already settled

- **Time-travel reads work through the pinned `node.execute` passthrough.** Proven by #522 / closes #519: `Collection(cid: ["<cid>"])`, `_version { cid }`, and `_commits` all resolve through the same GraphQL surface the runtime uses — no new `defra_node` API, no upstream defradb.rs ask. The spike gives a regression fence on that primitive.
- **Leak set is scoped** (from the [#503 comment](https://github.com/sourcenetwork/defra-agent/pull/503#issuecomment-4730934767)): the only things that genuinely cannot be reconstructed from config are the **provider wire-shape** and the **MCP-resolved tool schemas**; `ctx.now` is already durable. Everything else (system prompt, sampling defaults, tool selection) is reconstructable via time-travel.

## What's NOT settled — the crux

The spike proved we can read *a* document as of *a* known CID. It did **not** answer #519 step 2: **given a past request, which commit CID of each contributing input document was live at call time?** Two strategies, and this issue should pick one (Lean first — see below):

- **(A) Stamp source CIDs at render time.** When the owned loop assembles the completion request, record the composite-commit CID of each contributing document onto `InferenceCall`. Projection reads exactly those CIDs. Exact and deterministic; costs a few durable fields and a write-time read of `_version`.
- **(B) Time-correlate.** Derive "commit at-or-before the request's render point" from `ctx.now` (already durable) against each document's `_commits` DAG. No new writes, but needs a well-defined at-or-before rule, assumes a usable ordering across the commit DAG, and needs a tie-break — the spike did not exercise this lookup.

#519's framing leaned toward (B)-as-free, but the at-or-before *lookup* is unproven; (A) is strictly sufficient on its own. Decide here.

## Work

1. **Decide (A) vs (B)** for request→commit resolution; if (B), define and test the at-or-before lookup rule.
[...]

PR #1059, opened 2026-08-07
Durable rendered-request fact records, fenced persist-before-send

jackzampolin:
Durable fact records for provider calls: every completion body the runtime sends is persisted as a `RenderedRequest` document **before** the send, and the ordering is fenced in Lean.

Related: #840 (durable rendered inputs), #523 (verified reconstruction), #722 (context resolution). **This PR closes none of them automatically** — see "What is NOT delivered".

---

### 1. The Lean fence — persist-before-send

`crates/gents/proofs/Proofs/RenderedCapture.lean` models capture as the only legal predecessor of a provider send. Seven theorems, **zero `sorry`**:

| theorem | claim |
|---|---|
| `sent_implies_durably_captured` | reaching `sent` forces the store to bind this key to this request |
| `sent_requires_a_capture_step` | stronger — every `assembled → sent` trace factors through an actual capture step, so a fail-open sink cannot produce one |
| `capture_key_determines_request` | one key binds at most one canonical request, for the life of the store |
| `capture_idempotent` | identical redelivery succeeds with no write |
| `capture_rejects_rebinding` | conflicting redelivery is an integrity error and leaves the store untouched |
| `capture_failure_blocks_send` | a rejected capture makes `sent` permanently unreachable |
| `attempt_distinguishes_facts` | a repair retry is a separate fact |

**Axioms are clean.** Verified independently, not taken from the commit message — `#print axioms` on all seven under the pinned toolchain:

```
'RenderedCapture.sent_implies_durably_captured' depends on axioms: [propext]
'RenderedCapture.sent_requires_a_capture_step' depends on axioms: [propext]
'RenderedCapture.capture_key_determines_request' depends on axioms: [propext]
'RenderedCapture.capture_idempotent' depends on axioms: [propext]
'RenderedCapture.capture_rejects_rebinding' depends on axioms: [propext]
'RenderedCapture.capture_failure_blocks_send' depends on axioms: [propext]
'RenderedCapture.attempt_distinguishes_facts' does not depend on any axioms
```

No `sorryAx`, no `Classical.choice`.

`CanonicalRequest` is **opaque** — equality is the only operation, so `capture_key_determines_request` is stated over the request *value*, not over a digest. A modeled hash column would have added an unmodeled "the writer computed it honestly" assumption to a theorem that does not need one.

The capture key is a five-component tuple with componentwise equality, not a delimited string: `(agent_did, session_id, request_scope, turn_index, attempt)`, where `request_scope` is the injective JSON pair `[request_id, capture_scope]`.

Two scope limits are **declared as emitted boundaries rather than assumed away**: `boundary.rendered-capture.assembled-request-artifact` and `boundary.rendered-capture.key-encoding-injectivity`.

### 2. The `RenderedRequest` fact record

New collection (`crates/gents-schemas/schemas/agent/rendered_request.graphql`), registered as a `DEFAULT_BASELINE` entry with an authored pin — not via `AddCollection`, which is mutually exclusive with baseline under `baseline_ensure.rs`.

**There is deliberately no `request_hash` column.** A stored digest is self-attested: the same writer chooses both the bytes and the digest over them, so it proves nothing an auditor could not have been lied to about. Integrity comes from the DefraDB field commit CID for `request_json`, computed over the block that was actually stored and replicated with it. The idempotency test asserts against `_commits`, reading back `fieldName` in Rust — `_commits` evaluates its `fieldName` filter in memory with `unwrap_or(true)`, so a malformed filter silently degrades to no filter.

### 3. Transport-seam capture

Capture happens at the **transport seam**, as the innermost HTTP client, not pre-transport. This is the architectural point of the slice: the ChatGPT-Codex transport rewrites the body *after* rig serializes it (hoists system text into `instructions` and strips it from `input`, sets `store:false`/`stream:true`, deletes `max_output_tokens`/`temperature`/`top_p`, forces `strict:false` on every tool), and Grok injects `store:false`. Capturing pre-transport would make the equality claim false for two of four provider kinds.

The transport contract is **positive, not permissive**: inside a capture scope, a completion body with nothing armed and nothing previously claimed is **refused outright**. `claim_pending()` returns `Armed` / `Resend` / `Unexplained`. `Resend` exists because rig's SSE source reconnects by re-posting — it re-captures under the same coordinates, idempotent for identical bytes and an integrity error for changed ones.

`the_captured_row_is_the_body_codex_rewrote_not_the_one_rig_serialized` and `the_captured_row_carries_the_store_false_grok_injects` compose the real provider wrappers over the real capture client. Both wrappers are generic over their inner client, so hoisting capture above them compiles and streams fine while persisting a request the backend never receives — the ordering the whole seam placement exists for is now fenced.

### 4. The widened sink

`completion_factory::loop_config` previously hardcoded `on_rendered_request: None`, leaving the compaction summarizer, title generation, and the one-shot runner uncaptured. All provider call sites now capture, under per-kind scope labels: `inference.N`, `compaction.N`, `compaction_fallback.N`, `title.N`, `oneshot.N`. Each loop starts its own turn/attempt counters at zero, which is why `capture_scope` is a key component.

The scope spans the **whole `handle_request` body**, not just the inference block. `BehaviorDaemon::handle_request` compacts ~90 lines before `run_inference`, and `StripThenSummarize` is the default strategy — so on the default path, for any session over threshold, the pre-request summarizer was issuing real provider calls (whose input is the *entire pre-truncated transcript*) under no ambient scope: zero rows, zero diagnostics. `model_backed_compaction_is_captured_like_every_other_provider_call` drives a real summarizer through the daemon and asserts, measured at the mock backend rather than at the loop, that every posted body has exactly one row carrying exactly those bytes. It reports 3 bodies against 2 rows on the old scope placement.

### 5. `AssemblyTrace` — the four unrecoverable assembly inputs

`provenance_json` carries a versioned manifest whose `AssemblyTrace` records the four things that are genuinely **not** reconstructible from the durable transcript:

1. **The provider assistant `message_id`** — stamped into in-memory history, persisted as `id: None`.
2. **The exact threaded tool-result text per call id** — the loop threads `truncate_text(model_facing_text, tool_result_truncation_mode(name), …)` while persistence re-derives from `AgentToolCall.result` with `TruncationMode::Head`, different limits, and `model_observation_for_tool_result`.
3. **The full effective post-compaction message list** — per-turn compaction is a *sticky* mutation (`*history = compacted`), so one turn's summary governs every later turn.
4. **The `build_path` discriminator** — repair calls `build_request` directly and never applies the output clamp, so a repaired attempt carries raw `max_tokens` while the original carries the clamped one.

The clamp *value* is deliberately not stored: it is a pure function of the assembled request plus durable config, and `completion_request_input_estimate` does not read `max_tokens`, so one pass reproduces it exactly.

---

### Participant ACP authorization — blocked on [sourcenetwork/defradb.rs#1318](https://github.com/sourcenetwork/defradb.rs/issues/1318)

**No `@policy` appears in any shipped SDL.** `EmbeddedNode` exposes no policy or relationship API, `create_document_acp` is `pub(crate)`, and the constructed `LocalDocumentACP` is never retained on the node. Worse than merely unavailable: **a `@policy` id that is never installed deploys cleanly and leaves rows anonymously readable** — unregistered documents are public. Shipping a `@policy` directive here would have been an authorization claim with no enforcement behind it.

### At-rest encryption

**No `encrypt: true`, no `encryptFields`.** These protect the replicated CRDT block deltas **only**. The local node writes plaintext into the datastore and builds its indexes from it, so on a single-node install the payload is readable by anyone who can query the node. The only local mechanism is `NodeBuilder::with_at_rest_encryption_key`, which gents does not use. A green `encrypt: true` test would have been confidentiality theater.

### Therefore: **#840's participant-boundary criterion is NOT met by this PR.**

Stated plainly, because capture is **on by default**: `request_json` plus `provenance_json` — the whole conversation, the system prompt, and the tool surface — are persisted in plaintext and are anonymously readable on the node. The SDL, the sink, and the final commit all say so; it belongs here too.

**Replication status, precisely — this is three different sets and only two exclude it:**

| set | `RenderedRequest` | where |
|---|---|---|
| `BRANCHABLE_COLLECTION_NAMES` (bulk backfill on pairing) | **excluded** | `gents-schemas/src/lib.rs:175` |
| agent P2P reconcile profiles | **absent** | `agent/p2p_reconcile/profiles.rs` |
| desktop live-fixture runner subscription | **excluded, with a test** | `apps/gents-desktop/src-tauri/src/runner/live_fixture/replication.rs:165,183` |
| **production desktop client subscription** | **INCLUDED** | `gents-desktop-core/src/client/schema.rs` — `subscribed_collection_names()` returns `RUNTIME_COLLECTION_NAMES ++ ALL_COLLECTION_NAMES`, and `RENDERED_REQUEST_NAME` is in the latter (`gents-schemas/src/lib.rs:151`); there is no filter |

So a paired desktop **does** replicate these rows. That is consistent with how `AgentMessage` already behaves — the desktop is the user's own device — but `RenderedRequest` carries strictly more than `AgentMessage` does: the assembled system prompt and the full resolved tool surface, not just conversation turns. Reviewers should decide deliberately whether that is wanted before merge, rather than inheriting it from catalog membership. If it is not, the fix is a filter in `subscribed_collection_names()` mirroring the live-fixture one.

---

## The branchability finding

The branchability workstream (former Task 8 — flipping **existing** collections to `@branchable`) was **deleted entirely**, for two independent reasons:

1. **It is unnecessary.** Per-field and composite commit blocks are written **unconditionally** (`db-blocks/src/write.rs:59`, composite at `:268`). `is_branchable` gates only the *collection-level* block (`db/src/doc_mutator.rs:259`), and it appears exactly once in `crates/query/` — as an ACP-scope argument at `runner/commits.rs:195`. The content address this design relies on does not depend on it.
2. **It is impossible under the pinned DefraDB.** `validate_branchable_not_mutated` rejects every patch shape that flips it, and populated collections cannot be dropped and recreated.

No existing collection is made branchable by this PR. The **new** `RenderedRequest` collection is authored `@branchable` from birth — the one moment the property can be chosen, since the flip is irreversible in both directions. It is taken not for the commit CID (see above) but as the sole precondition for two capabilities this fact record will want and could not acquire later: DefraDB's branchable-collection sync, which is a newly paired peer's only bulk backfill path and rejects non-branchable collections outright, and collection-scoped ACP read decisions, which is the enforcement surface #840 needs once defradb.rs#1318 lands.

---

## Known follow-ups

- **#523's fidelity contract is unresolved.** The issue text says "the exact bytes that were sent." What is persisted is **canonical JSON** of the body at the transport seam — semantically equal to what went on the wire, but key order and serializer formatting may differ. Before #523 can be called complete, either the issue's acceptance language moves to semantic-JSON equality or a literal-bytes capture path is added. The Lean model declares this as `boundary.rendered-capture.assembled-request-artifact` rather than papering over it.
- **`session_id` validation gap.** `ChatArgs::session_id` is `Option<String>` with **no `value_parser`** (`crates/gents-cli/src/cli/args.rs:713`). Because `AgentToolCall.tool_call_key` is an unescaped `"{session_id}:{tool_call_id}"`, a caller-controlled `session_id` containing `:` makes key shadowing **realizable**, not hypothetical. (The capture key itself is not exposed to this: it encodes the scope as a nested JSON pair and has injectivity tests against exactly these delimiter-collision shapes. The gap is in the pre-existing `tool_call_key`.) This is the live instance behind the `boundary.rendered-capture.key-encoding-injectivity` boundary.
- **Payload size and retention are unbounded.** Each row stores `request_json` plus an `AssemblyTrace` containing the conversation again in native form — roughly doubling an already quadratic per-turn payload — once per turn per attempt, with no pruning, TTL, or size guard, and no operator off-switch for capture. Content-addressing or compressing the fact record is the known optimization; acknowledged in `rendered_request/mod.rs`.
- **`gents-protocol::row` has no `RenderedRequest` type yet** — readers currently hand-parse `serde_json::Value`. Deferred to the projection work, where readers actually appear; adding it now would be unexercised surface.
- **The sink's `create_conflict` branch is uncovered.** It is reachable only via a genuine concurrent-writer race; every deterministic construction takes the `lookup` branch instead.

---

## Verification

| gate | result |
|---|---|
| `lake build` (`crates/gents/proofs`) | Build completed successfully; `rg sorry Proofs/` → nothing |
| `#print axioms` on all 7 theorems | `[propext]` only; one needs no axioms at all |
| `cargo test -p gents` | 0 failures across all 15 targets incl. doc-tests |
| `cargo test -p gents-schemas -p gents-protocol -p gents-migration` | 0 failures |
| `cargo test -p gents-cli` | 0 failures |
| `cargo check --workspace --all-targets` | clean; no workspace-crate warnings |
| `cargo fmt --all --check` | clean |

**Mutation probe.** The plan flagged that `attempt_distinguishes_facts` was proven in Lean with no Rust fence behind it — an earlier probe replacing the loop's `attempt` with a literal `0` passed `cargo check` and failed **no test**. That hole is now closed at both the loop-argument and durable-row level. Re-applying the probe fails four fences:

- `capture_seam_reports_distinct_attempts_and_the_repair_build_path` → `[(0,0,Budgeted),(1,0,Budgeted),(1,0,Budgeted),(1,0,Repair)]` vs expected `[(0,0,…),(1,0,…),(1,1,…),(1,2,Repair)]`
- `capture_seam_reports_the_repair_build_path_from_the_first_poll_branch` → `[(0,0),(0,0),(0,0)]` vs `[(0,0),(0,1),(0,2)]`
- e2e `a_retried_attempt_is_its_own_durable_fact` → durable coordinates `[("inference.1",0,0),("inference.2",0,0)]` vs `[("inference.1",0,0),("inference.1",0,1)]`
- e2e `a_repaired_attempt_is_a_second_fact_with_a_different_canonical_request` → request terminates `failed`, the sink having correctly rejected the rebound key

The probe was reverted; the delivered tree passes all four.

**`cargo test -p gents-cli` was red before this PR** and is fixed here: `graphql_string_field` writes `""` for an unset `reasoning_effort`, so every profile `gents init` creates round-trips through `config export` as an empty string, which #988's new validator rejected. The validator now reads empty as unset, matching what the runtime already does.

Also fixed in the final commit, both found by adversarial review rather than by the suite: the uncaptured pre-request summarizer described above, and the structurally fail-open transport that made it invisible.


jackzampolin:
Independent verification note, from re-running the gate outside the workflow that produced this branch.

**One intermittent failure found**, filed as #1060: `run_agent_shutdown_is_prompt_while_request_waits_for_backend_capacity` timed out (`timed out waiting for 1 chat request(s), observed 0`) on one full-package run, then passed in isolation (0.53s) and passed a full lib re-run (1361/0). Low rate — the workflow ran the suite twice end to end at 1923/0 both times.

I did **not** determine whether it predates this branch, and I'd rather say so than assume. The test file is untouched here, but the path it exercises is not: capture is default-on, which puts a synchronous DefraDB read-then-write on the critical path of every provider call immediately before the send. A deadline-based wait for a chat request to reach the backend is exactly what added pre-send latency would destabilise. #1060 records what would settle it — N runs of the lib target on `main` versus this branch.

Not blocking; all ten checks pass. Flagging because CLAUDE.md treats flaky tests as defects rather than noise, and because if it *is* branch-induced it's the first observable symptom of the capture-cost follow-up already listed in the description.

**Methodology caveat worth passing on:** `cargo test … | tail -N` reports *tail's* exit status, so a red suite reads as green. The gate agent caught this in its own first run and re-ran unpiped; I initially repeated the mistake reading its output. Every result above is from an unpiped run.

iverc (approved):

jackzampolin:
Review follow-up pushed in 4999cad7.

- Removed the public arbitrary capture-sink override; the external fault hook can only fail closed.
- Reduced common-path capture cost: reconstructible turns omit the duplicate native transcript, dynamic request-context and compacted turns retain the oracle, and byte-identical SSE reconnects reuse the successful durable write. Changed reconnect bodies still reach the integrity sink.
- Made Responses prompt_hash cover both instructions and input.
- Refused empty provider streams when the transport capture is still armed.
- Fixed #1060 by synchronizing the capacity/shutdown test on durable InferenceCall=running instead of unrelated HTTP arrival.
- Tracked the signer-versus-claimed-principal boundary in #1064. Adding branchable does not solve global uniqueness across replicas.

Local gates: lake build; cargo test -p gents; cargo check --workspace --all-targets. All green.

Scope remains explicit: this PR establishes the durable captured oracle and CapturedOnly provenance. Historical CID reconstruction and a Verified projection verdict remain Plan 2.

PR #1061, opened 2026-08-07
Propagate deterministic inference seeds

### LINKED ISSUE #1058: Propagate an explicit sampling seed through inference requests
## Why

Jack Bench fixed-digest experiment plans vary an explicit `seed` axis while holding task, grader, model/profile, adapter, hint, and budget identities fixed. Gents currently has no sampling-seed field in its inference profile or `request submit` overrides, so a runner cannot truthfully execute or retain that axis. Recording a seed only in benchmark metadata would mislabel identical runtime behavior as distinct cells.

This is also useful outside Jack Bench for replayable provider fixtures and bounded comparisons. The seed is a requested provider parameter, not a promise of bitwise determinism across hardware or server revisions.

## Contract

- Add an optional unsigned sampling seed to the durable inference-profile contract and per-request overrides.
- Expose it through `gents config profile set` and `gents request submit`.
- Resolve and retain the effective seed with the rendered/completion request.
- Propagate it to supported OpenAI-compatible chat-completions request bodies (and other wire APIs only where the provider contract supports it).
- Preserve the same effective seed across transport retries and compaction continuations unless an explicitly documented policy says otherwise.
- Include the requested/effective seed in trace/ATIF or adjacent retained request evidence so benchmark importers can verify it.
- Reject unsupported or lossy provider paths rather than silently dropping a requested seed.

## Acceptance

- Unit tests cover profile defaulting, per-request override precedence, serialization, and invalid values.
- Provider fixture tests assert the exact outbound seed and retry stability.
- CLI tests cover both profile and request flags.
- The Harbor Gents adapter can accept a `GENTS_SEED` value and prove the retained effective seed matches the benchmark cell.

Context: source-inc/gents#988 added the ATIF/Harbor adapter used by Jack Bench; jackzampolin/amygdala#202 tracks the fixed-digest experiment runner.

jackzampolin:
Closes #1058

## Summary

- add nullable sampling seeds to AgentRequest and InferenceProfile through frozen-baseline versioned migrations
- preserve request-over-profile precedence and reject negative or unsupported-provider seeds before inference
- propagate seeds through runtime, CLI, desired state, desktop, background paths, and profile mutation/query surfaces
- pass the seed through the Harbor adapter, read the persisted request back, and bind result metadata to the persisted value
- make repeated same-collection migration steps idempotent without reactivating inactive ancestors

## Verification

- `CARGO_INCREMENTAL=0 cargo fmt --all -- --check`
- `CARGO_INCREMENTAL=0 cargo test -p gents`
- `CARGO_INCREMENTAL=0 cargo test -p gents-migration`
- focused CLI, desktop, watcher, provider-boundary, and document-config tests
- `python3 scripts/harbor/test_gents_agent.py` (8 tests)
- `sh -n scripts/harbor/run_gents.sh`
- Python bytecode compilation for the Harbor adapter and tests

## Review notes

The first local Claude review found missing frozen-baseline migration steps, Lean field catalogs, provider fail-closed validation, and persisted Harbor readback; all four were addressed. Subsequent full-diff local Claude runs exhausted their turn caps without a verdict, so this remains draft for repository review.


jackzampolin:
Superseded by #1062. The aggregate-budget branch contains this complete seed series, and #1062 now targets main so the deterministic seed and request-wide budget contract can be reviewed and landed as one change.

PR #1065, opened 2026-08-07
Make runtime facts durable and bind exact DefraDB provenance

jackzampolin:
## Summary

This draft implements the critical signed-ingest and exact-provenance durability checkpoint from #1063.

- verifies the cryptographic author of the exact incoming `AgentRequest` version and keeps requester attribution distinct from commit authorship
- binds request claim, selected configuration, transcript inputs, compaction inputs, inference attempts, rendered provider payloads, tool facts, forks, and terminal outcomes through exact `_docID` + composite-CID + verified-signer references
- publishes a formally specified `RunTimelineManifest` that freezes exact sources, records open-domain gaps, and preserves explicit ACP-denied omissions instead of silently claiming completeness
- makes finalized transcript, compaction, tool result/approval, and response-outcome facts immutable or fail-closed on rebinding
- rejects ambiguous logical twins across runtime, lifecycle, session, memory, interrupt, queue, timeline, and desktop retry paths; writers mutate the resolved physical document
- routes remote GraphQL and transaction access through one authenticated identity boundary with exact-host audience binding, issuer verification, bounded token lifetime, refresh timeout, TLS off loopback, redirect refusal, and cross-origin refusal
- treats subscriptions as wake-up hints while durable rescans and reconciliation remain the correctness mechanism
- documents branchability, CID, identity, retention, gossip, and deferred ACP/encryption boundaries for all shipped schema collections

These schema changes are deliberately breaking. No compatibility migration is included.

## Durable provenance chain

```text
verified signed request version
  -> agent-signed exact claim descendant
  -> exact signed config + transcript + compaction sources
  -> signed running InferenceCall
  -> immutable RenderedRequest
  -> exact send authorization edge
  -> signed terminal outcome
  -> frozen partial-exact or verified-exact timeline manifest
```

Logical IDs remain correlation and idempotency keys. They are never treated as document identity or as proof of the version consumed.

## Identity and policy boundary

Every embedded document read/write defaults to the configured node DID, while explicit trusted query identity still takes precedence. Gents refuses unsigned nodes and refuses a node signer that differs from the runtime principal. Remote document access carries a verified DefraDB bearer on GraphQL and transaction REST operations.

ACP policy documents and encryption/key custody remain intentionally deferred workstreams. This PR supplies attributable reads/writes and cryptographic provenance; it does not claim that future policy authorization or confidentiality is already installed.

## DefraDB dependency

[...]

Issue #1063, opened 2026-08-07
Schema architecture audit: conversations, responses, projections, and configuration


jackzampolin:
## Goal

Audit Gents' document and projection schemas as one architecture, then turn the findings into focused implementation issues. The database is the control plane and the durable fact store, so schema identity, lineage, versioning, and lifecycle semantics must be explicit enough to support arbitrary later projections—including the product UX—without relying on lookup conventions or process-local knowledge.

This is a tracking issue, not a request for a single schema rewrite. Each track below should be reviewable and completable independently.

## Shared review questions

For every collection or projection contract, record:

- What real-world entity or immutable fact does one document represent?
- What is its stable logical id, its DefraDB `_docID`, and the exact commit CID/version used for provenance?
- Which identity authors each transition, and which signatures/hash links let a later reader verify that lineage?
- What is the intended uniqueness scope: global, principal, session, request, turn, or deployment?
- Which relationships are durable ids versus denormalized query conveniences?
- Which fields are immutable facts, mutable lifecycle state, cached materializations, or presentation-only data?
- Which state combinations are illegal, and where are they enforced: schema, Lean transition model, conformance test, or writer validation?
- Are field types and nullability honest (`DateTime` versus `String`, absent versus empty string, list versus encoded JSON)?
- Do indexes match actual hot queries without turning every descriptive field into an index?
- Is the collection's branchability, replication, retention, and migration story intentional?
- Can a future projection reconstruct the relevant state from durable facts without guessing which duplicate row or mutable version applied?

Cross-cutting findings should become child issues with a narrow invariant and migration plan. Changes to legal lifecycle states or transitions follow the foundation flow: Lean model → conformance tests → implementation.

## Track A — Conversation and session schemas

Primary inventory:

- `AgentSession`
- `AgentConversation`
- `AgentRequest`
- `AgentMessage`
- `CompactionEntry`
- `AgentToolCall`, `AgentToolResult`, and `AgentToolApproval`
- `Goal` and session fork/retry/parent lineage where they participate in the transcript

Review focus:

- Clarify the overlap and ownership boundary between `AgentSession` and `AgentConversation`.
- Define and enforce uniqueness scopes for session, request, message sequence, compaction sequence, and tool-call identities. In particular, decide the invariant for `(agent_did, session_id, request_id)` rather than relying on UUID generation and `limit: 1` lookups.
[...]

PR #1087, opened 2026-08-09
Use DefraDB-native schema provenance and signing

jackzampolin:
Part of #1063. This replaces closed oversized PR #1065 with the reviewable implementation path.

## Architectural scope

- DefraDB remains the control plane and durable fact store.
- No request-intent/execution split, deployment-assignment subsystem, distributed scheduler, or parallel persistence framework.
- ACP policies and encryption are intentionally deferred. This PR establishes the signed document/version provenance they can later govern.
- Schema changes are intentionally breaking; no Lens migration is included for this pre-release reset.

## What changed

- Audits the 43 collections for branchability, identity, CID use, retention, writers, and projections.
- Keeps shared canonical facts and desired configuration branchable while keeping host-local runtime state local and out of P2P profiles.
- Registers existing agent keys with DefraDB and configures CLI, desktop, and initialized offline nodes to author signed commits.
- Consolidates ordinary GraphQL execution on the identity-aware DefraDB retry path and removes duplicate wrappers.
- Binds request-scoped messages, responses, tool calls, inference calls, compactions, subagent lineage, and trigger sources to exact physical DefraDB document IDs.
- Carries exact request document/version provenance into immutable rendered-request captures and fails closed when concurrent heads make the authored CID ambiguous.
- Makes timeline and adapter projections consume and corroborate those physical edges, including forked history and request-only goal, steering, and background-completion continuations.
- Uses the real single-node inference demo to verify signed provenance plus timeline, ATIF, OpenAI-Codex, LangGraph, and multi-agent projections.

## Review correction

The follow-up correctness review produced 15 findings. Twelve are corrected here:

- Steering now clears both obsolete tool-edge fields and atomically commits its exactly bound user message with the continuation in one DefraDB transaction.
- Background notifications use the real parent request logical/physical pair; their synthetic value is only an idempotent `message_key`.
- Background-completion wakes are request-only continuations rather than copied spawn bridges.
- Watcher and subagent scans quarantine malformed rows independently while preserving same-session exclusion and allowing later repaired rows to run.
- Timeline loading scopes nested child-session rows without weakening validation for rows claiming the requested timeline; logical-only external captures remain supported only when every physical ID is absent.
- Hook bindings cannot enter half-bound state, `RenderedRequest` participates in projection ACP filtering, and demo source-edge assertions require signed provenance.
- `Proofs/DurableLineage.lean` models logical/physical pair coherence, per-row quarantine, steering normalization, and message-before-request publication; the generated R4C witness fences the Rust behavior.

Three findings are intentionally retained under the breaking provenance contract:

- Logical-only approval rows cannot authorize a physical tool-call document.
- Offline session fork requires initialized signing identity rather than permitting unsigned writes.
- Claim success requires the mutation-authored CID; a post-update read cannot safely reconstruct it under branchable concurrency.

The review run also exposed an existing background-cancellation race. The operator-authored terminal cause is now persisted before signalling the live worker, preventing generic `interrupted` from overwriting `userCancelled`.

## Validation

- `lake build` (zero `sorry`s)
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p gents` full package suite green:
  - 1,425 unit tests passed; 1 ignored
  - 335 conformance tests passed
  - 41 lifecycle E2E tests passed
  - 59 runtime E2E tests passed; 1 ignored
  - 99 subagent E2E tests passed
- Background cancellation race regression passed 5/5 consecutive real-DB runs.
- Focused CLI projection ACP and demo provenance tests passed.
- Real `gents demo run pipeline --keep-home` acceptance run with signed provenance and all projection formats passed on the base implementation.
- Independent Claude review completed; all applicable concrete findings were addressed. Grok review attempts exhausted their turn budget without returning findings.

## Review notes

- The complete PR is about 7.8k additions / 2.0k deletions across schema call-site propagation, fixtures, generated projection contracts, and acceptance coverage. It does not add a second control plane.
- Public projections redact rendered-capture document locators; structural request/version provenance remains available by contract.
- DefraDB currently does not identify the mutation-authored CID when two visible heads share the newest height, so capture fails closed instead of guessing.
- Remote HTTP caller authentication, ACP policy definitions, encryption, retention jobs, and multi-node replication acceptance remain follow-up work.


jackzampolin:
Final readiness pass complete. The callsite audit covered exact _docID joins, logical/physical binding pairs, schema-selected fields, GraphQL escaping, transaction-conflict retry, mutation response parsing, background wake coalescing, and timeline projection validation. The final CI follow-up moved the remaining direct conformance mutation onto the shared retry path. Local gates passed: cargo test -p gents, cargo check --workspace --all-targets, Lean proof replay, and the focused 336-test conformance rerun. GitHub CI is now 10/10 green on a53d7d41; the PR is non-draft, CLEAN, and MERGEABLE.

PR #1095, opened 2026-08-11
Align durable tool outcome projections

jackzampolin:
## What

- Persist trusted command, native filesystem, meta-tool, and policy-denial failures as typed durable outcomes while preserving model-facing detail.
- Carry background timeout class and `deadline_exceeded` completion reason through the durable bridge.
- Make Codex and trace projections prefer canonical `ToolCallState` plus persisted failure class; result-text inference is compatibility-only for historical rows without lifecycle state.
- Move stream equality and collaboration event decisions into a Codex-independent projection model.
- Keep an emitted tool item kind monotonic across reciprocal-link flicker and late links.
- Emit one `started -> completed` lifecycle for a linked spawn; child presentation changes do not re-open or re-complete that item.
- Reuse the generic `gents-protocol` client turn projection for both live Codex turns and history reconstruction.
- Centralize persisted response-status aliases in the protocol crate instead of decoding them separately in GraphQL, desktop, and Codex paths.

## Why

The runtime previously had two competing meanings for a tool result. Trusted adapters rendered useful failure envelopes as `Ok(String)`, so persistence could record `completed` while consumers inferred failure by parsing arbitrary result text. The Codex shim also used Codex wire enums and raw child storage strings as its equality state, allowing equivalent lifecycle aliases and child updates to re-emit the same spawn item.

Turn status had the same duplication: the Codex live stream, Codex history reconstruction, generic client protocol, GraphQL client, and desktop client each carried overlapping request/response precedence logic.

The durable runtime now owns outcome semantics. The generic client projection owns turn derivation. The Codex shim observes those models, de-duplicates with neutral state, and serializes to Codex only at the edge.

## Projection contract

- Persisted lifecycle and a valid persisted failure class are authoritative.
- Legacy result parsing runs only when lifecycle state is absent.
- `claimed`, `processing`, `inputRequired`, and `running` normalize to one child presentation state.
- A stable tool-call ID retains its emitted item kind through link visibility changes.
- A linked spawn emits at most one `item/started` and one `item/completed`; later child state changes do not emit another tool lifecycle.
- A durable completed command cannot expose a contradictory nonzero exit code from stale result text.
- Live and historical Codex turn statuses are `derive_persisted_attempt` followed by a wire-only `ClientTurnState -> codex::TurnStatus` mapping.

## Formal model

The Codex turn projection now composes `Proofs.Client` rather than restating request lifecycle, response precedence, monotonicity, and terminal coherence. The remaining Codex layer maps `ClientTurnState` to Codex phases and adds the acknowledged local-interrupt override.

Generated Codex conformance cases are evaluated from that composition and cite the generic client theorems. The genuinely Codex-specific models remain for subagent item shape, reasoning items, binding, child visibility, and wire lifecycle behavior.

This follow-up removes 343 lines from the Codex Lean projection and fixture source while retaining the generated Rust fence.

## Checks

- `git diff --check`
- `lake build` — full proof suite passes
- `CARGO_INCREMENTAL=0 cargo test -p gents-protocol --lib` — 131 passed
- `CARGO_INCREMENTAL=0 cargo test -p gents-cli --lib` — 582 passed
- `CARGO_INCREMENTAL=0 cargo test -p gents --test conformance` — 336 passed
- `CARGO_INCREMENTAL=0 cargo test -p gents --lib` — 1,467 passed, 1 ignored
- Focused spawn lifecycle, projection-kind flicker, generic turn precedence, durable poison-text, timeout, policy-denial, and command exit-code regressions pass.

The full `cargo test -p gents` gate reaches an unrelated existing `e2e_runtime` failure: nine completion-retry/rendered-capture cases panic at `admission/permit.rs:202` with “can call blocking only when running on the multi-threaded runtime,” then time out with their request still `processing`. The library and complete conformance suites pass before that integration failure.

Earlier baseline runs also found unrelated pre-existing P2P terminal-redrive and trace-fixture failures.

Stacked on #1062.

### DIFF
--- a/crates/gents-cli/src/commands/codex_shim.rs
+++ b/crates/gents-cli/src/commands/codex_shim.rs
@@ -32,6 +32,7 @@ mod handlers;
 mod history_projection;
 mod host_runtime;
 mod progress;
+mod projection_state;
 mod protocol;
 mod store;
 mod subagent_projection;
--- a/crates/gents-cli/src/commands/codex_shim/background.rs
+++ b/crates/gents-cli/src/commands/codex_shim/background.rs
@@ -12,12 +12,11 @@ use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 
 use super::command_projection::{
-    codex_command_status, command_execution_item, command_output_payload, tool_projection_status,
-    ToolProjectionStatus,
+    codex_command_status, command_execution_item, command_output_payload,
 };
 use super::progress::{
-    decode_gents_tool_call_progress, gents_tool_progress_query, tool_completed_at_ms,
-    GentsToolCallProgress,
+    decode_gents_tool_call_progress, gents_tool_progress_query, observed_tool_status,
+    tool_completed_at_ms, GentsToolCallProgress,
 };
 use super::projection_state::ProjectionStatus;
 use super::protocol::{now_millis, send_notification};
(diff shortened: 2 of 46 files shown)


jackzampolin:
Background watcher durability follow-up landed in `511c170d`:

- retries transient DefraDB query failures instead of abandoning the watcher
- retains tracked command keys when a query snapshot temporarily omits the row
- observes the durable tool lifecycle directly after the item kind has been established, removing the fallback that could fabricate a successful command completion from presentation-field drift
- adds integration regressions for both a missing row that later terminalizes and an actual query error followed by schema/database recovery

Verification: `cargo test -p gents-cli --lib` (584 passed) and `cargo check --workspace --all-targets`.

PR #1103, opened 2026-08-12
Reuse same-commit release artifacts

jackzampolin:
## What changed

Release workflows now provide immutable Git provenance explicitly to `gents-cli`'s build script. When an explicit commit SHA is present, the build script no longer watches Git administrative files that `actions/checkout` rewrites on every retry. Local builds retain the existing Git discovery and dirty-state behavior.

This is stacked on #1102 so it uses the trusted, runner-affine release target established there.

## Why

Two identical unsigned macOS dispatches on #1102 still rebuilt and linked `gents-cli`, taking 524s and 426s inside Cargo. The final crate remained dirty because checkout rewrote the watched `.git` HEAD/ref files even though source and commit were unchanged.

The release workflows continue to set `SCCACHE_RECACHE=1`; this change reuses the trusted Cargo artifact only and does not permit release builds to consume PR-populated sccache objects.

## Measurement

Two identical unsigned macOS dispatches of `ef1585af` on the same trusted target tree produced:

| Cache state | Cargo build | Workflow job | Raw binary | gzip -9 | Archive |
|---|---:|---:|---:|---:|---:|
| first explicit-metadata artifact | 424s | 8m32s | 70,300,112 | 26,575,189 | 26,637,333 |
| identical-SHA retry | 0.53s | 1m34s | 70,300,112 | 26,575,189 | 26,637,341 |

The retry avoids 423 seconds (99.9%) of Cargo work and cuts the complete job by 6m58s (81.6%). The raw binary and gzip payload sizes are identical; the eight-byte archive difference is packaging metadata. This is specifically an exact same-commit retry improvement—a source edit still requires the representative final-crate compile/link measured in #1102.

Raw runs: [first artifact](https://github.com/source-inc/gents/actions/runs/31608891843), [identical-SHA retry](https://github.com/source-inc/gents/actions/runs/31609716279).

## Validation

- `cargo test -p gents`
- `cargo check --workspace --all-targets`
- `cargo fmt --all -- --check`
- focused explicit/fallback build-script output harness
- focused `gents-cli` version metadata unit test
- actionlint on both release workflows
- `git diff --check`

PR #1102, opened 2026-08-12
Speed trusted macOS releases and fix graph metrics

jackzampolin:
## Measured impact

The v0.11.0 macOS release spent **2,005s (33m25s)** inside Cargo on a cold release-only target tree. With this change, two clean-checkout runs took **524s and 426s (8m44s and 7m06s)**: **73.9–78.8% faster / 3.83–4.71x speedup**. Cargo rebuilt only `gents-cli`; dependencies and the core runtime stayed reusable.

The release workflow could land on any of four Studio registrations, but only `studio-2-2` currently owns the trusted 8.4 GB release target produced by v0.11.0. This change pins macOS releases to that registration so future releases deterministically reuse prior trusted Cargo artifacts.

The release still sets `SCCACHE_RECACHE=1`: any rustc invocation bypasses PR-populated sccache objects. This trades automatic runner failover for trusted warm-target reuse.

Raw runs:

- v0.11.0 cold trusted target: https://github.com/source-inc/gents/actions/runs/31554205416
- pinned trusted target, clean checkout 1: https://github.com/source-inc/gents/actions/runs/31601223603
- pinned trusted target, clean checkout 2: https://github.com/source-inc/gents/actions/runs/31602179888

| Cache state | Cargo build | Raw binary | gzip -9 | Archive |
|---|---:|---:|---:|---:|
| v0.11.0 cold trusted target | 2,005s | 70,300,272 | 26,580,064 | 26,641,927 |
| pinned trusted target, clean checkout 1 | 524s | 70,299,648 | 26,577,430 | 26,639,171 |
| pinned trusted target, clean checkout 2 | 426s | 70,299,648 | 26,577,430 | 26,639,171 |

The optimized measurements were unsigned dry runs, so their size differences include the missing Developer ID signature and are not a binary-size claim.

## Measurement correctness

The v0.11.0 reports' 1,129–1,141 normal-package counts were inflated by ANSI coloring, not DefraDB v0.18.0 or dependency growth. `dtolnay/rust-toolchain` exports `CARGO_TERM_COLOR=always`; Cargo colors the repeated `(*)` marker, so the old plain-text suffix removal failed.

Reproduction on the v0.11.0 lockfile:

| Target | Published schema v1 | Corrected schema v2 |
|---|---:|---:|
| macOS arm64 | 1,141 | 819 |
| Linux arm64 | 1,129 | 813 |
| Linux amd64 | 1,131 | 815 |
| all targets | — | 962 |

Schema v2 now:

[...]

PR #1104, opened 2026-08-12
perf: attribute release compile time and linked size

jackzampolin:
## Measured build-time impact

The corrected cold macOS arm64 run used a fresh target, direct rustc, no compiler cache, 12 Cargo jobs, `CARGO_INCREMENTAL=0`, and Rust 1.97.1:

- stripped release build: **1,788s (29m48s)**
- peak RSS: **10,174,103,552 bytes**
- raw binary: **70,693,328 bytes**
- gzip-9: **26,994,237 bytes**
- graph: **819** target-scoped normal packages, **962** all-target packages, **68** duplicate names

This reproduces the v0.11.0 2,005-second release regime within 10.8% and explains the 209--214-second experiment: the experiment read the host's warm shared compiler cache, while the tagged release and this run compiled the cold native graph. Signing and packaging are outside this 1,788-second Cargo measurement.

Top aggregate compile contributors were `librocksdb-sys` 519.58s, `gents-cli` 405.37s, `aws-lc-sys` 293.14s, `cranelift-codegen` 189.13s, `gents` 178.74s, `iroh` 158.03s, `gents-migration` 156.83s, and `rmcp` 121.55s. Package timings overlap under parallel compilation and are not summed into wall time.

Raw run and retained artifact: https://github.com/source-inc/gents/actions/runs/31613287268

## What changed

Adds a manual, cold macOS arm64 release-attribution workflow and a reusable measurement script. Each run uses a fresh target directory with compiler caching disabled and uploads:

- Cargo's raw timing HTML and extracted per-unit/per-package JSON rankings
- full Cargo metadata and the resolved feature tree
- duplicate-version inverse paths
- pinned `cargo-bloat` 0.12.1 per-crate linked-size JSON
- build log, exact commands/tool versions, peak RSS, binary/gzip sizes, and the schema-v2 dependency report
- a compact machine-readable `summary.json`

The raw reports remain workflow artifacts rather than checked-in generated data. The script refuses a nonempty target directory unless a local-only override is explicit, preventing a warm build from being mislabeled cold.

This is stacked on #1103 and #1102.

## Why

The corrected graph count tells us that the CLI resolves 819 macOS packages, but package count alone cannot identify expensive compilation or linked binary contributors. This workflow keeps wall time, peak memory, binary size, compressed size, graph size, aggregate package compile duration, and linked code size as separate signals so the next removal can be chosen from evidence.

The workflow is manual and uses a fresh `${RUNNER_TEMP}` target on the `studio-1` support registration. It does not use the trusted signed-release target or shared sccache, and it cannot contaminate or consume signed-release compiler artifacts.

## Linked-size attribution

`cargo-bloat` 0.12.1 performed a separate 18m01s symbol-preserving rebuild. Its 122,703,208-byte inspection binary is not the shipped binary; only its 57,632,052-byte text section is used for linked-code attribution. The largest linked contributors were `gents_server` 7,106,696 bytes, `std` 6,201,385, `gents` 5,701,092, `librocksdb_sys` 3,189,724, DefraDB `db` 2,590,420, DefraDB `query` 2,155,308, `cranelift_codegen` 2,107,580, and DefraDB `p2p` 1,881,492.

The full workflow took 51m29s because it intentionally performed both builds. That total is not reported as release build time.

## Validation

- `cargo test -p gents` (full package suite)
- build-attribution parser self-test
- feature-tree and duplicate-tree command probes
- Bash syntax and ShellCheck
- actionlint for the new workflow and changed CI workflow
- `git diff --check`

## Tradeoffs and roadmap

This adds an intentionally expensive manual cold-build job, not a required PR check. It installs one pinned inspection tool and retains reports for 30 days. It does not establish a budget yet; normal variance and at least one repeated baseline still need to be measured.

The timing and linked rankings independently point at the same upstream boundaries: SourceHub/AWS-LC, the combined Iroh plus legacy-libp2p graph, and Wasmtime/Cranelift. Those are tracked in sourcenetwork/defradb.rs#1398, #1399, and #1400. RocksDB remains the largest native cold-build contributor, but it is part of Gents' required runtime storage path rather than an optional removal candidate.

### DIFF
--- a/.github/workflows/build-attribution.yml
+++ b/.github/workflows/build-attribution.yml
@@ -7,6 +7,12 @@ on:
         description: Exact 40-character commit SHA expected at the dispatched ref.
         required: true
         type: string
+  workflow_call:
+    inputs:
+      candidate_sha:
+        description: Exact 40-character commit SHA expected at the caller's ref.
+        required: true
+        type: string
 
 concurrency:
   group: build-attribution-${{ github.ref }}
--- a/.github/workflows/build-attribution.yml
+++ b/.github/workflows/build-attribution.yml
@@ -1,0 +1,86 @@
+name: Build attribution
+
+on:
+  workflow_dispatch:
+    inputs:
+      candidate_sha:
+        description: Exact 40-character commit SHA expected at the dispatched ref.
+        required: true
+        type: string
+
+concurrency:
+  group: build-attribution-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  macos-arm64-cold:
+    name: macOS arm64 cold release attribution
+    runs-on: [self-hosted, macOS, ARM64, studio, studio-1, ci-support]
+    timeout-minutes: 120
+    env:
+      CARGO_BUILD_JOBS: ${{ vars.CARGO_BUILD_JOBS || '12' }}
+      CARGO_INCREMENTAL: "0"
+      CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+      GENTS_BUILD_GIT_SHA: ${{ github.sha }}
+      GENTS_BUILD_GIT_REF: ${{ github.ref }}
+      GENTS_BUILD_GIT_TAG: ${{ github.ref_type == 'tag' && github.ref_name || '' }}
+      GENTS_BUILD_GIT_DIRTY: "false"
+      HOMEBREW_NO_AUTO_UPDATE: "1"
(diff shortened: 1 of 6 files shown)


jackzampolin:
The cumulative tested branch is now landing through rollup PR #1109 against `main` so the full optimization set has one authoritative CI run and one merge. Keeping this draft open as the focused measurement/review record until the rollup lands; do not merge this slice separately.

jackzampolin:
Landed as part of the cumulative squash merge in #1109 (`020b17db`). Closing this slice PR without a separate merge to avoid duplicating the rollup history.

Issue #1123, opened 2026-08-14
Client fresh-applied schemas no longer interoperate with v0.11's migration-chained collections (schema_version lineage divergence)


noahsaso:
Follow-on from #1122 with a sharper root cause, found live: after fixing the mobile app's stale bundled SDLs (gents-mobile#11 — bundle now byte-identical to gents main), phone-authored documents STILL fail to merge server-side:

```
Block merge failed … error=missing metadata: Collection not found for schema_version_id: bafyreic7me42…   ← corrected phone, fresh store
Block merge failed … error=missing metadata: Collection not found for schema_version_id: bafyreidm25tx…  ← older install's zombie blocks (persist across server restart)
```

Mechanism: the server's conversation-plane collections now carry migration-chain lineage (v0.11 evolved them via gents-migration steps), so the server's current `schema_version_id` is a chain tip. The client (gents-mobile's `LocalDBSchemaSmoke`) fresh-applies the final SDL to a fresh store, minting a genesis version CID. Identical SDL text ⇒ different version identity ⇒ every client write is unmergeable. This interop held until v0.11 only because those collections had no post-baseline migration history — the baseline pins WERE fresh-apply CIDs.

Consequences: as of v0.11, no fresh client store can author documents against an upgraded server, regardless of how current its bundled schemas are. This blocks all mobile chat.

Direction options (pick one deliberately):
1. **Client replays the same chain**: ship the migration chain (baseline + steps) to clients — e.g. embed gents-migration's registry data in the FFI/schema bundle so client stores build the same lineage and tips match. Heaviest but structurally honest.
2. **Server-side version acceptance**: the merge path accepts a document whose schema_version_id resolves to a schema DEFINITION identical to a known version's (content-equivalence aliasing), recording the alias. Contained to db-merge; keeps clients simple.
3. **Collections stop chaining for client-authored planes**: keep conversation-plane collections on re-pinned baselines only (the pre-v0.11 accidental invariant, made deliberate) — cheap but constrains schema evolution forever.

Also: the older install's unmergeable blocks are retried indefinitely from persisted state across server restarts (zombie retry noise); whatever direction is chosen, permanently-unknown versions need a terminal disposition (park + surface, not retry-forever) — same observability ask as #1122.

Live state preserved: server v0.11 home at ~/.gents (two unknown-version block families in its pending state), phone build 2026-08-14 with byte-identical bundle.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1043, opened 2026-08-05
Baseline re-pin orphans existing homes: UnknownLineage on upgrade across schema-modifying releases (no legacy lineage support)


noahsaso:
## What happened

Upgrading a live dev home across the persona-catalog release (#1014) bricked boot:

```
Error: ensure_migrations
Caused by:
    unknown lineage for collection AgentDirectoryEntry: versions [bafyreib3nckkjgeabgwk7uy2h4xc7ztmivf7qkrfzghrjpln6ze4i6e52m]
    do not match the migration baseline; export/import required (no legacy support)
```

The home was created while `AgentDirectoryEntry`'s baseline pin was the old CID. #1014 modified the collection's SDL and **re-pinned the baseline in place** (`gents-migration/src/registry.rs`, via the `chain_replay_prints_baseline_pins_for_authoring` authoring flow). Homes born under the old pin now have a lineage the registry no longer recognizes — there is no chain entry from old CID → new CID, and the loader's answer is "export/import required."

## Why this matters beyond dev

This is the standard upgrade path, not an edge case: **every schema-modifying release re-pins, and every home that predates the release hits UnknownLineage on its first boot after upgrading.** The schema-driven field sweep (#682/#688) handles *fields* for recognized lineages, but the lineage gate rejects the store before the sweep can run. Adding new collections is safe (fresh lineage); modifying existing ones is the trap. We only survived prior releases because they happened to only add collections.

Related but distinct: #983 (replay error-loop on unknown schema_version), #984 (unique-index poison pill). This one is the *designed* posture ("no legacy support") meeting reality.

## Direction

The registry needs a legacy-lineage mechanism so a re-pin records where homes may be coming FROM, e.g.:

- `baseline_entry!` grows an optional `legacy_versions: &[&str]` — the CIDs this collection was previously pinned at. `ensure_migrations` treats a store at a legacy CID as *upgradable*: register the new SDL (DefraDB schema add produces the new version), let the #682 field sweep patch rows, done. The authoring test can print the outgoing pin alongside the new one so the author moves it into `legacy_versions` instead of deleting history.
- Guard test: re-pinning a collection without carrying the old CID into `legacy_versions` fails the authoring flow — make forgetting impossible, not disciplined.
- Lean-first per house rules: the upgrade step wants a small model (legacy-recognized ⇒ converges to the same post-state as fresh-baseline; unknown lineage still refuses).

[...]

PR #1125, opened 2026-08-14
fix(migration): conversation-plane collections stay fresh-apply compatible (#1123)

noahsaso:
Implements #1123 option 3, deliberately: client-authored (conversation-plane) collections must remain fresh-apply compatible, because paired clients mint genesis schema-version CIDs from bundled SDL and a migration-chain tip CID (a DAG-CBOR block with `heads`) can structurally never match one.

Empirical basis (three-way probe, methodology in the issue thread): a fresh client's batch-apply and per-collection ordered apply both produce `bafyreic7me42…` for AgentRequest — the exact CID paired phones send — while the migration path produced a different chain-tip because AgentRequest alone carried a post-baseline `PatchVersioned` step (background-completion snapshot fields). Every phone-authored document was unmergeable ("Collection not found for schema_version_id"); reads worked, writes black-holed silently.

Fix: fold the step's fields into the baseline SDL registration (how the other four conversation collections already register), re-pin AgentRequest's baseline to the fresh-apply CID, remove the redundant step, update the two lineage guard tests. gents-migration 22/22 green; full suite green modulo the environmental lake site.

Hard cutover (accepted): existing v0.11 homes on the old chained lineage fail ensure_migrations on next boot — covered by the standing archive-on-upgrade policy (#1043, closed by design). Constraint going forward, worth a guard test in a follow-up: conversation-plane collections evolve by baseline re-pin only, never by chained steps, until #1123's option 1 or 2 lands.


noahsaso:
Now self-enforcing: `828b9bb64` adds the permanent guard from the standing-constraint note — `fresh_apply_parity.rs` replays the full migration chain and asserts every client-authored collection's active version CID equals a fresh direct apply of its current SDL (the assertion message explains the re-pin rule to whoever trips it), plus a fence test in gents keeping the guarded set in lockstep with the machine template's per-collection client rules. Both were observed RED against a simulated recurrence of exactly this defect before going green. gents-migration 23/23; fmt + workspace all-targets clean.

jackzampolin:
Pushed two commits after an in-depth review (Jack's session; findings verified against the code before implementing):

**`ef4d81f3`** — merge of current main. The `rust-and-cli (cli)` failure on the previous run was the live demo inference gate 404-ing on a model name; main removed that gate in d9954f8d, so this should come back green.

**`cd460bf7`** — review follow-ups. The core re-pin was verified sound (folded SDL reconciles with frozen SDL + deleted patch, all CID references consistent, pin enforced loudly at ensure time) and is untouched. What changed is guard coverage and fidelity:

- The static step guard now covers all 13 `CLIENT_AUTHORED_COLLECTIONS` and both step kinds — `PatchInPlace` keeps the version CID, so CID parity can't see that divergence; only a static guard can. Red-green verified: a planted in-place step on `AgentRequest` fails the build with a message that says to re-pin, not freeze-and-chain (the old `"changed schema must be frozen"` message prescribed exactly the #1123 pattern for the other 12 collections).
- The fence now runs in the PR runtime smoke step (it previously only ran on post-merge main pushes), its claim-collection literals come from `gents-protocol`, and a new desktop-core test ties `BEARER_CONTROL_PLANE_COLLECTIONS` to the guard list. Its doc is corrected (BearerPairingReady is engine-written; AgentDirectoryEntry filters on the home DID; PeerEndpoint is in the discovery/network-control templates) and its coverage claim scoped — the `DataPlanePairingDesired` gap is #1137.
- `fresh_apply_parity`: catalog position-lookup replaces the 13-arm match, a relation-free precondition test pins the client model's validity, the module doc cites #1123/#1125 and the defradb convergence point instead of the (never-committed) task report and probe file, and the two node boots run under `tokio::join!`.
- The `…and_migrations_are_idempotent` test now actually re-runs `ensure_migrations` and asserts the no-op.
- `fresh_node` deduped into `tests/common/` (tempdir leak kept deliberately and documented — defradb.rs #1309 sweep tasks).

Local gates: `gents-migration` 24/24, fence 1/1, `gents-desktop-core` 147/147, `cargo check --workspace --all-targets` clean.

PR #1160, opened 2026-08-20
refactor(sync): make requests own session projection

jackzampolin:
## Summary

- make AgentRequest the only client-authored input and let the runtime atomically claim it with the canonical AgentSession and AgentConversation projections
- model admission rejection, monotone reconciliation, and request/session ownership in Lean first, then drive Rust through the generated conformance cases
- remove desktop/mobile session materialization, startup conversation repair, legacy hook-created sessions, and the shared-app session-fork surface
- keep Codex threads as ordinary canonical Gents sessions; the Codex shim now projects pending and durable request state without persisting a second session kind or repairing unknown threads
- terminalize request and conversation state atomically while preserving a newer latest-request pointer
- box provider-specific completion futures at the runtime seam to keep polymorphic stack frames bounded

This is intentionally a clean cut with no migration or compatibility path. The desktop bridge contract moves from 0.9 to 1.0 because desktop_session_fork is removed. The result is 2,252 additions and 4,646 deletions across the model, conformance suite, runtime, clients, and adapter projection.

## Invariants

- clients submit requests; the runtime owns durable session projection
- accepted request IDs are monotone and replayed requests cannot rematerialize state
- a claimed request and its canonical session/conversation projection commit in one transaction
- deterministic admission failures become terminal failed requests without minting sessions
- Codex is an adapter view over canonical sessions, not a persisted session category

## Verification

- lake build
- cargo test -p gents (2,291 passed, 4 ignored across all targets)
- cargo check --workspace --all-targets
- cargo test -p gents-desktop-core --tests (all targets pass; corrected one stale canonical behavior assertion and reran submission_api 4/4)
- cargo test -p gents-desktop-bridge (121 passed, 3 ignored including docs)
- desktop frontend build
- desktop Vitest (352 passed, 15 live-only skipped)
- Codex shim tests (32 passed, 10 ignored)
- bridge snapshot suite (60 passed)
- cargo fmt --all
- git diff --check

Reviewed with both Grok and Claude CLIs; actionable ownership, adapter-boundary, lifecycle, and stack findings are resolved.


crates/gents/src/lifecycle/transition.rs
```diff
@@ -23,7 +23,173 @@ fn request_view_is_terminal(view: &RequestViewRow) -> bool {
...
+                }
+            };
+            // The request mutation may be an idempotent no-op because the
+            // stream writer already committed its terminal edge. The guarded
+            // projection update must still run so that retry repairs a stale
+            // conversation. If the projection statement fails, discard the
+            // whole attempt and commit the authoritative request alone in a
+            // fresh transaction. Never trust commit-after-error behavior from
+            // the storage engine.
+            if let Err(error) = txn.execute_local_response(conversation_mutation).await {
```
jackzampolin: **[bug]** `execute_request_projection_transaction` maps any conversation-mutation `Err` (including retryable DefraDB conflicts) onto a request-only commit, then returns `Ok` to `retry_terminal_persistence_operation`. After the discard, the request has not yet been terminalized, so a combined retry could still land both writes; instead the helper succeeds as soon as the request-only txn commits. Conversation recovery ranking/sweeps were removed in this PR, so a last-in-session terminal can leave `AgentConversation.status` stuck at `processing` with a stale `latest_request_id`. Lean `Request.Projection.terminalCommitAfterProjectionAttempt` describes a projection *statement* error, not a conflict that should retry the atomic pair.

**Suggestion:** If `is_defradb_transaction_conflict_text` (or the other retryable storage strings already used in claim/redrive) matches, return `Err(error)` so the outer retry re-attempts request+projection together. Keep the request-only fallback for genuine statement/schema failures. If request-only remains a path, reintroduce a bounded conversation repair so a last request cannot leave the sidebar in `processing`.

crates/gents/src/session/query.rs
```diff
@@ -27,7 +31,6 @@ pub(super) async fn load_session_document_optional(
                 }},
                 limit: 1
```
jackzampolin: **[bug]** Conversation reads were switched to `order: { updated_at: DESC }, limit: 1` because duplicate `session_id` rows exist in the wild (pre-unique-index stores and replica forks — the deleted comment on this file spelled that out). `load_session_document_optional` still uses unordered `limit: 1`. `resolve_behavior_for_request` uses that result as the session pin: an empty/stale replica can hide the pin (so a request is dispatched with the default behavior and later rejected at claim) or, if replicas disagree, pick the wrong non-empty `behavior_id` and reject a legal request. Codex `load_conversation` has the same unordered `limit: 1` (`crates/gents-cli/src/commands/codex_shim/thread_projection/storage.rs`), so the adapter can bind a stale `latest_request_id`/title after claim has converged every row.

**Suggestion:** Rank session reads the same way conversations are now ranked (newest `started`/`updated_at`, then `_docID`), or query without `limit` and apply the claim-side `behaviorCompatible` check over every row. Give Codex `load_conversation` `order: { updated_at: DESC }` (and a `latest_request_id` tie-break) so it cannot project a stale replica.

crates/gents-desktop-bridge/src/snapshot/session.rs
```diff
@@ -622,13 +635,22 @@ fn build_pending_turn(
...
+                    &keyed_steering_request_ids,
+                ),
                 timestamp: normalize_optional(row.timestamp.as_deref()),
             }
         })
         .collect::<Vec<_>>();
 
-    if materialized_user_turn_count(&messages) > active_turn_index {
+    let exact_owner = has_materialized_user_owner(&messages, request_id);
+    let legacy_owner = logical_turn_index_for_request(store, agent_did, session_id, request_id)
```
jackzampolin: **[suggestion]** `has_durable_user_owner` is correctly request-id scoped (and the protocol comment says missing/unrelated ids must not fall back to turn-count heuristics). `build_pending_turn` still suppresses the pending bubble when `materialized_user_turn_count > logical_turn_index`. That is what keeps a retry from re-showing the original user text, but replica user rows with the same content and different sequences count as extra turns (`sequence:{seq}:{content}`), so a later pending request can lose its bubble. Same-sequence duplicates are covered by the new unit test; different-sequence forks are not.

**Suggestion:** Treat a retry as owned when any visible user message carries the retry root (or original `request_id`), and drop the session-wide turn-count fallback. If the heuristic must stay for pre-key transcripts, dedupe by `(request_id, content)` rather than by sequence.

PR #1161, opened 2026-08-20
RFC: agentic SDLC — compile ECSS-E-ST-40C into a document DAG

jackzampolin:
**Draft for alignment — design only, no implementation yet.** Two docs, no code.

- [`docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md`](docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md) — the design
- [`docs/references/README.md`](docs/references/README.md) — which clauses of ECSS-E-ST-40C / 12207 it leans on and where to get them

## The idea

A GitHub issue labelled `sdlc:<A|B|C|D>` seeds a document DAG that carries it through the V-model to a PR. Humans keep the two tips — ideation and validation — and the trough is agent-run but reclaimable.

The framing is not "design a methodology for agents." It's **compile an existing one**: lift ECSS-E-ST-40C clause 5 and re-cast each activity as a bounded agent with defined inputs and outputs. Those processes exist because software failure could end lives, and they lost to agile on *overhead cost*, not correctness. Agents are cheap at exactly that overhead, so the trade-off flips.

## Why gents specifically

The mapping is almost frictionless because the control plane is already a document database:

| ECSS | Gents |
| --- | --- |
| Process activity (§5.4.2) | Task + behavior |
| Expected output `[DDF, SDD; PDR]` | A document in the `DDF` collection |
| Activity precedence | `EventTrigger`, `event_kind: created` |
| What an activity may produce | `DatastoreToolSurface` |
| Joint review (§5.3.3) | Review behavior + its RID ledger |
| CM baseline | `supersedes` chain, append-only |
| **Traceability matrix** | `derived_from` edges, DID-signed |
| Annex R tailoring | `TailoringRule` collection, queried at dispatch |

The traceability matrix row is the one worth pausing on. It's the single most expensive artefact in the standard, and here it's a `SELECT` over provenance the persistence layer already writes.

## Three findings that shaped the design

**1. Annex R is already a machine-readable tailoring table.** Keyed by stable requirement id, one row per expected output, `Y`/`N`/`Ytba` per criticality category. The issue label isn't an on/off switch — it's a criticality category that selects the subgraph. `sdlc:D` on a bug fix runs requirements → code → verify; `sdlc:A` runs the full V.

**2. `created`-only triggers are sufficient, and that's not a workaround.** `event_kind` v1 rejects everything but `created` (`crates/gents-cli/src/desired_state/validate.rs:727`), which looks like it blocks "a human overrides a node and downstream re-derives." It doesn't. Under ECSS §5.3 / 12207 §6.3.5 you never edit a baseline in place — you issue a superseding revision under change control. So an override is a *create*, it fires an edge, and the prior revision stays in the DAG as signed history. The append-only constraint and the standard's change-control discipline are the same constraint. **No runtime change, no Lean work** — this composes already-proven transitions.

**3. Verification and validation must stay separate nodes.** ECSS splits validation against the TS (§5.6.3, `[DJF, SVS; CDR]`) from validation against the RB (§5.6.4, `[DJF, SVS; QR, AR]`). Verification asks *did we build what the spec says*; validation asks *does the spec answer the need in the issue*. CI only ever does the first — and the second is exactly where an agent loop goes wrong, with a perfectly-implemented misreading of the issue.

That's also why the RB/TS split is load-bearing: the requirements baseline is the customer's (§5.2), the technical specification is the supplier's (§5.4.2). Collapse them and the SRR gate has nothing to approve, and final validation ends up grading the implementation against a spec the same agents wrote.

## Human surfaces

The transport is not the semantics. A GitHub comment and a paired mobile session both land the same document — a RID with a disposition, or a superseding revision.

GitHub issue comments first (`gh` is already reachable from pack prompts). Paired desktop/mobile sessions next, and those need no new plumbing: artefacts replicate to any paired client already, and `PairedRemoteOnly` exists.

## Structural precedent

`demo/repo-maintenance/` is a 6-node graph that already ends in a real PR — fan-out scanners, adversarial verification barrier, count-balanced work-package accounting, worktree isolation, bounded CI repair loop. This is that graph with a different seed and a longer artefact chain.

## Decisions from review

Folded in [`01be8353`](https://github.com/source-inc/gents/commit/01be8353d16881f088b95ca0bed66505c896a168) after Mike's comments. Details live in the spec.

1. **Fan-out width** — declared by the architecture node, not hardcoded. v1 bound is 1–6. Annex K.4 is the selection criteria.
2. **Intake argues** — a vague issue parks on an inbound RID. Cascade is adjacent-only. Left-of-V agents argue the spec; right-of-V agents argue the implementation.
3. **Re-derivation** — three change classes on the RID under a job-level eagerness policy (ECSS §4.3.3.4). Class 1 parks; class 2 re-derives and flags; class 3 re-derives quietly. Pre/post-approval folded together for v1.
4. **SRR for `sdlc:D`** — stays a hard human gate as an explicit MGT tailoring override of Annex R. Current target is one V for a 2–4 week feature; recursive V-of-Vs is out of bounds.

## Not in this PR

Implementation (`demo/sdlc/`), the paired-session review behavior, packaging packs as a `make` target for *other* repos, and recursive V-of-Vs. The distribution story is a separate concern; building this pack first is what discovers its requirements.

### DIFF
--- a/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
+++ b/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
@@ -1,0 +1,338 @@
+# Agentic SDLC pack — design
+
+**Status:** brainstormed in session 2026-08-20 on `agent/space-dev`; this document is the
+durable record and the source for the tracking issue.
+**Prototype target:** gents itself — GitHub issues labelled `sdlc:<A|B|C|D>`, driven against
+`workstation-1:8000` (GLM-5.2).
+
+## The bet
+
+Building software is the same craft it was twenty years ago: decompose a need, agree
+interfaces, implement, verify, validate. Heavyweight process — V-model, ASPICE, ECSS,
+ISO/IEC/IEEE 12207 — did not lose to agile because it was wrong. It lost because its
+overhead (traceability matrices, keeping contracts in sync, artefact generation) is too
+expensive for projects without clear requirements and strong up-front design.
+
+Agents are cheap and fast at exactly that mechanical overhead. Collapse its cost and the
+trade-off flips: rigour once affordable only for spacecraft becomes affordable for normal
+work.
+
+So the job is not to design a methodology. It is to **compile an existing one into agents** —
+lift ECSS-E-ST-40C and 12207 clause by clause and re-cast each activity as a bounded agent
+with defined inputs and outputs.
+
+The V, re-read as a value gradient: human judgement is worth most at the two top corners,
+ideation and validation, which stay hard human gates. The trough — architecture, component
+design, implementation, verification — is delegated, but **reclaimable**: a human can reach
+into any node and override it, and the machinery re-derives whatever that change
+invalidated downstream.
+
+Gents is an unusually good substrate for this because the control plane is already a
+document database. Every ECSS artefact is a document, every activity is a trigger edge, and
+DID-signed provenance means the traceability matrix — the single most expensive artefact in
+the standard — falls out of the persistence layer instead of being maintained by hand.
+
+## Constraints established by research
+
+Verified against this tree and the two standards, not assumed:
+
+- **`event_kind` is `created`-only.** `crates/gents-cli/src/desired_state/validate.rs:727`
+  hard-rejects anything else: *"v1 supports only \"created\""*. This does **not** block the
+  reclaimable-node property — see [Revision, not mutation](#revision-not-mutation).
+- **GitHub is already reachable via shell.** Packs call `gh pr create`, `gh pr view`,
+  `gh pr checks` directly from task prompts
+  (`demo/repo-maintenance/tasks/maintenance-publish-task/prompt.md:17`,
+  `demo/code-review/tasks/review-recon-task/prompt.md:9`). Issue-comment ingest needs a task,
+  not a new tool or integration.
+- **`GLM-5.2` is already the pack default** (`demo/repo-maintenance/agent-behaviors/
+  maintenance-publish/object.json`), so retargeting to `workstation-1:8000` is an endpoint
+  variable, nothing more.
+- **Clients render replicated documents, not bespoke views.** The desktop app "consumes the
+  replicated document surface through `gents-desktop-core`", and `apps/fixture-host` already
+  runs `BootstrapPolicy::PairedRemoteOnly`. RB/TS/DDF documents appear in a paired client
+  because they replicate. Mobile/web review is a pairing problem, not an integration.
+- **`demo/repo-maintenance/` is the structural template.** Six nodes, fan-out scanners, an
+  adversarial verification barrier, count-balanced work-package accounting, worktree
+  isolation, a real PR, and a bounded CI repair loop. This pack is that graph with a
+  different seed and a longer artefact chain.
+- **The standard is pre-annotated for this.** Every ECSS clause tags its outputs
+  `[file, DRD; review]` — e.g. `Software architectural design [DDF, SDD; PDR]`. Output
+  collection and gating review are given, not inferred.
+
+## Scope note: this does not touch Lean
+
+A pack is configuration applied to an already-proven runtime. Trigger dispatch, the request
+lifecycle, and fan-in accounting are Lean-fenced today; this work composes those existing
+transitions rather than adding new ones. The V's state machine lives in the schema graph —
+which collection a node may write, and which edge fires next — not in runtime semantics.
+
+If the prototype turns up a required runtime change (a new event kind, a new trigger
+predicate), that change follows the normal foundation flow and starts in
+`crates/gents/proofs/`. The design below is deliberately built to avoid needing one.
+
+## The V as a document DAG
+
+The pack primitives map onto the standard with almost no impedance:
+
+| ECSS concept | Gents primitive |
+| --- | --- |
+| Process activity (e.g. §5.4.2) | Task + behavior |
+| Expected output `[DDF, SDD; PDR]` | A document in the `DDF` collection |
+| Activity precedence | `EventTrigger`, `event_kind: created` |
+| What an activity may produce | `DatastoreToolSurface` on its `ToolSelection` |
+| Joint review (§5.3.3) | A review behavior + its RID ledger |
+| Review item discrepancy | A `Rid` document with a disposition |
+| Configuration management baseline | `supersedes` chain, append-only |
+| Traceability matrix | `derived_from` edges, DID-signed |
+| Annex R tailoring | A `TailoringRule` collection, queried at dispatch |
+
+### Revision, not mutation
+
+The obvious reading of "a human overrides a node" is an in-place edit, which would need an
+`updated` trigger we do not have. That reading is wrong on the standard's own terms.
+
+ECSS §5.3 and 12207 §6.3.5 both put artefacts under configuration management: you do not
+edit a requirements baseline, you issue a **superseding revision** under change control. So
+an override is:
+
+```
+create RequirementsBaseline { supersedes: <prior_doc_id>, rationale: …, author: <human DID> }
+```
+
+which fires a `created` edge, re-derives downstream, and leaves the prior revision in the DAG
+as signed history. The append-only constraint of the trigger engine and the change-control
+discipline of the standard are the same constraint. No runtime gap.
+
+Downstream re-derivation is scoped by `derived_from`: a node re-runs when any document it
+derived from has been superseded. That is a query, not a subscription.
+
+## Collections
+
+ECSS's own file taxonomy, one collection each:
+
+| Collection | ECSS file | Holds |
+| --- | --- | --- |
+| `RequirementsBaseline` | RB | One document **per requirement**, individually supersedable |
+| `TechnicalSpecification` | TS | Software requirements, derived from RB |
+| `InterfaceControl` | ICD | External interfaces |
+| `DesignDefinition` | DDF | Architecture and per-item design (SDD), build config (SCF) |
+| `DesignJustification` | DJF | Verification/validation specs and reports — the *why* |
+| `ManagementFile` | MGT | Plans, schedule, tailoring decisions |
+| `Rid` | — | Review item discrepancies, with disposition |
+| `SdlcJob` | — | Seed; carries issue number, criticality, head SHA |
+
+Every artefact document carries:
+
+- `derived_from: [doc_id]` — upstream provenance; the traceability matrix is `SELECT` over
+  this, never a maintained artefact
+- `supersedes: doc_id | null` — revision chain
+- `ecss_clause: string` — e.g. `5.4.3.1a`, the stable requirement id from the standard
+- `criticality: A | B | C | D`
+
+Granularity matters: **one document per requirement**, not one document per baseline. A human
+who disagrees with requirement 7 supersedes requirement 7, and only what derived from
+requirement 7 re-derives. Coarse documents would make every override re-run the whole V.
+
+## Nodes
+
+Lifted from ECSS-E-ST-40C clause 5. Each is a behavior whose write surface is exactly its
+expected outputs — the requirements node *physically cannot* write a `DesignDefinition`.
+
+| # | Clause | Node | Reads | Writes |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
```diff
@@ -0,0 +1,338 @@
...
+## Open questions
+
+1. **Fan-out width for `sdlc-item-design`.** `repo-maintenance` uses 5–10 areas by default.
+   A single GitHub issue probably wants fewer. Bound it, or let the architecture node declare
+   it?
+2. **Does the intake node get to argue?** If an issue is too vague to produce a requirements
+   baseline, does it fail the run, or post a comment asking for clarification and park until
+   a human replies (which is just an inbound RID)? The second is more useful and costs
+   nothing extra.
+3. **Re-derivation blast radius.** A superseded RB requirement invalidates everything
```
mike7c2: **NOODLING WARNING**

There is a subtle fork here. Above when we were talking about descending the V we are in the "pre approved artifact" regime i.e. this is just the back and forth for everyone to be comfortable with the project ideation. In that "pre approved artifact" mode the class 1 / class 2 discussed below don't actually exist.

**BUT** - I think we can use the same classification here in all honesty, at least for a first pass. It boils down to class 1 is a problem which ultimately needs the originator to fix it.

Which comes to how deeply and faithully we adopt the spec, for us the pre/post approval distinction isn't a big deal so I do think we can fold them together.

---------------------------------------------------------------------------------------------------------------------------------

We could adapt 4.3.3.4 which involves classifying changes - their classification is intended to capture the blast radius. In the main literature this is about whether we are handling it internally or going back to our customer, in our analogy here this is about whether we are going to go back to the originator (our issue) and try to get it refined, or whether it can be entirely mitigated within the stack owned artifacts.

Class 1 change - something that was significant, may result in changes to the initial artifact, goes to the customer / originator for review (this is the XY - your asked us to do this, you want that, the best way to get that is actually ....) .... this is "ask the meat"

Class 2 change - this is more like a small change in an intermediate artifact, this is a bump up a level on the left side of the V, essentially "the dude that wrote the plan didn't think about this and now we are trying to implement it's obvious we should change it, lets adjust the spec and come back". This is "do what we think is right and later tell the meat"

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
```diff
@@ -0,0 +1,338 @@
...
+- Paired-session review from desktop/mobile (documents already replicate; needs the
+  artefact-scoped behavior)
+- Packaging the pack as a `make` target consumable by *other* repos — the distribution story.
+  It is a separate concern, but building this pack first is what discovers its requirements.
+- 12207 organizational and agreement processes (§6.1, §6.2) — out of scope; this pack covers
+  the technical processes (§6.4) only.
+
+## Open questions
+
+1. **Fan-out width for `sdlc-item-design`.** `repo-maintenance` uses 5–10 areas by default.
```
mike7c2: How does this impact the design? I feel that this will strongly vary by task and that the design spec will imply a number of parallel areas that would set the fanout rather than us hardcoding one.

If a magic number is easier for first steps, 1-6 feels good to me. Annex K (K.4) calls out over and under selection as pathological modes and suggests some criteria that can be used to determine if a particular thing is worthy of its own area

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
```diff
@@ -0,0 +1,338 @@
...
+   A single GitHub issue probably wants fewer. Bound it, or let the architecture node declare
+   it?
+2. **Does the intake node get to argue?** If an issue is too vague to produce a requirements
```
mike7c2: Definitely yes! The intake node MUST argue.

The way I am visualising this, we feed something in at the top of the V

The first node digs in, it may (should!) come back with questions that help to improve the spec. It may take several turns to satisfy it, following the value gradient, this is the highest value excercise, "ideation refinement"

After the first node is able to complete its task without further input, we drop down to the next node, the same will happen again. With some luck, the parent can answer the subordinates questions and refine its artifacts from its own input, but there is also the possibility that they found a gap which needs to be defined in the ideation itself, so we can cascade back up the left side of the V at any point, any distance, when an agent encounters something not covered already. Note it always follows a cascade pattern, we never talk to a non adjacent node in the V.

(fwiw: 4.3.3.4: According to its effects, a change proposal or a departure request is processed through the different levels of the organization. The appropriate level to decide its disposition is the level for which the effects of the change or departure have no repercussions on the commitments made to the customer.)

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
```diff
@@ -0,0 +1,338 @@
...
+   baseline, does it fail the run, or post a comment asking for clarification and park until
+   a human replies (which is just an inbound RID)? The second is more useful and costs
+   nothing extra.
+3. **Re-derivation blast radius.** A superseded RB requirement invalidates everything
+   downstream of it. For a late override that could mean re-running code and verification.
+   Do we re-derive eagerly, or mark downstream documents stale and let a human trigger the
+   re-run?
+4. **Where does the SRR hard gate live for a `sdlc:D` fix?** Annex R may switch SRR off for
```
mike7c2: Heh ok :D

So this boils down to what exactly we are building _right now_

As posed I think this is an interesting pattern to pursue for "small app development" ( yo, make me a tool that does this thing ) to "feature integration" where I imagine the "loop" is eating about 2-4 weeks of human work and realistically is adding some feature to an established codebase.

( I think this may extend to larger development tasks but it's a bigger bet that I don't think we need to make right now - the quote I would point at that gives us some direction there is:  "The life cycle processes of this document can be applied concurrently, iteratively and recursively to a software system and incrementally to its elements.". I think we should call that recursion explicitly OOB today, but I think the pattern which can handle larger blocks of work will converge to V of Vs. )

The bounds are:
* Waiting for a human to answer blocks things
* Doing the wrong thing incurs another loop

So firstly, I think this is an on the fly decision every time, which is based on some risk assessment.

What's interesting about this is those two things can happen in parallel.

I wonder if we can make these harder decisions fork, basically we are going to log a "I made this decision" to make it obvious to a human the decision was made.

Depending on whether or not we are collapsing the classification model (3.) ... it smells like we want 3 classes in our own incarnation:

Class 1 - Must be escalated back to human for approval
Class 2 - A decision must be noted **AND** flagged for human review, work carries on (may be invalidated later by human, but we can optimistically take the agents decision in the moment)
Class 3 - Something resolved naturally within the stack by the agents that was entirely answered from artifacts and does not require human attention. Decision is still noted, but not elevated for attention.

.... this smells like a strong configuration parameter which is essentially going to configure how eargerly the stack approaches work 🤔

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
mike7c2: A way to look at this is that agents on the left of the V are aligned WITH the implementors AGAINST the specs / input artifacts and will argue if they are not good
( This lets the implementors say "HEY THIS IS A MORE EFFICIENT WAY TO REALISE THE SAME GOAL, UPDATE THE SPEC SO I CAN DO A CLEVER" )

Agents on the right of the V are aligned WITH the specs AGAINST the delivered artifacts, and will argue with the implementors

mike7c2 (commented):
This is frikking awesome ❤️ 

In particular I think the selected subset, and the shape of feeding (and refining) issues as an input and getting a PR as an output is solid. I'd also note that it is extremely pragmatically balanced in being faithful to the source material and tailoring to something that _could work today_

One thing I'd note, I think we can go "bigger" with this pattern to where we might be recursively running sub Vs under a parent V. At that point the gates that are currently human fall to the parent V. The human interacts with the top V and by extension gets much larger scoped PRs back (this is where the value of the change recording will 10x as we try to review larger bodies of work effectively). That's a much bigger bet though, we should keep it out of bounds for now.

I dropped some comments on the open questions, I don't have any objections to the material that's there

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
jackzampolin: Folded into 01be8353.

Fan-out is declared by `sdlc-architecture` as a property of the SDD, not a pack constant. v1 bound is 1–6. Annex K.4 is the selection criteria the architecture behavior uses (over- and under-selection as pathological modes).

See the Nodes section (architecture "declares items") and [Decisions from review](https://github.com/source-inc/gents/blob/01be8353d16881f088b95ca0bed66505c896a168/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md#decisions-from-review) §1.

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
jackzampolin: Folded into 01be8353, including the left/right alignment follow-up.

Intake must argue: a vague issue writes an inbound `Rid` via the GitHub bridge and parks; it does not fail the run. Cascade is adjacent-only — any distance up the left side, never skip-level — which is already what the write-surface fences enforce. Left-of-V agents are aligned with implementors against inherited specs; right-of-V agents are aligned with specs against delivered artefacts.

See [Argument, cascade, alignment](https://github.com/source-inc/gents/blob/01be8353d16881f088b95ca0bed66505c896a168/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md#argument-cascade-alignment).

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
jackzampolin: Folded into 01be8353.

Adopting §4.3.3.4 classification, with pre/post-approval folded together for v1. Three classes on the `Rid` under a job-level eagerness policy: (1) escalate and wait, (2) log + flag and continue, (3) note and do not elevate. Class 1 is the parked re-derivation path; 2 and 3 re-derive. This replaces the eager-vs-stale binary.

See [Change classification](https://github.com/source-inc/gents/blob/01be8353d16881f088b95ca0bed66505c896a168/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md#change-classification).

docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md
jackzampolin: Folded into 01be8353.

Current target is one V for a small-app / feature-integration job (~2–4 weeks of human work). Recursive V-of-Vs is explicit OOB and listed under deferred work. SRR stays a hard human gate at `sdlc:D` as an explicit MGT tailoring override of Annex R. The three-class model is the eagerness parameter — waiting and optimistic work can run in parallel under class 2.

See Prototype scope and [Decisions from review](https://github.com/source-inc/gents/blob/01be8353d16881f088b95ca0bed66505c896a168/docs/superpowers/specs/2026-08-20-agentic-sdlc-pack-design.md#decisions-from-review) §4.

jackzampolin:
Cross-linking #1314 as a likely implementation substrate for this design.

The ECSS activity DAG is exactly the kind of high-level workflow that should be expressed by instantiating operator-registered graph packages and connecting typed public results to entries. The SDLC pack should continue to own its domain semantics—RB/TS artifacts, RID gates, tailoring rules, verification versus validation, and criticality-selected topology—without owning the internal Tasks, triggers, or execution details of generic implementation, review, build/test, integration, and publication components.

This suggests a useful DSL layering:

- the SDLC/tailoring DSL selects and configures registered lifecycle components;
- GraphCompositionIntent resolves package instances, public ports, and approved adapters;
- raw GraphIntent remains the compiler IR and package-author surface;
- ordinary EventTriggers and GraphRun remain the execution substrate.

That also gives recursive V work a cleaner future boundary: recursion composes another registered SDLC package instance rather than dynamically authoring a lower-level graph. #1314 intentionally leaves recursive and unbounded cycles out of its initial release.

jackzampolin:
closing as a spike that will require more work.

PR #1193, opened 2026-08-25
feat: expose a model-callable graph compiler

jackzampolin:
## Summary

- expose one opt-in `compile_graph` custom tool that accepts only `GraphIntent`, compiles it against the caller DID and operator-owned capability catalog, and publishes the resulting routes atomically
- never accept an executable `GraphPlan`, Task prompt, behavior, tool selection, or arbitrary target collection from the model
- return stable, path-addressed diagnostics with no writes when compilation fails
- use generated `schemars` input schema and register the reserved tool name; operators can place it in `approval_required_tools`
- keep execution separate through existing bounded document writes after ordinary runtime reconciliation
- add checked-in evaluation cases for valid plans, invented capabilities, missing inputs, and structural bounds

The tool is currently an evaluation/programmatic custom tool; production document-config wiring is deliberately deferred until the experiment graduates.

## Review resolution

Claude, Grok, and Sol independently found that the earlier three-tool revision/run design let a model supply a self-hashed plan and raced trigger reconciliation. This stack removes that controller, removes GraphRun/GraphRevision entirely, and reuses existing Tasks, EventTriggers, config transactions, and the owned completion loop. The local GLM-5.2 review covered seven lenses and returned **APPROVE** after one documentation cleanup. A final post-fix GLM-5.2 review on workstation-1 covered eight lenses (authorization, publication, GraphQL safety, compiler correctness, formal conformance, schema migration, architecture reuse, and concurrency/durability) and returned **APPROVE** with zero candidates and zero confirmed defects. The rewritten stack is 2,714 added lines versus 4,372 before the review pass (38% fewer).

## Verification

- `cargo test -p gents` — all package unit, conformance, integration, and E2E suites passed
- `cd crates/gents/proofs && lake build`
- `git diff --check`

## Stack

1. #1190 — formal publication boundary and architecture
2. #1191 — pure typed compiler and conformance fence
3. #1192 — transactional EventTrigger publication over existing Tasks
4. **this PR** — single model-callable compile-and-publish tool and eval pack

### DIFF
--- a/demo/graph-pipeline/eval_cases.json
+++ b/demo/graph-pipeline/eval_cases.json
@@ -1,0 +1,262 @@
+[
+  {
+    "name": "valid_single_stage",
+    "intent": {
+      "graph_id": "document-review",
+      "nodes": [
+        {
+          "node_id": "worker",
+          "capability_id": "worker",
+          "capability_revision": "v1"
+        }
+      ],
+      "edges": [],
+      "entries": [
+        {
+          "name": "input",
+          "collection": "PipelineInput",
+          "schema": "PipelineInput/v1",
+          "to": { "node_id": "worker", "port": "input" }
+        }
+      ],
+      "limits": {
+        "max_nodes": 2,
+        "max_edges": 2,
+        "max_depth": 2,
+        "max_fan_out": 2
+      }
+    },
+    "expected_accepted": true,
+    "expected_diagnostic_codes": []
+  },
+  {
+    "name": "unknown_capability_fails_closed",
+    "intent": {
+      "graph_id": "document-review",
+      "nodes": [
+        {
+          "node_id": "worker",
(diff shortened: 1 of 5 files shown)


demo/graph-pipeline/eval_cases.json
```diff
@@ -0,0 +1,115 @@
+[
```
iverc, line 1: question: would it be beneficial to have more eval cases defined here?

jackzampolin: Yes. I expanded the checked-in pack from four to eight cases: a valid two-stage route plus repair cases for stale capability revisions, schema mismatch, and cycles. The exhaustive edge cases remain in the compiler unit suite and the generated 16-case Lean conformance matrix, so this pack stays focused on model proposal/repair behavior.

PR #1221, opened 2026-08-26
fix: normalize Grok OAuth stream usage

jackzampolin:
## Summary\n\n- normalize Grok OAuth streaming usage into the existing durable inference usage projection\n- retain usage across streamed response events so graph watch can display it\n- cover the provider event variants with focused tests\n\n## Stack\n\nDepends on the bundled runtime PR.\n\n## Verification\n\n- cargo test -p gents\n- cargo check --workspace --all-targets


jackzampolin:
Superseded while restructuring GitHub stack #1224 into independently compiling execution-contract, lifecycle-proof, durable-runtime, and bundled-package layers. The replacement stack preserves the feature content and makes the former core diff reviewable by ownership boundary.

Issue #564, opened 2026-06-26
EventSource/SubagentSource rescan completeness: paginate past the doc cap, bound the bridge query


jackzampolin:
Follow-up from the ultracode review of #561 (umbrella #551). #561 made EventDelivery D1 substantive and retired the rescan-gap deviations, but two runtime-completeness residuals remain — recorded as a Boundary in that PR rather than left silent.

## DH-1 — EventSource rescan is capped, completeness is not total
`crates/defra-agent/src/trigger_engine/event_source.rs` seeds and re-queries at most `SEEN_DOCS_SEED_LIMIT` (10_000) docs per source collection with no pagination/cursor (`event_source.rs:43,398,423,459`). For source collections larger than the cap, the tail beyond the first 10k is never surfaced by rescan and stays dependent on the lossy subscription path (the code's own `tracing::warn!` concedes this).

The Lean `rescanTick` transition models a **complete** rescan (dumps every unprocessed persistent doc), so `deviation := none` + substantive D1 assert more completeness than the runtime delivers at scale. #561 records this as `boundary.event-delivery.rescan-doc-cap` carrying the residual `missed_event_observation` mode. **This issue tracks eliminating it** (paginate past the cap, or a watermark/cursor), after which the Boundary can be removed.

v1 explicitly does not target catalog-scale source collections (`event_source.rs:41`), so this is not urgent — but it's the gap between the model's completeness claim and the runtime.

## R1 — SubagentSource bridge query is unbounded
`subagent_source.rs:382-405` `load_running_bridge_doc_ids` has no limit clause (unlike EventSource's cap). Self-limiting in practice (filtered to `lifecycle_state=running` AND non-empty `child_request_id`, i.e. live concurrency not history) but O(N) with no ceiling. Mirror the EventSource pattern: add an explicit limit + warn-on-hit.

## Done when
- EventSource rescan surfaces docs beyond `SEEN_DOCS_SEED_LIMIT` (pagination/cursor), and `boundary.event-delivery.rescan-doc-cap` is removed from `Boundaries.lean` + `coverage.rs`.
- SubagentSource bridge query has an explicit bound.

jackzampolin:
[...]

PR #1263, opened 2026-08-27
fix(runtime): bound subagent bridge recovery scan

jackzampolin:
## Summary

- cap the periodic running subagent bridge recovery query at 10,000 rows
- preserve the existing running/non-empty-child filters and response parsing
- warn when the query reaches its cap so delayed rows are observable
- cover the rendered recovery query and its explicit bound

Addresses #564 (R1 only). The EventSource pagination and Lean boundary work remain out of scope and the issue should stay open.

## Validation

- `git diff --check`
- `cargo test -p gents --lib trigger_engine::subagent_source::query_tests::running_bridge_rescan_query_is_explicitly_bounded -- --exact`
- `cargo test -p gents --test conformance subagent_source::recovery_materializes_orphan_child_request_for_running_subagent_tool -- --exact`
- `cargo test -p gents`

The workspace check retains the existing `gents-cli` dead-code warning for the secscan matcher description field.

### DIFF
--- a/crates/gents/src/trigger_engine/subagent_source.rs
+++ b/crates/gents/src/trigger_engine/subagent_source.rs
@@ -35,6 +35,21 @@ use super::{FireIntent, FireResult, TriggerKind, TriggerSource};
 
 const TOOL_CALL_COLLECTION: &str = "AgentToolCall";
 const SUBAGENT_SOURCE_RESCAN_INTERVAL: Duration = Duration::from_secs(5);
+const RUNNING_BRIDGE_RESCAN_LIMIT: usize = 10_000;
+
+fn running_bridge_rescan_query() -> String {
+    format!(
+        r#"{{
+            AgentToolCall(
+                filter: {{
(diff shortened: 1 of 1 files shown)


jackzampolin:
Independent review found a blocking liveness/foundation mismatch in this R1 interpretation. A hard cap on the unordered global bridge query can let 10,000 persistent ineligible rows indefinitely starve an eligible row after subscription loss. That also conflicts with the current EventDelivery boundary, which explicitly states SubagentSource is not capped, and with its no-deviation orphan-convergence contract. Marking this draft pending a decision: close this attempted slice, or authorize a foundation-backed fair cursor/pagination design. The larger design is intentionally outside this PRs approved scope.

jackzampolin:
Closing after independent review found that the naive hard cap can starve eligible bridge rows and contradicts the current SubagentSource completeness contract. The branch and worktree are being preserved as audit evidence; a safe replacement needs fair pagination/cursoring or authoritative scoping, foundation-first.

Issue #1217, opened 2026-08-26
Design: EVM transaction primitive — principal PKI and operator-declared contract capabilities


jackzampolin:
## Summary

Two related primitives, so agents can act on EVM chains without the runtime ever handing them a way to author an arbitrary transaction:

1. **Principal PKI** — how a gents principal gets a chain identity, where the key lives, and how a tool signs without the key entering agent context.
2. **Operator-declared contract capabilities** — the operator declares *(chain, contract address, ABI subset, per-parameter policy)*; the runtime generates a bounded tool from that declaration and does the ABI encoding in Rust. The agent supplies **typed arguments, never calldata**.

The second is the reusable one. Given it, specific on-chain actions become configuration rather than new Rust, the same way `BoundedWriteTool` made bounded DefraDB writes configuration rather than new Rust.

Motivating consumer is an autonomous trading agent (research and execution as separate least-privileged principals), but that's the driver, not the scope. Prior context: jackzampolin/amygdala#20.

---

## Why this shape

Most "governed agent wallet" designs in the wild — Project Aegis is the one we looked at closely — exist to constrain an agent that **can already author arbitrary transactions**, because their reference deployment hands the agent a terminal and a `.env` with the signer key in it. Everything downstream of that decision is compensation: decode whatever calldata the agent produced, figure out what it actually moves, and decide whether to allow it.

That decode step is where the bugs live. Aegis's own pre-deploy adversarial review found five issues, four of them in exactly that layer:

| Their finding | Root cause |
|---|---|
| C1 (critical) — destination allowlist **fail-open**: swaps, native sends, LP mints, Permit2 pulls never decoded the recipient | decoding arbitrary calldata |
| C3 (critical) — a paired approval inside a batch left a standing allowance; **the first fix was still bypassable** via a TOCTOU gap between approve and revoke | approval semantics |
| C2 (high) — Permit2 metered *how much* moved but not *where* | decoding arbitrary calldata |
| H1 (high) — Universal Router command mask `0x1f` folded high command IDs onto swap decoders, under-metering spend | decoding arbitrary calldata |
| F1 (low) — a legacy short-payload decode path defaulted to allow | decoding arbitrary calldata |

[...]

PR #1309, opened 2026-08-29
fix(eth): harden submit, expand explain, and add live GLM e2e

jackzampolin:
## Summary

Review follow-up on the EthTool stack (#1217 / stack #1308).

- `gents tools explain` and desktop explain expand `eth_tool_ids` the same way the live snapshot does
- Submit runtime: signing key must match `from`; nonce is `max(pending, in-flight journal+1)`; EIP-1559 fees are `2 * baseFee + tip`; `eth_call` simulates the estimated-gas transaction before sign
- `uintN`/`intN` values that do not fit the declared width fail closed
- Apply-time validation requires EIP-712 `domain.chainId` to match the EthTool chain
- Gas/fee cap misses are `PolicyDenied`
- Live e2e: GLM-5.3-Flash-NVFP4 on workstation-1 uses `base-sepolia_query` against live Base Sepolia

## Stack

1. #1299 schema
2. #1300 keys
3. #1301 RPC
4. #1302 query tool
5. #1303 reads
6. #1304 submit
7. #1305 writes
8. #1306 typed data
9. **this PR** hardening + live e2e (`eth-wallet-hardening` → `eth-wallet-typed-data`)

Part of #1217.

## Test plan

- [x] `cargo test -p gents --lib eth::submit::tests`
- [x] `cargo test -p gents --lib eth::call_tool::tests`
- [x] `cargo test -p gents --lib eth::calls::tests`
- [x] `cargo test -p gents --lib tool_surface::tests`
- [x] `cargo test -p gents-cli --lib desired_state`
- [x] `GENTS_ETH_LIVE=1 cargo test -p gents --features live-e2e --test e2e_live eth_tool_live_model_queries_base_sepolia -- --ignored --test-threads=1 --nocapture`


crates/gents/src/eth/submit.rs
```diff
@@ -389,14 +404,94 @@ async fn persist_receipt(
...
+
+async fn max_in_flight_nonce(
+    node: &EmbeddedNode,
+    from: &str,
+    chain_id: u64,
+) -> Result<Option<u64>> {
+    let from = escape_graphql_string(&from.to_ascii_lowercase());
+    let query = format!(
+        r#"{{
+            EthSubmission(filter: {{
```
jackzampolin: **[suggestion]** `max_in_flight_nonce` loads every `EthSubmission` for `(from_address, chain_id)` and filters `prepared` / `submitted_unknown` in memory. There is no status `_in` filter, no `order`/`limit`, and invalid nonces are dropped with `u64::try_from(...).ok()`. DefraDB does not apply a default page size here (unlimited queries return all rows), so this is correct while the set is small, but every write now scales with lifetime submission history for that key. A corrupt/negative nonce is also fail-open (ignored rather than aborting nonce allocation).

**Suggestion:** Filter in GraphQL (`status: { _in: ["prepared", "submitted_unknown"] }`), then take max nonce. Optionally `order: { nonce: DESC }, limit: 1`. Treat a negative nonce as an error instead of skipping it.

crates/gents/src/eth/rpc.rs
```diff
@@ -149,7 +149,7 @@ impl<T: JsonRpcTransport> EthRpcClient<T> {
     #[allow(dead_code)] // Wired into generated write tools in the next stack commit.
     pub(crate) async fn simulate_transaction(&self, transaction: Value) -> Result<Value> {
         self.ensure_chain_id().await?;
-        self.rpc("eth_call", json!([transaction, "latest"])).await
+        self.rpc("eth_call", json!([transaction, "pending"])).await
```
jackzampolin: **[suggestion]** Simulate switched from `latest` to `pending`. That is the right tag for sequential in-flight state, but several public L2 endpoints treat `pending` as an unknown block. `is_pruned_history_message` already classifies `"unknown block"` / `"header not found"` as non-retried pruned-history errors, so writes would fail closed on those RPCs before sign. The new `simulate_call_includes_estimated_gas` test only checks that `gas` is present, not the block tag, nonce, or fee fields. The `#[allow(dead_code)]` on `simulate_transaction` is also stale now that submit calls it.

**Suggestion:** Keep `pending` if you have evidence the target RPCs support it; otherwise try `pending` and fall back to `latest` on unknown-block errors. Extend the scripted test to assert `params[1] == "pending"` plus `nonce` / `maxFeePerGas` / `type`. Drop the dead_code allow and the outdated comment.

crates/gents/src/tool_surface/behavior_config.rs
```diff
@@ -117,13 +118,18 @@ impl BehaviorToolConfig {
         behavior_name: &str,
         selection: &crate::document_config::ToolSelectionDocument,
         surfaces: &[crate::document_config::DatastoreToolSurfaceDocument],
+        eth_tools: &[crate::document_config::EthToolDocument],
         ceiling: &ToolCeiling,
         custom_tools: Vec<CustomToolFactory>,
     ) -> Result<Self> {
         let merged = crate::document_config::merge_datastore_tool_surfaces(selection, surfaces)?;
+        let expanded =
+            crate::agent::document_view::expand_eth_tools_from_docs(selection, eth_tools)?;
```
jackzampolin: **[suggestion]** Explain expansion is the right behavior, but it is implemented by making `agent::document_view` `pub(crate)` and calling it from `tool_surface`. `agent` already depends on `tool_surface`, so this is a new intra-crate cycle and it punches explain/CLI/desktop through agent internals. `from_tool_selection_document` now always expands with `&[]` and fail-closes if `eth_tool_ids` is set, which is correct for the dedicated constructor but easy to miss for future callers of the convenience wrapper.

**Suggestion:** Move `expand_eth_tools_with` / `expand_eth_tools_from_docs` to `crate::eth` or `document_config` and keep `document_view` private.

crates/gents/src/eth/call_tool.rs
```diff
@@ -934,6 +935,19 @@ fn dyn_to_json(value: &DynSolValue) -> Value {
     }
 }
 
+fn integer_bit_width(ty: &str, signed: bool) -> Option<usize> {
+    let prefix = if signed { "int" } else { "uint" };
+    let suffix = ty.strip_prefix(prefix)?;
+    if signed && ty.starts_with("uint") {
```
jackzampolin: **[nit]** `if signed && ty.starts_with("uint")` is dead. `integer_bit_width(..., true)` uses `strip_prefix("int")`, which already returns `None` for any `uint*` type, so the guard never runs. The helper also duplicates `integer_kind` in `calls.rs` (same 8..=256 / multiple-of-8 rules).

**Suggestion:** Delete the dead guard. Parse width once (share with `integer_kind`) so ABI encoding and declaration validation cannot drift.

PR #1400, opened 2026-09-05
feat(backend): Claude model discovery via GET /v1/models, and discover-models --write

edjroz:
Supersedes #1390 (same content; re-opened from a `gents-ai/gents` branch so GitHub's native stacked-PR view applies).

## Summary

- Claude arm in `discover_models`: `GET https://api.anthropic.com/v1/models?limit=100` with the credential bearer,
  `anthropic-version: 2023-06-01` and `anthropic-beta: oauth-2025-04-20`; identifiers trimmed (the same `identifier()` trim now
  applies to Codex catalog records).
- `gents config backend discover-models --write` persists the discovered ids into the backend's `models[]` — for all three
  OAuth kinds. Requires `--backend-id`; a models-only mutation (name, endpoint, api_key, max_concurrent, enabled,
  probe_status untouched); skipped with `write_skipped` when discovery returns zero ids, so `models[]` is never wiped.

Commits (stacked on PR2): 7ce34b6a (discovery arm + CLI error/credential arms + help text), 28237cb9 (`--write`).

## Notes

- This is the first path that ever persists a catalog: before it, discovery only printed, and `gents init` seeds `models[]`
  with the single `--model-name`. Codex's `/model` picker reads `models[]`, so this is what makes the full Claude list appear.
- Discovery is explicit only: no login runs it, and automatic (fleet) discovery keeps skipping agent-scoped kinds.
- `--write` bypasses `InferenceBackend::validate` (same as `backend set` passing `None`); the no-lockout conjunct is opt-in.
- `has_more` is ignored: a catalog over 100 models would be truncated.
- A hand-edited `https://` endpoint on a Claude row becomes the discovery base; the placeholder `claude-cli://subscription`
  maps to the default base.
- One GET per invocation, no inference.

## Verification

- Opus review (with PR2's trees): approved, 0 Critical / 0 Important.
- Unit: `backend_provider` 10/10 (mock `/v1/models` three ways; path, `?limit=100`, both headers, trimming asserted);
  `cli_config_backend` discover tests incl. zero-ids and the `--write`-without-`--backend-id` guard. On this exact tip
  (2026-09-05 split gate): `cargo check` 0; `backend_provider` lib tests 10/10; gents-cli `discover` tests 6/6.
- Live (dev machine): `GET /v1/models` with the subscription bearer → 200, 11 ids; `--write` turned `models[]` from 1 to 11
  with every other field byte-identical, idempotent on rerun, no restart needed.

## Known gaps / follow-ups (not in this PR)

- Two Anthropic base-URL literals; duplicated auth-guidance arms across the three kinds.
- `--write` not validated (above); `has_more` truncation.

### DIFF
--- a/crates/gents-cli/src/cli/args.rs
+++ b/crates/gents-cli/src/cli/args.rs
@@ -2406,6 +2406,11 @@ pub(crate) struct BackendDiscoverModelsArgs {
         help = "Agent home directory used to resolve the local agent DID for OAuth-credential backend discovery (defaults to ~/.gents). Pass --agent-did instead to target a specific agent"
     )]
     pub(crate) home: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "Replace the backend document's models[] with the discovered models (requires --backend-id; nothing is written without this flag)"
+    )]
+    pub(crate) write: bool,
 }
 
 #[derive(clap::Args)]
--- a/crates/gents-cli/src/cli/args.rs
+++ b/crates/gents-cli/src/cli/args.rs
@@ -2398,12 +2398,12 @@ pub(crate) struct BackendDiscoverModelsArgs {
     pub(crate) api_key_env_var: Option<String>,
     #[arg(
         long,
-        help = "Agent DID owning the ChatGptCodex OAuth credential (defaults to the local agent). Only used for ChatGptCodex backends, whose bearer is a DefraDB document rather than an api_key"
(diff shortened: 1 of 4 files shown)


jackzampolin:
Restacked onto current `main` (includes #1414–#1419).

- #1398 `claude/track-b/pr1-messages-wire-lean` @ 4ab7b58b
- #1399 `claude/track-b/pr2-oauth-login-probe-gate` @ 74fad6a1
- #1400 `claude/track-b/pr3-model-discovery` @ 843ee9c4 (tip)

Kept the cleanup's `CLAUDE.md` → `AGENTS.md` symlink and did not restore deleted `docs/`. Claude credential sentence is in `AGENTS.md`. `upsert_oauth_credential` still goes through `ConfigAccess::write_local`.

#1401 (docs tip) closed. Usage-limit 429 UX filed as #1422 — not in this stack.

PR #1444, opened 2026-09-09
fix: avoid redundant client startup work

jackzampolin:
## Summary

Fifth layer above #1442. Remove two redundant startup costs without changing sync ownership or readiness:

- Run the complete schema migration/verification/materialization engine once, instead of calling it through two equivalent wrappers.
- Subscribe the existing permitted collection set with one native DB batch. The DB already treats existing subscriptions idempotently. This avoids repeated authorization checks and full subscription-list persistence.
- Emit debug-level startup phase timings with no document content.
- Assert a single exact-scope subscription batch and fail-closed error handling.

The privacy/local-only exclusions, schema lineage checks, observer ordering, pairing authority, and recovery/streaming test budgets are unchanged.

## Measurements

Local aged-history diagnostic samples reduced duplicate migration work by 245 ms, collection subscription setup from 564 ms to 237 ms, and total client startup from 1661 ms to 1044 ms. Concurrent builds mean these are diagnostic samples, not isolated benchmarks.

## Stack and validation

Ready for review. The merged native DB/Regolith dependencies are supplied by tip #1450. The full six-layer candidate on current `main` passes the complete runtime suite, all-target workspace check, 227 desktop-core tests, and all six canonical enrollment cases, including aged history and offline recovery. No local dependency paths, schema changes, or rollout.

### DIFF
--- a/crates/gents-desktop-core/src/client/core/bootstrap.rs
+++ b/crates/gents-desktop-core/src/client/core/bootstrap.rs
@@ -38,10 +38,21 @@ impl ClientCore {
         paths: DesktopPaths,
         options: ClientCoreOptions,
     ) -> Result<Self> {
+        let started = Instant::now();
+        let mut previous = started;
+        let mut checkpoint = |stage: &'static str| {
+            let now = Instant::now();
+            tracing::debug!(target: "gents_desktop_core::startup", stage,
+                stage_ms = now.duration_since(previous).as_millis(),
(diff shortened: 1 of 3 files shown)


jackzampolin:
Stack: #1438 → #1439 → #1441 → #1442 → #1444. Local runtime, workspace/all-target, and 227 desktop-core tests pass. Related storage improvement: sourcenetwork/regolith#194. No deployments or merges are part of this PR.

PR #1453, opened 2026-09-09
gents-loop: the completion loop as a library crate, buildable for wasm32-wasip1

vertexclique:
Gents Cloud runs a gent's model-facing logic as a sealed `wasm32-wasip1` guest, one turn per instantiation. That needs the completion loop to exist without sockets, without a filesystem and without the native tool surface, while staying the same loop the laptop runs.

This branch carves it out:

- `crates/gents-loop`: the owned loop behind three seam traits (`SessionHook`, `StreamWriter`, `RequestLifecycleControl`), with `run_loop_stream<M: CompletionModel, H: SessionHook>` as the entry point. No I/O of its own.
- `gents` builds on it and re-exports what it already exposed, so nothing downstream moves.
- The feature gates that let the seam compile for `wasm32-wasip1`.

What consumes it: `packages/gent` in gents-cloud compiles this loop into a content-addressed `.afb`, pinned by digest, executed under fuel, memory and wall-clock bounds. The cloud's supervisor replays the guest's effects as the agent's own DID, so the guest never writes a document itself.

Verified: the workspace builds and its tests pass; the wasm side is proven in gents-cloud, where a turn runs end to end through the real fiber against a scripted model, streams its tokens, dispatches a tool call, and lands the same effects hash on a repeat run.


.github/scripts/compile-rust-ci-suite.sh
```diff
@@ -13,7 +13,11 @@ separate_package=""
 
 case "${suite}" in
   runtime)
-    packages=(gents)
+    # gents-loop is the completion loop `gents` is built on, extracted so it
```
jackzampolin: **Suggestion:** The new comment says `gents-loop` is extracted so it compiles for `wasm32-wasip1` and that "the runtime step below runs" its tests. This script only native-compiles (`cargo test --all-targets --no-run`). `setup-rust.sh` still installs only `wasm32-unknown-unknown`. PR runtime smoke tests never execute `-p gents-loop`. Full `cargo test -p gents-loop` is main-only. The crate does compile for wasip1 today, but the PR's defining claim is unenforced and the comment overstates what CI does.

**Fix:** Add `rustup target add wasm32-wasip1` and `cargo check -p gents-loop --target wasm32-wasip1` (no `native` feature) on the runtime or support shard. On PRs, run `cargo test -p gents-loop --lib` — those 129 tests are DefraDB-free. Trim the comment to match.

crates/gents-loop/src/compaction.rs
```diff
@@ -0,0 +1,1266 @@
...
+            now: Arc::new(chrono::Utc::now),
+        }
+    }
+
+    pub fn with_summary_output_limit(mut self, limit: usize) -> Self {
+        self.summary_output_limit = Some(limit);
+        self
+    }
+
+    pub fn with_backend_id(mut self, backend_id: String) -> Self {
```
jackzampolin: **Suggestion:** After the dyn-compat port, `ProviderReductionEngine::reduce` never reads `backend_id`. Admission scoping now lives only on `gents::compaction::backend_scoped_reduction_engine`. The sole caller still chains `.with_backend_id(backend_id.clone())` and then wraps the engine, so the setter looks load-bearing and is a no-op. A later caller that copies the old `with_backend_id` pattern would silently skip `admission::scope_backend`.

**Fix:** Remove `backend_id` / `with_backend_id` / `backend_id()` from the guest engine, and drop the `.with_backend_id(...)` call in `completion_factory.rs`. Keep scoping exclusively on the native wrapper.

crates/gents-loop/src/lib.rs
```diff
@@ -0,0 +1,145 @@
+//! The gent's owned completion loop, carved out of `gents` (G-1) so it
+//! compiles for `wasm32-wasip1`: the guest half of the H18 split (see
+//! `docs/gents-cloud-v1.md` §15.0). Everything here is provider-input
```
jackzampolin: **Suggestion:** Crate docs point at `docs/gents-cloud-v1.md` §15.0. That path does not exist in the tree (`docs/` is absent). The H18/guest-loop claim is otherwise accurate (no `std::fs`/`std::net`/`println!`/DefraDB/thread-spawn in production `gents-loop` code; `tokio::spawn` is test-only).

**Fix:** Drop the dangling path, or land the doc before citing it.

crates/gents-loop/src/compaction.rs
```diff
@@ -0,0 +1,1266 @@
...
+
+// pub, not private: gents' own compaction test suite reaches these through
+// `crate::compaction::{history, summary}` (a private submodule in this crate
+// is invisible to a dependent crate's own test build).
+#[path = "compaction_history.rs"]
+pub mod history;
+#[path = "compaction_summary.rs"]
+pub mod summary;
+#[cfg(test)]
+mod tests {
```
jackzampolin: **Suggestion:** An empty `#[cfg(test)] mod tests` exists only to narrate that tests stayed in `gents` and to point at "the gents-loop crate report". It restates the move instead of explaining a constraint.

**Fix:** Delete the empty module. If a pointer is needed, one line on the `pub mod history` visibility comment is enough.

crates/gents/src/home.rs
```diff
@@ -0,0 +1,180 @@
+//! A gents home's persisted `init.json` and its well-known paths.
+//!
+//! A "gents home" is a directory (`~/.gents` by default) holding
+//! `init.json` (this module's [`StoredInitConfig`]), the agent's identity
+//! key, and the embedded DefraDB data directory. This module is the
+//! single, canonical place that shape and those paths live: `gents-cli`
+//! reads and writes them here (`gents-cli/src/home_state.rs`), and any
+//! other crate that provisions a gents home from scratch -- gents-cloud's
```
jackzampolin: **Suggestion:** New `gents::home` module docs spend a page on gc-cell, clap `ValueEnum` bounds, and "before this module existed". That is design history, not a constraint. Unrelated to the loop extraction.

**Fix:** Keep the serde `bound(deserialize = ...)` comment (that one is a real WHY). Cut the module docs to: home path helpers and `init.json` shape, shared by CLI and other provisioners.

crates/gents/src/hook.rs
```diff
@@ -1137,6 +1138,111 @@ impl DefraSessionHook {
...
+            .await
+    }
+
+    async fn persist_stream_tool_result_progress(
+        &self,
+        tool_result: &ToolResult,
+        internal_call_id: &str,
+    ) -> anyhow::Result<bool> {
+        self.persist_stream_tool_result_progress(tool_result, internal_call_id)
+            .await
```
jackzampolin, line 1219: **Nit:** Most `SessionHook` methods use `DefraSessionHook::foo(self, ...)` UFCS. `persist_stream_tool_result_progress` and `persist_inflight_assistant_turn` call `self.foo(...)`. Inherent methods in `hook/persistence/message_spawn.rs` currently win, so this is not recursion — but it is the one pattern that would infinite-loop if those inherent methods were renamed or removed.

**Fix:** Use the same UFCS as the neighboring methods.

PR #1454, opened 2026-09-10
Packs and plugins, built, published and installed from the registry

vertexclique:
## Onboarding stack integration — 2026-09-15

Immediate parent: #1510 (`perf/onboarding-chat-responsiveness`), above provider #1509. Pack slot/provenance work will target this branch; configurator ergonomics will target that pack layer. This integration preserves the original registry commits and the current DefraDB/Regolith pins.

Integration review fixes:
- Verify requested namespace/name/version against downloaded pack/plugin manifests before installing them under a coordinate.
- Retain the canonical PackPlugin declaration across installation and execution; do not replace authored metadata/authority with artifact-derived guesses.
- Remove the unrelated production config-apply sleep and use tracing for package resolution.

Model-facing registry/slot adapters remain explicit follow-up work in the configurator layer. Registry graph installation is still explicitly refused by this foundation; no end-to-end model-facing registry support is claimed.

Current integrated validation: workspace check passed before review fixes; full runtime and focused registry/plugin checks in progress. The detailed validation below is historical evidence from the original branch, not a claim that it has all been rerun on the stack.

---

Packs could only ever be the ones compiled into the binary. `gents pack list` showed what shipped with the build, `gents pack install` could install nothing else, and a customer who wrote a pack had no way to give it to anyone. This makes a pack something you build, publish and install from a registry, and it makes the capabilities a pack carries a first-class thing of their own.

## Two artifacts, and the line between them

**A plugin is a complete Afterburner `.afb`.** One compiled module that runs sandboxed, publishable and installable on its own, and runnable by name. `gents plugin` is its whole surface.

**A pack is a `.tar.gz`.** It holds `manifest.json`, the assets that manifest declares, and, under `plugins/`, the compiled `.afb` files for the plugins it carries. Installing a pack installs its plugins into the same store `gents plugin install` uses, so a plugin that arrived inside a pack is as runnable by name as one installed alone.

The split is the point. A pack is a bundle of things; an `.afb` is one program. An earlier revision of this branch made the pack itself an `.afb`, which meant nesting an `.afb` inside an `.afb` and reading a pack manifest out of the member set meant for a program's own source. The tar.gz says what it is, and any tool can open it.

```text
manifest.json          the pack manifest, including its plugin declarations
<asset>                every asset the manifest declares
plugins/<name>.afb     one compiled Afterburner package per plugin
```

Two properties, both tested:

**A pack is the same pack wherever it came from.** The digest is over the declared contents, never the container, so neither the route a pack took nor the compression it arrived under changes its identity.

**Packing is reproducible.** The same pack directory packs to the same bytes, so a published pack can be compared rather than trusted.

## What you can do now

```sh
gents pack build packs/shipping            # compile its plugins, write one .tar.gz
gents pack build --all                     # every pack in this repository
gents pack publish shipping-0.1.0.tar.gz   # push it to the registry
gents pack search formatting
gents pack install shipping                # bundled if it is, otherwise from the registry

gents plugin build ./format_check          # compile one plugin on its own
gents plugin publish format_check-0.1.0.afb
gents plugin install acme/format_check
gents plugin list
gents plugin run format_check '{"path":"src"}'
gents plugin remove acme/format_check
```

`pack install` resolves bundled first, so today's behaviour does not change, then the registry. A download is checked against the digest the registry advertised **before it is opened**, and a mismatch names both digests rather than warning. Fetched artifacts are cached content-addressed under the home, so a second install does not refetch. The default registry is `https://packs.gents.xyz`, overridable with `--registry` and `GENTS_REGISTRY`.

## Every language Afterburner compiles, and each one proven

`pack build` and `plugin build` share one function, so the CLI cannot accept a language on one path and refuse it on the other. All eight compile, and each has a test that compiles a real package and inspects the artifact:

| language | shape of the artifact |
| --- | --- |
| Rust, Go, C, C++, JS, TS | `wasm32-wasip1` WASI command |
| Ruby | `wasm32-wasip1` WASI command, via wasi-vfs |
| Python | a self-contained `emscripten-pyodide` bundle |

Python is the one that is not a WASI command, and it is the one this branch spent the most care on. It was refused here for a while, on the grounds that its bounds were not enforceable. That reason stopped being true: the wall clock was wired upstream (afterburner-sh/afterburner#65), and a compiled bundle now runs under fuel, memory, wall clock and stdin like everything else. The refusal is gone with it, and a Python plugin is built, published, installed and run in the transcript below.

## How a plugin runs, and what it cannot do

`gents::plugin::PluginRunner` runs one in-process on Afterburner's own runtime. The ABI is deliberately narrow: canonical JSON arguments on stdin, one JSON value on stdout, stderr is diagnostics. Nothing else crosses.

- **One run per instantiation.** A plugin keeps nothing between calls.
- **Effective authority is an intersection, never a union.** What the plugin declared, narrowed by what the caller allows. A wide ceiling grants nothing the plugin did not ask for.
- **A plugin never listens.** Structurally (a WASI command has no socket import) and again by explicit refusal at validation.
- **A plugin whose artifact cannot be run with every bound enforced is refused at admission**, naming the axes that would be dropped, rather than run with a bound silently missing. The classification comes from `afterburner::afb_run::bounds_for`, asked rather than copied: a language name cannot answer it, since Ruby compiles to a fully bounded WASI command while Ruby source enforces nothing.
- **Every bound is its own verdict.** Fuel, memory and wall clock are distinguishable outcomes, not a generic failure. Output and diagnostics are bounded and truncation is stated where it shows.
- **Output is validated.** Stdout that is not a single JSON value is `BadOutput` with the reason, never a silent empty result.

## Verification

Run on this branch, in debug:

```
cargo test -p gents --lib pack                      44 passed, 0 failed
cargo test -p gents --lib plugin                    21 passed, 0 failed
cargo test -p gents-cli --lib commands::plugin      14 passed, 0 failed
cargo test -p gents-cli --lib commands::           713 passed, 0 failed
```

`cargo fmt --check` is clean. `cargo clippy` reports two errors on this branch, both in test files it does not touch (`crates/gents-cli/tests/suites/cli_trace_export.rs` and `crates/gents/tests/conformance/durable_reduction.rs`); they are on `main` as well and are left alone rather than folded into this change.

The language tests compile a real package per language through the real compiler; the plugin tests build a real Rust plugin and run it. No fixture stands in for either.

## The whole loop, against a running registry

Not a unit test: `gents` built from this branch, against a registry process on `127.0.0.1:8899`.

```
$ gents plugin build ./doubler --out /tmp/doubler.afb
✓ compiled acme/doubler@0.1.0 (precompiled wasm32-wasip1)
  digest sha256:04e104d66b348e0d0006ba8c00e25f4534e4a6fb11f3a5bf4071e0db524ddd31  (24015 bytes)

$ gents plugin publish /tmp/doubler.afb
{ "namespace": "acme", "name": "doubler", "version": "0.1.0",
  "registry_response": { "kind": "plugin", "size_bytes": 24015,
    "digest": "04e104d66b348e0d0006ba8c00e25f4534e4a6fb11f3a5bf4071e0db524ddd31" } }

$ gents plugin install acme/doubler
{ "namespace": "acme", "name": "doubler", "version": "0.1.0",
  "digest": "sha256:04e104d66b348e0d0006ba8c00e25f4534e4a6fb11f3a5bf4071e0db524ddd31" }

$ gents plugin run acme/doubler --input '{"n":21}'
{ "doubled": 42, "input": 21 }
```

And the same plugin carried inside a pack, which is the other half of the design:

```
$ gents pack build ./shipping --out /tmp/shipping.tar.gz
{ "pack": "shipping", "namespace": "acme", "version": "0.1.0", "size_bytes": 24641,
  "pack_digest": "sha256:75b54ec82664e9eee14ccb104dbc4fe39394aef063bedd6cef3d4029b14f7675",
  "plugins": [ { "name": "doubler", "artifact": "plugins/doubler.afb" } ] }

$ tar tzf /tmp/shipping.tar.gz
README.md
manifest.json
plugins/doubler.afb

$ gents pack publish /tmp/shipping.tar.gz
{ "kind": "pack", "namespace": "acme", "name": "shipping", "version": "0.1.0" }

$ gents pack install acme/shipping
gents pack install: resolved acme/shipping from registry (acme/shipping@0.1.0)
{ "pack": "shipping", "source": "registry",
  "digest": "sha256:75b54ec82664e9eee14ccb104dbc4fe39394aef063bedd6cef3d4029b14f7675",
  "installed_plugins": [ { "namespace": "acme", "name": "doubler", "version": "0.1.0",
    "digest": "sha256:04e104d66b348e0d0006ba8c00e25f4534e4a6fb11f3a5bf4071e0db524ddd31" } ] }

$ gents plugin run acme/doubler --input '{"n":50}'
{ "doubled": 100, "input": 50 }
```

The plugin's digest is identical whether it was installed on its own or arrived inside the pack, and the pack's install placed it under the pack's namespace rather than a default one.

And the language that used to be refused, through the same loop:

```
$ gents plugin build ./pydoubler --out /tmp/pydoubler.afb
{ "namespace": "acme", "name": "pydoubler", "version": "0.1.0", "language": "python",
  "size_bytes": 4639228,
  "digest": "6b4b93f80accedceab3bbe442309ea8190361a9cc189b247a50fc29af33dcbac" }

$ gents plugin publish /tmp/pydoubler.afb && gents plugin install acme/pydoubler

$ gents plugin run acme/pydoubler --input '{"n":21}'
{ "doubled": 42, "input": 21 }
```

That last one took two more fixes to be true rather than merely allowed. A Pyodide bundle carries CPython, which will not instantiate under the default 64 MiB memory ceiling and spends far more than the default 100 million instructions before the plugin's first line. Both failed honestly (a trap, then an exhausted fuel budget) but a language that is admitted and then cannot answer is not supported. A default budget is now taken from the artifact's own startup floor, which Afterburner reports (`afb_run::startup_floor`) because that is where the dispatch shape is known.

## What that run found, which no unit test had

Running it end to end is what turned up four defects that every test in the tree was passing over:

- **`gents plugin run` panicked before the guest produced a byte.** Running a guest is synchronous and the WASI runtime under it blocks on a runtime of its own, so calling it from the async command dispatcher aborted with "cannot start a runtime from within a runtime". It runs on the existing blocking pool now.
- **A pack could be built here and refused by the registry.** The manifest had no namespace field at all while the registry required one, so no pack this branch built could be published. `namespace` is a manifest field now, defaulting to `gents`, and both readers use that default.
- **Every published pack silently lost its tags**, because the registry indexed `keywords` and the manifest writes `tags`.
- **`gents pack install` asked the plugin route for a pack.** The registry serves the two kinds under different paths; the client had one set of routes for both.
- **A pack-carried plugin was recorded under a default namespace**, so two packs from different namespaces carrying a same-named plugin would have silently replaced each other.


crates/gents-cli/src/commands/pack/registry.rs
```diff
@@ -0,0 +1,697 @@
...
+                     hand with `{} -o {name}-{version}.tar.gz` and install that file",
+                Self::curl_equivalent(&url, false)
+            )
+        })?;
+        let status = response.status();
+        anyhow::ensure!(
+            status.is_success(),
+            "downloading {namespace}/{name}@{version} from the registry failed ({status})"
+        );
+        Ok(response
```
jackzampolin, line 203: **Bug:** `RegistryClient::download` buffers the whole response with `response.bytes().await` and no size cap, `Content-Length` check, or client timeout (`reqwest::Client::new()`). `PackArchive::from_bytes` enforces `MAX_PACK_BYTES` (64 MiB) only *after* the body is already in memory, so a hostile or MITM registry can OOM the CLI. Plugin install uses this same unbounded download.

**Fix:** Stream with a hard cap (pack: `MAX_PACK_BYTES`; plugin: Afterburner's AFB cap) and abort when `Content-Length` exceeds it or when more than the cap is read. Build the reqwest client with connect/read timeouts.

crates/gents-cli/src/commands/pack/registry.rs
```diff
@@ -0,0 +1,697 @@
...
+                     request by hand is: {}",
+                Self::curl_equivalent(&url, false)
+            )
+        })?;
+        Self::json_or_error(response, &url).await
+    }
+
+    pub(crate) async fn package(&self, namespace: &str, name: &str) -> Result<Value> {
+        let kind = self.kind.path();
+        self.get_json(&format!("/{kind}/{namespace}/{name}")).await
```
jackzampolin, line 149: **Bug:** Namespace, name, version, and the advertised digest are interpolated into URLs and filesystem paths with no single-segment validation or percent-encoding. `split_namespace` only splits on the first `/`, so `acme/foo/bar` becomes `GET /api/v1/packs/acme/foo/bar`, which matches the version route `{ns}/{name}/{version}` rather than the package route. A `..` segment can normalize onto the other artifact surface (`/packs/x/../packages/...`), reopening the pack-vs-plugin route mix-up this PR otherwise fixed. `fetch_pack` also does `cache_dir.join(format!("{advertised}.tar.gz"))` before requiring the digest to be 64 hex, so a registry-supplied `../...` digest is a path join, not a cache filename.

**Fix:** Reuse the plugin store's `checked_component` (or `is_valid_pack_name`) on namespace and name before any HTTP call; reject `/`, `..`, empty, and non-snake_case segments. Percent-encode path segments. Require the advertised digest to be 64 lowercase hex before using it as a filename.

crates/gents-cli/src/commands/pack/mod.rs
```diff
@@ -29,169 +32,105 @@ pub(crate) async fn dispatch(command: PackCommand) -> Result<()> {
...
-            )
-            .unwrap_or_else(|error| panic!("{}: {error:#}", manifest.name));
-            gents::config_client::DesiredStateApplyPlan::from_pack_config(&config)
-                .unwrap_or_else(|error| panic!("{}: {error:#}", manifest.name));
+/// Resolves a pack compiled into this binary first; only when that fails
+/// does it fall back to the registry, downloading, verifying, and caching
+/// the result. Both failures are reported together so a real problem with
+/// the bundled lookup is never masked by a registry error.
+async fn resolve_pack_source(name: &str, registry_override: Option<&str>) -> Result<PackSource> {
+    match resolve_pack(name) {
```
jackzampolin, line 110: **Bug:** `resolve_pack_source` only hits the bundled catalog when `resolve_pack(name)` succeeds on the raw CLI token. Bundled packs are registered as bare names (`mailbox`), so `gents pack install gents/mailbox` (the coordinate the README documents) fails bundled lookup, falls through to the registry, and for a graph pack downloads then refuses with "only a graph pack compiled into this binary can be installed". Bare `mailbox` never sees a newer registry version either. `--home` is also ignored here (`resolve_home_dir(None)`), so a registry-backed install always writes `~/.gents/packs/registry-cache` even when the operator passed `--home`.

**Fix:** After `split_namespace`, retry `resolve_pack(pack_name)` when the namespace is `DEFAULT_NAMESPACE` (or matches the bundled pack's namespace). Pass the install `--home` into `fetch_pack`.

crates/gents-cli/src/commands/pack/mod.rs
```diff
@@ -372,19 +319,25 @@ async fn install(args: PackInstallArgs) -> Result<()> {
...
             );
+            anyhow::ensure!(
+                matches!(pack, PackSource::Bundled(_)),
+                "{} is a graph pack; only a graph pack compiled into this binary can be \
+                 installed today, so it cannot be installed from the registry yet",
+                args.package
+            );
             super::graph::install(args, true).await
         }
         PackKind::Documents => {
```
jackzampolin, line 336: **Bug:** `validate_pack_manifest` allows `plugins` on any kind, and the README says installing a pack installs those plugins into the same store as `gents plugin install`. Only the `PackKind::Assets | PackKind::Plugins` arm calls `install_from_pack`. `PackKind::Documents` (and graph) materializes assets for config apply and never records plugins, so a document pack that declares `plugins/` artifacts will validate, download, and still leave `gents plugin run <name>` as "not installed".

**Fix:** Either install declared plugins for every kind after a successful pack install (using `manifest.metadata.namespace`), or reject `plugins` on document/graph packs in `validate_pack_manifest` so the contract is one-sided on purpose.

crates/gents-cli/src/commands/pack/registry.rs
```diff
@@ -0,0 +1,697 @@
...
+        downloaded
+    };
+
+    let archive = PackArchive::from_bytes(&bytes).with_context(|| {
+        format!("{namespace}/{name}@{version} from the registry is not a readable pack")
+    })?;
+    let digest = archive.digest().with_context(|| {
+        format!("{namespace}/{name}@{version} from the registry failed its own content check")
+    })?;
+    Ok(RegistryPack {
```
jackzampolin, line 320: **Bug:** After digest-verifying the tarball, `fetch_pack` records the *request* namespace/name and never checks them against `archive.manifest()`. Plugin install does the same: it records the CLI coordinate and only copies `language` from the `.afb`. Install then keys the asset cache and plugin store off the *archive* pack namespace/name. A registry mapping bug or a tarball whose inner identity differs from the URL can install another pack's plugins under the requested name (or report `registry (acme/widget@1.0.0)` while writing `pack: other_name`).

**Fix:** After `PackArchive::from_bytes` / `Afb::from_bytes`, require `manifest.name`, `metadata.namespace`, and `version` (and for plugins, `afb.manifest.package.{namespace,name,version}`) to equal the requested coordinate. Refuse on mismatch, naming both identities. Keep `RegistryPack.{namespace,name,version}` sourced from the archive, not the URL.

crates/gents-cli/src/commands/pack/mod.rs
```diff
@@ -398,7 +351,19 @@ async fn install(args: PackInstallArgs) -> Result<()> {
             }
             let temp = tempfile::tempdir()?;
             materialize(&pack, temp.path())?;
-            let (_, report) = crate::desired_state::load_manifest_root(temp.path());
+            // Authored packs omit the principal owner so install can bind it.
+            // Decode without a scope first; if the bundle needs an owner, fill
+            // a validation DID the same way catalog checks do.
+            let (manifest, report) = crate::desired_state::load_manifest_root(temp.path());
```
jackzampolin, line 357: **Bug:** Registry document-pack install materializes declared assets, then calls `load_manifest_root`, which only accepts `pack_config.json` at the temp root. `PackManifest.config` may name another declared asset; `validate_pack_manifest` does not require the config path to be `pack_config.json`. First-party packs happen to use that name, so bundled install still works; a registry document pack with `"config": "config.json"` downloads and materializes, then fails as "missing canonical pack_config.json".

**Fix:** Point the desired-state load at `manifest.config` (still requiring it to be a declared, distributable asset), or require `config == "pack_config.json"` in `validate_pack_manifest` so a pack that cannot be installed cannot be published.

crates/gents-cli/src/commands/pack/registry.rs
```diff
@@ -0,0 +1,697 @@
...
+        );
+        Ok(response
+            .bytes()
+            .await
+            .with_context(|| format!("reading the download body from {url}"))?
+            .to_vec())
+    }
+
+    pub(crate) async fn publish(&self, token: &str, bytes: Vec<u8>) -> Result<Value> {
+        let url = self.api("/publish");
```
jackzampolin, line 211: **Suggestion:** `RegistryClient::publish` always `POST`s `/api/v1/publish` and ignores `RegistryKind`. GET paths were split into `/packs` vs `/packages` specifically because hitting the wrong surface 404s; publish has no equivalent and no fake-registry test. Afterburner's plugin client also uses `/api/v1/publish` with `Content-Type: application/octet-stream`; this client sets neither a kind path nor a content type. If the gents registry distinguishes pack vs plugin publish, `gents pack publish` is the same class of mix-up as the old install route.

**Fix:** Give publish a kind-specific path (or an explicit content-type/header the registry documents), set `Content-Type: application/octet-stream`, and cover both surfaces with a fake-registry test the way download already is.

crates/gents/src/plugin.rs
```diff
@@ -0,0 +1,616 @@
...
+            verdict: PluginVerdict::BadOutput,
+            output: serde_json::Value::Null,
+            diagnostics: append_notes(diagnostics_base, &notes),
+            fuel_used: output.fuel_used,
+            wall_ms,
+        };
+    }
+
+    match serde_json::from_slice::<serde_json::Value>(&output.stdout) {
+        Ok(value) => PluginOutcome {
```
jackzampolin, line 468: **Suggestion:** `outcome_from_exit` notes a non-zero WASI exit code in diagnostics but still returns `PluginVerdict::Success` when stdout is one JSON value. `gents plugin run` then prints that value and exits 0. A plugin that writes `{"ok":true}` and `proc_exit(1)` is indistinguishable from a successful run unless the caller inspects diagnostics.

**Fix:** Map `code != 0` to a failure verdict (new or `BadOutput`/`Refused`) and have `plugin/run.rs` propagate it as an error. If exit codes are not part of the ABI, drop the note so Success is honest.

crates/gents-cli/src/commands/afterburner_build.rs
```diff
@@ -0,0 +1,372 @@
+//! One canonical way to compile an Afterburner package source directory
```
jackzampolin, line 1: **Suggestion:** Several new modules lead with long comments that narrate design history, rejected alternatives, and "why this file exists" rather than a short invariant. Same pattern in `pack.rs` (`vertexia: kept in sync by hand…`), `registry.rs` (the `/packages` mix-up story), and the Cargo.toml afterburner pins. That restates the code and will rot independently of the admission/digest checks.

**Fix:** Keep comments to the non-obvious constraint (e.g. "digest the bytes before `PackArchive::from_bytes`"; "serialize toolchain compiles in tests"). Move history to the PR description.

crates/gents-cli/src/commands/pack/mod.rs
```diff
@@ -353,16 +295,21 @@ fn prune(args: PackPruneArgs) -> Result<()> {
...
+        "pack": pack.manifest().name,
+        "current_digest": pack.digest(),
         "removed_digests": removed,
     }))
 }
 
 async fn install(args: PackInstallArgs) -> Result<()> {
-    let pack = resolve_pack(&args.package)?;
+    let pack = resolve_pack_source(&args.package, args.registry.as_deref()).await?;
+    eprintln!(
```
jackzampolin, line 306: **Nit:** `install` prints `eprintln!("gents pack install: resolved {} from {}", …)` on every install. Repo rules say `tracing`, never `println!`; this also has no tracing span around registry fetch. The curl helper at `registry.rs:106` tells the user to `gents pack install <file>`, which is not a supported argument.

**Fix:** Use `tracing::info!` (or drop the line; JSON output already includes `"source"`). Fix the curl text to `gents pack fetch` / a real file-install path if you add one.

jackzampolin:
Published stack: #1509 → #1510 → #1454 → #1511 → #1512. This PR now targets perf/onboarding-chat-responsiveness and remains draft while integrated validation completes. #1512 owns the remaining unified model-facing pack preview/binding/management integration.

PR #1456, opened 2026-09-10
fix: keep mobile replication live under contention

jackzampolin:
## Summary

- bound every embedded transaction phase and release the Gents write gate after one bounded rollback attempt
- model hydration delivery, ambiguous transport outcomes, and terminal persistence separately, then converge failed terminal writes on a later sweep
- consume native document-change subscriptions and coalesce observer bursts at the database boundary
- remove per-lease cancel-mirror scans while retaining direct tool-call handling and periodic durable recovery
- add a deterministic 32-writer hot-document regression that proves all writes persist, observer invalidations coalesce, and the following write remains live
- scope pairing reconciliation wakes, skip redundant migration baselines, and expose phase-level enrollment/reconnect latency diagnostics
- consume DefraDB durable reconnect recovery and batched collection subscription setup from the stacked database fixes

## Correctness boundary

Transaction watchdogs are operational containment outside the current Lean transition model; they do not claim an ambiguous DefraDB transaction was cleaned up. Storage timeouts retain phase-specific diagnostics and flow through retry::transaction_storage_failure.

Hydration convergence is modeled Lean -> generated conformance -> Rust. An indeterminate push confirms nothing and remains pending. Delivery success followed by a failed terminal write also remains pending and converges on a later sweep; repeated delivery is idempotent. Periodic durable recovery owns correctness without application heartbeats.

Current main already supplies native desktop subscriptions and sticky resync, so the superseded desktop observer patch is intentionally omitted.

## Validation

- full Gents unit suite: 2,268 passed, 4 ignored
- Lean/generated conformance: 265 passed
- all subsequent integration groups passed
- workspace all-target check passed
- deterministic app-pairing contract: 5/5, covering streaming, offline reopen, 2,500-revision recovery, local transcript pagination, and independent DB-to-observer projection
- deterministic 32-writer regression retains every write and keeps the following canonical write live
- live app-pairing E2E against workstation-2 / GLM-5.3-Flash-NVFP4: 1/1; fresh server and client through client-visible completed conversation in 10.86s wall clock

On the final dependency stack, clean client startup measured 55-116ms and aged startup 130ms. Collection subscription setup measured 4-11ms, initial snapshot 3-23ms, offline reopen 0.78-1.05s, steady request delivery 95-215ms, and DB-to-observer projection approximately 0-4ms at test resolution. No deterministic run used event-drop recovery.

## Dependency stack

Pinned to the DefraDB stack ending at sourcenetwork/defradb.rs#1748 (3173a2e8), stacked on #1747 and #1742. That stack makes durable subscription intent authoritative, batches setup/removal, restores Go-compatible list semantics, and heals failed transport installation and cleanup in-process without application-side replay machinery.


jackzampolin:
Independent agent review found one blocker and two follow-ups. The blocker is resolved in `aa7a81628`: DefraDB push failures are now modeled as indeterminate after dispatch, with attempted documents separate from confirmed-complete delivery; Lean proves attempts stay within the authorized selected set, generated conformance covers both counts, and Rust tests repeated fail-after-dispatch attempts. The real terminal-store failure path now feeds the modeled failed transition before returning its phase-specific error. The 32-writer claim remains deliberately narrow: it proves persistence, coalescing, and following-write liveness, not 4,096-entry overflow recovery.

jackzampolin:
## Post-refactor integration and retest ledger

This branch is now a green behavioral checkpoint to replay onto the completed configuration-runtime refactor, not a base that the refactor must descend from.

### Event delivery status

The phone-critical drop storm is structurally contained, but the raw event bus is not globally lossless. The pinned DefraDB revision provides document-change invalidations that coalesce by `(collection, document)`, retain a bounded wakeup, and turn distinct-document overflow into a sticky `resync_required` signal. Critical Gents consumers now treat notifications as hints and recover from durable state:

- request watcher: coalesced document changes plus authoritative queue scan;
- session hydration: collection-filtered wakeups, bounded delivery attempts, and periodic durable recovery;
- runtime control watcher: sticky resync until a full reload succeeds;
- cancel mirror: direct tool-call handling plus periodic durable recovery, without a global scan on every lease/progress update;
- desktop/mobile projection: coalesced invalidation and sticky scoped/full resync.

The underlying per-event channel still drops when full and logs each drop synchronously. Several non-migrated consumers still subscribe to the raw global update stream and sweep broadly: pairing, enrollment, persona requests, directory projection, background completion, and trigger/goal/subagent sources. Their periodic scans protect eventual correctness, but they can still amplify load and latency. Pairing/enrollment/persona are the first migration targets because they sit directly on the dataplane boundary.

Existing projection regressions exercise both 12,000 revisions of one document (coalescing) and 12,000 distinct documents (real overflow, one resync, continued observation). Carry both through the refactor and add equivalent overflow/recovery coverage for each migrated dataplane consumer.

### Remaining correctness and stress work

1. Reproduce mixed local/HTTP writer contention deterministically. The isolated code-review run produced transaction conflicts and 60-second write-gate acquisition timeouts under concurrent graph scans and progress writes. The watchdog preserves Gents liveness, but it does not prove cleanup of an ambiguous DefraDB transaction or identify the primitive that stalled.
2. Preserve phase-specific transaction diagnostics and isolate begin/execute/callback/commit/rollback behavior. Retest that a timed-out operation cannot strand the following canonical write.
3. Port hydration as one vertical slice: Lean transition, generated conformance, Rust implementation, and tests. Retest confirmed delivery followed by terminal-write failure, eventual convergence, fail-after-dispatch ambiguity, authorized-document bounds, and idempotent replay.
4. Make and test the reconnect-triggered hydration retry policy explicitly; periodic durable recovery currently owns correctness.
5. Rerun the canonical mobile E2E from fresh enrollment/pairing through pending, streaming, and a completed reply projected from the phone database. Then cover offline reopen/catch-up, the 2,500-revision fixture, transcript pagination, and session-on-demand sync.
6. Measure each boundary separately: client write -> server visibility, claim -> first progress, response write -> phone DB, phone DB -> UI, and offline convergence. The last observed steady-state request delivery was 168-608 ms; offline reopen was roughly 3-9 seconds and remains the clearest latency target.
7. Rerun the code-review pack stress scenario after the refactor with a fixed worker count and record conflicts, write-gate timeouts, inference retries, event coalescing, resyncs, and total request amplification.

The merge gate is behavioral parity or improvement on this ledger—not a conflict-free textual rebase.

PR #1558, opened 2026-09-18
feat(runtime): native user services, OS logging, and independent desktop controls

jackzampolin:
## Scope

Replace the rejected custom rotating writer and detached log supervisor with a per-user native runtime service and OS-owned diagnostics:

- macOS LaunchAgent ownership and Linux `systemctl --user` ownership; no root daemon, linger change, or hidden supervisor fallback;
- explicit desktop Start, Stop, Restart, and menu-bar controls for the foreground `gents server` service, while closing a frontend leaves the runtime alone;
- `tracing-oslog` on Apple platforms and `tracing-journald` on Linux; service stdout/stderr is journal-owned on Linux;
- foreground CLI continues to use stderr, while native service definitions opt into OS logging;
- no application rotation, raw-child-output collector, locking/rename protocol, or custom retention quota. Existing desktop log files are left untouched.

The branch name is retained for stack stability; it no longer describes the implementation. This PR also closes the #1574 invalid-port startup fixture correction: the fixture uses the ordinary 30-second process-startup budget while retaining the rejection, actionable-error, and no-readiness assertions. Runtime deadlines are unchanged.

Closes #1548
Closes #1574

## Stack and baseline

Part of GitHub-native stack 1570 for #1543 (REST: `https://api.github.com/repos/gents-ai/gents/stacks/1570`). Its ordered members are #1557, #1558, #1559, and #1560. Landing uses GitHub's stack merge flow: merge a contiguous prefix beginning with the lowest unmerged entry; selecting a higher entry includes the lower entries. GitHub performs the follow-on rebase/retargeting for remaining entries. Do not change an individual member's base.

Original fetched-main baseline: `047a4ab9918c8c6f39209de920fcc4f08f09d26b`. The final stack integrates current main at `0deb7659c696e70efc2f87c3aae114ab8f490fe0`; its top-only fast-forward tip is `84878c3d0c0dadec724085f80784387a7b0939f4`, while this native-runtime boundary is `ef077d9dc`. The Linux image build exposed a missing Docker build-context asset: `Dockerfile.runtime.dockerignore` now has a narrow three-line exception for committed `third_party/defradb-explorer/dist`; no Rust or TypeScript source changed. The later main-only `d4d889145` (.18.1 and control-plane fix) merges cleanly and requires no stack rebase. The older published integrated evaluation tip was `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`.

## Published evidence (pre-final revision)

[...]

PR #1557, opened 2026-09-18
fix(logging): quiet resolver and no-op write noise

jackzampolin:
## Scope

Quiet hot-path resolver and known no-op write telemetry. Ignored-setting diagnostics stay with the common desired-state owner; clean, known zero-write attempts are DEBUG while unknown, retry, conflict, and recovery cases retain their useful visibility. Includes the initial storage-investigation handoff.

Closes #1546
Closes #1547

## Stack and baseline

Part of GitHub-native stack 1570 for #1543 (REST: `https://api.github.com/repos/gents-ai/gents/stacks/1570`). Its ordered members are #1557, #1558, #1559, and #1560. Landing uses GitHub's stack merge flow: merge a contiguous prefix beginning with the lowest unmerged entry; selecting a higher entry includes the lower entries. GitHub performs the follow-on rebase/retargeting for remaining entries. Do not change an individual member's base.

Original fetched-main baseline: `047a4ab9918c8c6f39209de920fcc4f08f09d26b`. The final stack integrates current main at `0deb7659c696e70efc2f87c3aae114ab8f490fe0`; its top-only fast-forward tip is `84878c3d0c0dadec724085f80784387a7b0939f4`, while this quiet-logs boundary is `b51464f7d`. The Linux image build exposed a missing Docker build-context asset: `Dockerfile.runtime.dockerignore` now has a narrow three-line exception for committed `third_party/defradb-explorer/dist`; no Rust or TypeScript source changed. The later main-only `d4d889145` (.18.1 and control-plane fix) merges cleanly and requires no stack rebase. The older published integrated evaluation tip was `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`.

## Published evidence (pre-final revision)

The integrated stack at `33bf617f4` recorded `cargo test -p gents --no-fail-fast` (3,040 passed, 8 ignored), `cargo check --workspace --all-targets`, `lake build`, generated contracts, 14 protocol capture tests, and corrected CLI enrollment/runtime harnesses (6 and 40). Those results are integrated-stack evidence, not isolated full-suite runs for each PR boundary.

The four N=1/C=1 live smoke catalogs at that same clean tip passed: progressive 4/4, mailbox 5/5, stewardship 10/10, and maintenance 8/8. They are smoke trials, not a larger acceptance cohort. The initial full CLI invocation was not green; load-dependent aged-runtime enrollment remains tracked in #1555.

## Current-source validation and remaining checks

[...]

Issue #1555, opened 2026-09-18
Flaky: aged-runtime enrollment can miss readiness within 30s


jackzampolin:
## Failure

`cargo test -p gents-cli` failed on 2026-09-18 in `contract::fresh_app_pairs_with_aged_runtime_with_bounded_latency`:

```text
aged runtime readiness fixture ready revisions=2500 elapsed_ms=61607
enroll/approve/readiness exceeded 30s or failed: Err(Elapsed(()))
```

At failure, the runtime had an `AgentBehaviorReadiness` row while the client query returned an empty `AgentBehaviorReadiness` array. The test diagnostics also recorded a runtime P2P history of 10,262 CAR blocks and a warning that a 3.44 MB CAR response was truncated at the 10,000-block server-side limit. These are correlated diagnostics, not a proven root cause. The same log includes `PushLog ... already being processed` and a saturated non-authoritative broadcast-pool warning.

Full-suite evidence is retained at `/tmp/gents-storage-footprint-cli-tests-2.log` in the storage-footprint worktree (failure lines 1135-1205).

## Isolated control

Using the already-compiled test binary, with no Cargo rebuild:

```sh
cd crates/gents-cli
../../target/debug/deps/cli_enrollment-8f7a4d2dbcaac05c \
  --exact contract::fresh_app_pairs_with_aged_runtime_with_bounded_latency --nocapture
```

passed: `1 passed; 0 failed; 5 filtered out; finished in 37.24s`. Its fixture completed in 17,998ms and enrollment in 7,161ms; the client saw 2,519 events but fetched only five documents, showing normal coalescing on the passing path. Evidence is retained at `/tmp/gents-cli-aged-enrollment-repro.log`.

## Scope and next step

The contract already serializes its six enrollment cases behind `ENROLLMENT_E2E_LOCK`; the recorded `cargo test -p gents-cli` runs integration binaries sequentially. This is therefore not resolved by adding another local mutex, and the 30-second bounded-latency assertion must remain meaningful.

Investigate the runtime/client P2P history-recovery path under 2,500 signed readiness rewrites: determine whether the current readiness document is delayed behind historical DAG/CAR transfer, and add deterministic observability/reproduction before changing limits, retries, or test budgets. This is separate from the storage-footprint change set. Related architectural context: #1393; the closed #630 documents a distinct historical PushLog-backpressure incident.

Issue #1543, opened 2026-09-18
Storage footprint investigation: where a Gents home's bytes go


jackzampolin:
## Summary

First byte-level look at what a Gents home actually stores. Source: one single-agent demo home (~4 hours of active work, 85 requests, 1,864 inference calls, 4,001 messages, 1,989 tool calls), inspected offline on a private copy with a read-only regolith scanner (no runtime started, source home untouched). Figures below are aggregate counts/sizes and collection/field names only.

**Headline: the conversation itself is ~30 MiB. The home is 2.1 GiB.** About 97% of the agent database is three things: full provider request bodies stored per inference call, full-text rewrites of the streaming reasoning tail, and the per-commit bookkeeping those writes drag along. Separately, the desktop log grows ~300 MiB/day while idle.

## Home breakdown (allocated)

| Component | MiB |
| --- | ---: |
| Agent database (69 SSTs, LZ4, 16 KiB blocks) | 1,701 |
| Desktop log (single file, no rotation) | 379 |
| Desktop database | 43 |

Agent DB: 2,522,401 keys, **4,309 MiB logical → 1,701 MiB on disk** (~2.5x LZ4). APFS allocated ≈ logical file size, so no filesystem distortion. 68 of 69 SSTs carried one compaction timestamp, so this is retained data, not un-compacted garbage.

| DefraDB store | Keys | Logical MiB | What |
| --- | ---: | ---: | --- |
| blockstore `b` | 806k | 2,664 | signed Merkle-DAG history blocks |
| datastore `d` | 106k | 1,515 | current document bodies + secondary indexes |
| headstore `h` | 988k | 68 | heads / priority indexes (all key bytes) |
| systemstore `s` | 621k | 62 | almost entirely the block→document index |

## Where the logical bytes go

| Data | MiB | Share | Read by | Verdict |
| --- | ---: | ---: | --- | --- |
| Transcript as current docs (sessions, messages, tool calls, inference-call metadata) | ~30 | 0.7% | everything | needed |
| `RenderedRequest.request_json` (doc + history block) | ~1,706 | 40% | `gents trace capture --include-body`, context-details tool on one provider path, conformance tests. Never selected by timeline/projections/desktop; excluded from P2P | opt-in debugging only |
| `RenderedRequest.provenance_json` (doc + history block) | ~1,243 | 29% | reduction-consumption gate (needs reduction keys + status), trace status column | needed, but embeds message copies nobody reads back |
[...]

PR #1559, opened 2026-09-18
fix(streaming): reduce progress cadence and reasoning rewrites

### LINKED ISSUE #1545: Streaming progress rewrites the full reasoning tail every ~100 ms: 164x write amplification
## Problem

Streaming progress persists the **full accumulated text on every flush**. `DefraStreamWriter::flush_snapshot` (`crates/gents/src/streaming.rs:151`) writes `content`, `reasoning`, `token_count`, `reasoning_progress_seq` to `AgentResponse` under `lifecycle.response_progress`, and the same transaction renews the lease and bumps `execution_progress_seq` on `AgentRequest`. DefraDB keeps every version forever, so each flush permanently adds a field block per changed field, a composite block, a collection block and a signature block per document, plus head and index keys.

Measured on the investigated agent database (see #1543):

| | |
| --- | ---: |
| `lifecycle.response_progress` commits | 42,638 (for ~1,864 inference calls) |
| Median gap between progress writes | 83 ms (81% under 150 ms) — the `DEFAULT_STREAM_BATCH_MS = 100` default; shipped packs use 5000 |
| `AgentResponse.reasoning` versions | 36,224 across 51 docs, avg 27 KB each |
| Bytes written for `reasoning` | 929 MiB |
| Final text of every stream, summed | 5.7 MiB → **~164x amplification** |
| Share of the database's logical bytes | 22% |
| `AgentRequest.execution_lease_expires_at` versions riding along | 42,292 over 85 docs |

Nothing reads an old version. The live readers (desktop `live_delta.rs` / `projection.rs`, `gents chat`, Codex and Grok shims) read only the newest value; the durable copy is `AgentMessage.reasoning`; `reset_tail` clears `AgentResponse.content`/`reasoning` at completion and at tool-result boundaries. `AgentResponse` is a subscribed `@branchable` collection, so every intermediate version is also gossiped to paired peers. LZ4 cannot recover this: each version is its own content-addressed value.

## Proposal (existing owners, no side channel)

1. Raise `DEFAULT_STREAM_BATCH_MS` from 100 toward 1000 (`crates/gents/src/config.rs:16`). ~10x fewer progress commits, and the same factor off the lease renewals and per-commit bookkeeping (~7% of the database) that ride on each one. The first-visible-token bypass keeps perceived latency.
2. Reduce `MAX_LIVE_REASONING_BYTES` from 64 KiB to a few KiB (`streaming.rs:18`). The live view shows a tail; the full text already lands in `AgentMessage.reasoning`.
3. Consider the same bounded-tail treatment for `content`, which is currently unbounded and fully rewritten each flush.
4. Check whether lease renewal needs to be a durable commit on every progress write or can ride at its own, slower cadence.

Estimated effect of 1+2 on this database: 929 MiB → roughly 10–15 MiB. That is an estimate; an isolated reproduction (same synthetic stream at 100 ms / 64 KiB vs 1000 ms / 4 KiB, scanned with the same tool) will be posted here.

The structural fix is upstream: DefraDB fields that keep no commit history (sourcenetwork/defradb.rs#1804). Gents should not grow a separate streaming channel.

## Validation

- Desktop live tail, `gents chat`, and both shims still render progressively (the `live_delta` patcher already handles append/replace).
- Storage regression check: bytes written per 1,000 streamed reasoning deltas, before/after.

### LINKED ISSUE #1554: Streaming enrollment test assumed sub-500ms flush after cadence default changed
Follow-up regression from #1545.

`cli_enrollment::contract::buffered_stream_chunk_reaches_client_during_provider_pause` used the same 500 ms visibility deadline as the unpaced first-visible streaming test. The paced fixture emits two small content frames 20 ms apart and deliberately keeps provider completion gated. After the production `stream_batch_ms` default moved to 1000 ms to reduce streaming write amplification, the buffered frame is not expected to flush inside 500 ms.

Observed in the full CLI suite on 2026-09-18:

- `streaming_content_reaches_client_before_provider_completion` passed.
- `buffered_stream_chunk_reaches_client_during_provider_pause` failed with `streaming content did not reach client projection before provider completion`.
- Provider requests were observed, the client observer remained healthy, and the deadline expired at the stale visibility assertion.

The regression fix keeps the unpaced contract at 500 ms and threads an explicit 2 s budget only through the paced cadence case. The provider completion semaphore remains closed until streaming content is visible, so terminal persistence cannot satisfy the test.

Acceptance: retain separate deadlines for immediate/unpaced visibility and timer-driven buffered visibility, with the latter derived from or documented against the production cadence default.

jackzampolin:
## Scope

Reduce durable stream churn without changing terminal ownership: default batching moves from 100 ms to 1,000 ms and the shared live reasoning preview moves from 64 KiB to 4 KiB. Preserve configured overrides, first-visible bypass, final transcript, lease semantics, and terminal owners.

The actual pending-snapshot replay measures 11 rather than 101 snapshots and 32,528 rather than 307,200 selected bytes at the same 4 KiB cap. The paced CLI visibility fixture is scoped to the new cadence; the ordinary 500 ms first-visible bound remains.

Closes #1545
Closes #1554

## Stack and baseline

Part of GitHub-native stack 1570 for #1543 (REST: `https://api.github.com/repos/gents-ai/gents/stacks/1570`). Its ordered members are #1557, #1558, #1559, and #1560. Landing uses GitHub's stack merge flow: merge a contiguous prefix beginning with the lowest unmerged entry; selecting a higher entry includes the lower entries. GitHub performs the follow-on rebase/retargeting for remaining entries. Do not change an individual member's base.

Original fetched-main baseline: `047a4ab9918c8c6f39209de920fcc4f08f09d26b`. The final stack integrates current main at `0deb7659c696e70efc2f87c3aae114ab8f490fe0`; its top-only fast-forward tip is `84878c3d0c0dadec724085f80784387a7b0939f4`, while this streaming boundary is `b1effd453`. The Linux image build exposed a missing Docker build-context asset: `Dockerfile.runtime.dockerignore` now has a narrow three-line exception for committed `third_party/defradb-explorer/dist`; no Rust or TypeScript source changed. The later main-only `d4d889145` (.18.1 and control-plane fix) merges cleanly and requires no stack rebase. The older published integrated evaluation tip was `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`.

## Published evidence (pre-final revision)

At the published integrated tip, `cargo test -p gents --no-fail-fast` recorded 3,040 passed and 8 ignored; `cargo check --workspace --all-targets`, `lake build`, generated contracts, 14 protocol capture tests, and corrected CLI enrollment/runtime harnesses (6 and 40) also passed. These are integrated-stack results, not isolated full-suite runs at every intermediate PR boundary.

At N=1/C=1, four live smoke catalogs passed at that clean tip: progressive 4/4, mailbox 5/5, stewardship 10/10, and maintenance 8/8. They are smoke trials, not a larger acceptance cohort. The initial full CLI invocation was not green; #1555 remains the tracked aged-runtime enrollment investigation.

## Current-source validation and remaining checks

At the final pushed source, `cargo test -p gents` passed 3,052 tests with 8 ignored; CLI units 922, CLI server 15, native integrations 11, bridge 167 / 2 ignored, launcher 3, Tauri 6, protocol 14, workspace all-targets, Lean, UI 501 / 14 skipped, desktop client 33, UI build, Prettier, Rustfmt, sidecar checks, ShellCheck, and package-boundary checks passed. The earlier exit-137 runtime compile interruption occurred before tests began; its retry passed, with no failed test.

Live smoke trials at clean production revision `a51eb92551242b9f28f68ec4ad8fc4cfc816373b` passed progressive 4/4, mailbox 5/5, maintenance 8/8, and stewardship 10/10 on its follow-up trial (each N=1/C=1, GLM-5.3-Flash-NVFP4). The first stewardship trial was intentionally interrupted after repeated invalid `AgentTools` plan proposals despite terminal `unknown canonical collection` feedback; all ten cases were unreported, not passing. Tracked in #1579; the follow-up pass does not erase that incomplete result. These are smoke trials, not a larger acceptance cohort.

The Linux eval image at that production revision built successfully: `gents-eval-runtime:storage-native-final-a51eb9255`, image `sha256:121eb06335daae569aa80733ef2220fbb750bff4e17afa06947e1258ebaab3e8`; revision and clean-source labels verified.

The subsequent `84878c3d0` commit changes only four CLI test files. Full-candidate CI at `a51eb9255` exposed four stderr assertions that filtered newly traced INFO/WARN diagnostics with fixture `RUST_LOG=error`. The fixtures now enable their asserted component diagnostics and isolate inherited native-service logging while retaining explicit overrides. No production, schema, proof, UI, or image-build source changed after the live trials. Validation rerun: runtime 3,052 passed / 8 ignored, workspace all-targets and Rustfmt passed; entire shim suite with parent `GENTS_SYSTEM_LOG=1` passed 32 / 8 ignored, including all four prior failures. [Corrected immutable full-candidate CI](https://github.com/gents-ai/gents/actions/runs/35404729133) passed all 11 jobs, including the full CLI suite and desktop Rust tests.

Real OS service/login/reboot/logging smoke remains unrun and belongs to the native-runtime validation boundary, not to this streaming change.

See `contracts/storage-footprint.md` for replay methodology, measured limits, and the distinction between logical serialized bytes/write counts and physical storage. See `contracts/native-runtime-service.md` for the parent service/logging boundary.


jackzampolin:
Storage-footprint stack (merge bottom-up): #1557 → #1558 → #1559 → #1560. Integrated eval tip: `33bf617f4`; live eval results will be recorded on #1560. Non-main-base closing keywords take effect after the relevant changes reach the default branch.

Issue #1425, opened 2026-09-08
Explore using DefraDB _docID for configuration references and runtime identity


jackzampolin:
Configuration documents currently carry logical keys such as `behavior_id`, `selection_id`, and `profile_id` alongside DefraDB's `_docID`. Explore using `_docID` as the persisted reference and runtime identity while retaining portable names in authored manifests.

This is deferred from the context-model refactor (`refactor/context-model`). That refactor keeps the existing logical IDs; this issue is exploration, not an implementation commitment.

Questions to resolve:

- How should manifest-local names resolve to document IDs during installation, and where should the installation mapping live using existing DefraDB documents and adapters?
- How do repeated apply, updates, deletion/recreation, pruning, and independent installations preserve the intended references?
- How should cyclic references be materialized and validated transactionally without exposing incomplete runnable configuration?
- Which graph, task, session, subagent, CLI, and UI consumers need names versus document IDs?
- How do existing DID ownership, ACP authorization, transaction, and P2P boundaries apply to resolution?
- Can explicit domain IDs be removed without replacing them with an equally complex identity layer?

The pinned DefraDB dependency derives a document ID from its genesis composite CID. Editing or replicating that document preserves its identity; independently materializing configuration requires examining whether the same genesis can be reproduced. Document identity is separate from revision identity: switching references to `_docID` does not pin a configuration revision or its transitive dependencies.

Expected result: an evidence-backed design recommendation, comparison with keeping logical IDs, a migration/conformance plan, and an inventory of affected reference owners. Verify assumptions against the pinned DefraDB source and existing config apply/recreation behavior before choosing an approach.

jackzampolin:
Linked follow-on epic: #1571 — persist streamed output once, with canonical transcript storage and document references. Its required stack is data-model/specification → Lean → conformance → implementation. This issue remains the reference-design workstream: the epic coordinates transcript/request/output references with this exploration, preserves the distinction between document identity and revision/idempotency/execution/provider identities, and does not assume every logical ID can be replaced mechanically.

PR #1560, opened 2026-09-18
feat(storage): lossless witnessed captures and footprint validation

### LINKED ISSUE #1543: Storage footprint investigation: where a Gents home's bytes go

### LINKED ISSUE #1544: RenderedRequest bodies are 69% of agent storage: stop re-storing the whole transcript on every provider call
## Problem

`RenderedRequest` is 69% of the logical bytes in the investigated agent database (see #1543): 1,864 rows, one per provider send, averaging 832 KB (max 2.2 MB). `request_json` is 856 MiB and `provenance_json` 621 MiB, and DefraDB stores each once as the current document and once as the history block, so ~2.95 GiB logical in total.

The content is almost entirely a re-rendering of data already held in other documents:

| Field | Stored MiB | Distinct content |
| --- | ---: | --- |
| `request_json` | 856 | 8.4 MiB unique chunks; 46.6 MiB differs from the previous body in the same session |
| `provenance_json` | 621 | 8.7 MiB unique chunks |
| pooled | 1,478 | 12.6 MiB — about the size of the transcript itself |

Each body is written once (`@immutable`), so this is not rewrite churn; it is O(turns²) growth from storing the whole transcript again on every call. Storage-engine compression cannot help: each body is a separate value and LZ4 works within one 16 KiB block.

## Who reads it

- `request_json`: only `gents trace capture --include-body` (`crates/gents-cli/src/commands/trace.rs:82`), the context-details tool on the `openai_chat_completions` path (`crates/gents/src/toolset/session_history/context_details.rs:50`), and conformance/probe tests. `run_timeline_fetch/event_loaders.rs` deliberately never selects it; the desktop never reads the collection; it is a local-audit collection excluded from P2P.
- `provenance_json`: the reduction-consumption gate (`crates/gents/src/provider_context_reduction.rs:419`, `_like` on the reduction key), plus `provenance_status` in the timeline/projections. Those need reduction keys, status, and admission join — not the embedded message copies in `assembly_trace`.

Capture is unconditional (`crates/gents/src/agent/builder.rs:271`) with no retention or policy.

## Proposal

Two separable changes:

1. **`provenance_json.assembly_trace` by reference.** Carry message keys / tool-call keys / reduction keys instead of embedded native-message copies. Keeps the consumption gate and status readers working. ~1.2 GiB of the 4.3 GiB here, no change to what the capture claims.
2. **`request_json` as a verifiable projection.** Store a SHA-256 of the canonical body plus references to its inputs (message keys, reduction keys, config commit CIDs, renderer/capture version). `trace capture --include-body` and the context-details tool re-render on demand and check the digest. The "this is what the provider received" claim stays checkable; a renderer change shows up as a digest mismatch instead of silent drift. Cost: a body whose renderer version is no longer available can be verified-as-different but not recovered.
   - Lossless fallback if that cost is unacceptable: prefix-delta storage (reference to the previous capture in the scope + suffix). 46.6 MiB instead of 856 MiB here, with no dependence on renderer determinism.

## Prior decision this must answer

The schema comment (`crates/gents-schemas/schemas/agent/rendered_request.graphql`) records a deliberate choice: *no `request_hash` column*, because a stored digest is self-attested, and the integrity witness is the DefraDB field-commit CID over the stored `request_json` bytes. `Proofs/RenderedCapture.lean` fences the send on the matching `(capture_key, canonical request)` being durable. A projection removes the bytes that CID is computed over, so option 2 below is a change to that decision and to the Lean model, not an optimisation under it. The prefix-delta fallback keeps the bytes recoverable and the fence meaningful, and is the smaller departure.

Also relevant: the collection is `@branchable` (deliberate and irreversible per the same comment), so DefraDB's filtered truncate cannot be used for retention of old captures today (see #1543).

## Constraints

- This touches the provider-input capture contract and its conformance oracles (`crates/gents/tests/conformance/rendered_capture.rs`, `tests/e2e_runtime/rendered_request_capture.rs`), so it follows Lean model → conformance → Rust.
- Fail-closed capture at the transport seam must be preserved: the digest and references must be durably written before the body is forwarded.
- `request_commit_cid` and the `request_json` field-commit CID witness (`rendered_request/commits.rs`) need a defined replacement (digest field commit).
- No second persistence owner; no data wipe of existing rows.

## Open decision

Projection (smallest, depends on deterministic re-rendering) vs prefix-delta (lossless, ~18x) for `request_json`. Change 1 does not depend on that decision.

### LINKED ISSUE #1553: Test: Tools export round-trip races runtime discovery and database ownership
Observed while validating the storage-footprint branch based on main 047a4ab9918c8c6f39209de920fcc4f08f09d26b.

Reproduction: cargo test -p gents-cli. The 901 unit tests passed; cli_config then reported 19 passed, 1 failed, 1 ignored. The failing test was cli_config_tools::tools_set_persists_host_root_and_export_round_trips_it. Its config export --root subprocess failed with: database directory is already locked for read-write access. An isolated run of the same compiled test subsequently passed, confirming timing dependence.

The fixture starts a server and writes through an explicit GraphQL endpoint, then exports without an explicit endpoint while keeping the server alive. resolve_config_access attempts runtime discovery and otherwise opens the embedded datastore, which conflicts with the still-running server when discovery does not select it.

Fix in progress on storage-footprint: stop and wait for the test server before the intended offline export. Preserve production locking, avoid sleeps/retries, and retain the host-root export assertion. Re-run the full CLI suite after the fix. This issue tracks the test defect separately from storage changes.

### LINKED ISSUE #1556: Flaky CLI status test assumes reconcile phase remains idle
## Failure

`cli_status::status_reads_local_runtime_context_by_default` failed in the CLI runtime suite:

```
assertion `left == right` failed
  left: Some("debouncing")
 right: Some("idle")
```

The failure is recorded in `/tmp/gents-storage-footprint-cli-tests-2.log` (the CLI runtime suite had 39 passing tests and this one failure).

## Reproduction evidence

Running the already-built `cli_runtime` binary in isolation passed immediately and passed five consecutive further invocations. The retained eight-invocation retry log is `/tmp/gents-storage-footprint-cli-status-rerun.log`.

This points to a timing-sensitive assertion, not a demonstrated product-status regression.

## Likely race

The fixture waits for quiescence, then reads status and asserts `AgentRuntime.reconcile_phase == "idle"`. `reconcile_phase` is a live diagnostic observation. The control watcher can receive a later document or backend-health event and transition it to `debouncing` after the quiescence helper returns, while readiness remains `ready` and the active generation remains valid.

The test should preserve its independent assertions for readiness, runnable behavior count, and the configured local GraphQL endpoint. It should avoid treating an idle reconcile observation as stable. Prefer a typed canonical reconcile-phase parse if the status context contract needs to expose it; otherwise remove this unstable exact-phase assertion rather than masking a product failure.

jackzampolin:
## Scope

Land the Lean model, generated conformance, lossless v2 capture container, bounded CID-witnessed deltas, compact provenance, and all writers/readers together. Existing v1/v3 data remains readable without a schema migration or history rewrite; old binaries cannot decode v2, so writer and readers deploy together.

This final stack PR also carries capture-boundary integration follow-ups, CLI fixture-race fixes, and the storage-footprint measurement/validation write-up. `storage-footprint-04-captures` remains the initial capture-boundary marker.

Closes #1544
Closes #1543
Closes #1553
Closes #1556

## Stack and baseline

Part of GitHub-native stack 1570 for #1543 (REST: `https://api.github.com/repos/gents-ai/gents/stacks/1570`). Its ordered members are #1557, #1558, #1559, and #1560. Landing uses GitHub's stack merge flow: merge a contiguous prefix beginning with the lowest unmerged entry; selecting a higher entry includes the lower entries. GitHub performs the follow-on rebase/retargeting for remaining entries. Do not change an individual member's base.

Original fetched-main baseline: `047a4ab9918c8c6f39209de920fcc4f08f09d26b`. The final stack integrates current main at `0deb7659c696e70efc2f87c3aae114ab8f490fe0`; its top-only fast-forward tip is `84878c3d0c0dadec724085f80784387a7b0939f4`, while this capture boundary is `ae755eb28`. The Linux image build exposed a missing Docker build-context asset: `Dockerfile.runtime.dockerignore` now has a narrow three-line exception for committed `third_party/defradb-explorer/dist`; no Rust or TypeScript source changed. The later main-only `d4d889145` (.18.1 and control-plane fix) merges cleanly and requires no stack rebase. The older published integrated evaluation tip was `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`.

## Published evidence (pre-final revision)

At `33bf617f4`, integrated validation recorded `cargo test -p gents --no-fail-fast` (3,040 passed, 8 ignored), `cargo check --workspace --all-targets`, `lake build`, generated contracts, 14 protocol capture tests, three desktop launcher tests, seven logger tests, and corrected CLI enrollment/runtime harnesses (6 and 40). The capture workload used ten real DefraDB captures: both payloads round-tripped exactly, turns 1–8 used deltas, turn 9 checkpointed, and retained serialized fields were less than half the full-body plus v3-manifest comparison.

Four live N=1/C=1 smoke catalogs passed at that clean tip: progressive 4/4, mailbox 5/5, stewardship 10/10, and maintenance 8/8. They are smoke trials, not a larger acceptance cohort. The original full CLI invocation was not green; #1555 remains open for aged-runtime enrollment under load.

## Current-source validation and remaining checks

At the final pushed source, `cargo test -p gents` passed 3,052 tests with 8 ignored; CLI units 922, CLI server 15, native integrations 11, bridge 167 / 2 ignored, launcher 3, Tauri 6, protocol 14, workspace all-targets, Lean, UI 501 / 14 skipped, desktop client 33, UI build, Prettier, Rustfmt, sidecar checks, ShellCheck, and package-boundary checks passed. The earlier exit-137 runtime compile interruption occurred before tests began; its retry passed, with no failed test.

Live smoke trials at clean production revision `a51eb92551242b9f28f68ec4ad8fc4cfc816373b` passed progressive 4/4, mailbox 5/5, maintenance 8/8, and stewardship 10/10 on its follow-up trial (each N=1/C=1, GLM-5.3-Flash-NVFP4). The first stewardship trial was intentionally interrupted after repeated invalid `AgentTools` plan proposals despite terminal `unknown canonical collection` feedback; all ten cases were unreported, not passing. Tracked in #1579; the follow-up pass does not erase that incomplete result. These are smoke trials, not a larger acceptance cohort.

The Linux eval image at that production revision built successfully: `gents-eval-runtime:storage-native-final-a51eb9255`, image `sha256:121eb06335daae569aa80733ef2220fbb750bff4e17afa06947e1258ebaab3e8`; revision and clean-source labels verified.

The subsequent `84878c3d0` commit changes only four CLI test files. Full-candidate CI at `a51eb9255` exposed four stderr assertions that filtered newly traced INFO/WARN diagnostics with fixture `RUST_LOG=error`. The fixtures now enable their asserted component diagnostics and isolate inherited native-service logging while retaining explicit overrides. No production, schema, proof, UI, or image-build source changed after the live trials. Validation rerun: runtime 3,052 passed / 8 ignored, workspace all-targets and Rustfmt passed; entire shim suite with parent `GENTS_SYSTEM_LOG=1` passed 32 / 8 ignored, including all four prior failures. [Corrected immutable full-candidate CI](https://github.com/gents-ai/gents/actions/runs/35404729133) passed all 11 jobs, including the full CLI suite and desktop Rust tests.

Real macOS LaunchAgent and Linux user-systemd smoke—installation, Start/Stop/Restart, login/reboot persistence, and OS-log inspection—has not been run. Those platform checks are explicitly separate from injected native-command tests and syntax validation.

See `contracts/storage-footprint.md` for codec/readership ownership, measurement limits, unresolved #1555, and no-GC/no-history-rewrite boundaries. See `contracts/native-runtime-service.md` for OS-service, logging, and real-OS validation boundaries.


jackzampolin:
Storage-footprint stack (merge bottom-up): #1557 → #1558 → #1559 → #1560. Integrated eval tip: `33bf617f4`; live eval results will be recorded on #1560. Non-main-base closing keywords take effect after the relevant changes reach the default branch.

jackzampolin:
Live eval checkpoint on integrated tip `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`: GLM-5.3-Flash-NVFP4, temperature 1, top-p 0.95, requested reasoning high; N=1/C=1 per suite. Progressive passed all 4 stages (108.4s); monitor-mailbox passed all 5 stages (97.7s). Shared eval contracts passed (20 JS + 46 Rust, 3 ignored), and isolated host fixture checks passed 9/9. Host stewardship and maintenance remain pending while an exact-revision Linux runtime image builds. This is a smoke trial, not an N=30 acceptance cohort. Private raw evidence is retained locally and is not attached because it includes runtime identities and potentially credentials.

jackzampolin:
## Completed live evals — storage-footprint tip

All four registered suites passed on clean source `33bf617f4f96cdd11cc24b5ae3c9ab4b59fa1efa`, without source/grader changes or reruns.

| Suite | Trial result | Passed stages | Runtime |
| --- | --- | --- | --- |
| progressive-configurator | 1/1 passed | 4/4 | 108.4 s |
| monitor-mailbox | 1/1 passed | 5/5 | 97.7 s |
| host-steward | 1/1 passed | 10/10 | 453.9 s |
| host-maintenance | 1/1 passed | 8/8 | 322.0 s |

All launchers exited 0; no failed, skipped, or unreported live cases. These are **N=1 / C=1 smoke trials per suite**, not a larger acceptance cohort or a disk-footprint benchmark.

Settings: GLM-5.3-Flash-NVFP4, temperature 1, top-p 0.95, requested reasoning high (provider enforcement not measured), default 1,800 s stage timeout. Used the repository make targets `live-configurator-eval`, `live-mailbox-eval`, `live-host-steward-eval`, and `live-host-maintenance-eval`, explicitly setting `GENTS_LIVE_CONFIG_RUNS=1` and `GENTS_LIVE_CONFIG_CONCURRENCY=1`.

Host suites used the current-source image `gents-eval-runtime:storage-footprint-33bf617f4`, immutable ID `sha256:b9c0c37d00df55d3d040fba903de1b7b70deb0f804b117f1c077428944d47026`; image labels match the exact commit and clean source. The fixture's normal teardown completed; no preexisting Docker resources were removed.

Harness validation also passed: 20 JavaScript tests, 46 Rust tests (3 live tests intentionally ignored by that deterministic invocation), and 9 real-container fixture tests.

Private receipts are retained below the operator's `.gents-eval/`, with these unique run IDs:
- `progressive-configurator-2026-09-18T16-58-26-926Z-bWbnbG`
- `monitor-mailbox-2026-09-18T17-01-19-734Z-AhJXXc`
- `host-steward-2026-09-18T17-07-23-520Z-gceHYP`
- `host-maintenance-2026-09-18T17-15-43-054Z-ODsOa2`

Each run retains execution/report JSON, immutable trial evidence, fixture/grader hashes, and raw logs; host runs additionally retain runtime archives and memory receipts. Raw evidence is intentionally not uploaded because it contains identities and can contain credentials. Inspect locally with `node scripts/evals/report.mjs <run-directory>`.

This does not resolve the separately tracked load-dependent enrollment failure #1555 or imply pending CI jobs have passed.

jackzampolin:
Landing integration update: the pushed 4f0d4b1 candidate passed fresh N=1/C=1 progressive-configurator (4/4) and mailbox (5/5) catalogs. Those results are not yet final landing evidence: GitHub native-stack mergeability exposed overlap with current main (0deb7659c: native tabs, v0.18 installers, DB Explorer). The stack has now been integrated locally with those changes, preserving one tray/status owner and adding release-sidecar staging plus stable AppImage executable checks. Updated Rust/UI validation is running before the rebased stack is pushed; remaining live catalogs will use that new exact revision. No services have been registered on a developer host.

jackzampolin:
Final pushed stack tip: `84878c3d0c0dadec724085f80784387a7b0939f4`. All four stack members remain mergeable; no merge has been performed.

The last commit changes only CLI test fixtures: four stderr assertions now enable the component's INFO/WARN diagnostics, and foreground child processes isolate inherited `GENTS_SYSTEM_LOG` while preserving explicit overrides. No production source changed after the live evaluations.

Validation:

- `cargo test -p gents`: 3,052 passed, 8 ignored.
- `cargo check --workspace --all-targets` and Rustfmt: passed.
- Full `cli_codex_shim` suite with parent `GENTS_SYSTEM_LOG=1`: 32 passed, 8 ignored, including all four former CI failures.
- Previously completed integrated checks: CLI unit/server/native-service tests, protocol capture tests, Lean, desktop bridge/launcher/Tauri tests, UI/client tests and build, packaging-script checks.
- Production revision `a51eb9255` live trials, N=1/C=1: progressive 4/4, mailbox 5/5, maintenance 8/8, stewardship follow-up 10/10. The first stewardship trial was interrupted after repeated invalid `AgentTools` plans; its ten cases remain unreported, and #1579 remains open. The follow-up pass is not a larger acceptance cohort or a resolution of that recovery defect.
- Exact-production Linux image build passed with verified clean-source/revision labels.

Fresh full-candidate CI: https://github.com/gents-ai/gents/actions/runs/35404729133 — all 11 jobs passed, including the full CLI and desktop Rust suites.

Remaining release-validation limits: no real LaunchAgent/user-systemd installation, login/reboot persistence, native log inspection, or signed-installer smoke was performed on user machines. Historical CLI enrollment flake #1555 remains open.

Storage measurements are logical serialized bytes: 9.44× reduction in the controlled preview-cadence workload and greater than 2× capture reduction in the growing-transcript fixture. These do not promise a whole-database ratio or reclaim pre-existing history. The broader persist-once/document-reference refactor remains future epic #1571, linked to #1425.

